PR #22, opened 2022-02-07
Delete KubeEnvironments + other things

nturinski:
PR ended up getting a bit away from me, but will break it down below
---
## T2 Azure Credentials

This PR depends on the following PRs:
  - https://github.com/microsoft/vscode-azure-account/commit/d4ec3d8e1e7dec16a03e1cb361b4e1d4f09a7a28
  - https://github.com/microsoft/vscode-azuretools/commit/d3b4e61a34908d8226b20d45881b48427e2cdd91
 
Because the Container Apps sdk is T2, I had to upgrade the credential types in order to use it.  Previously, I was getting around this using the `azure-identity` package, but this brought in a dependency on `keytar` which ended up breaking the vsix on Mac/Linux.  The changes related to getting ride of the `VisualStudioCodeCredentials` such as https://github.com/microsoft/vscode-azurecontainerapps/compare/nat/delete?expand=1#diff-17e919c422cbcd1b90fea9dac31ba7001932c46adae157768f07a1bf5cdf7b64L19 are directly related to this change.

This does mean that if Azure Accounts isn't updated, the extension won't work.  Since Azure Accounts is releasing with a new versioned API, I could check that version to throw an error for the user or I could just check the extension version and validate it.

---
## Delete KubeEnvironment

The language is still Container Apps environment as I mentioned.  If you try to delete an env that has container apps, it throws an error so to get around that, I prompt the user to delete all container apps first.  There was some feedback that it would be better to have to type the name of the container app environment before deleting since it is a pretty destructive event, so I pulled RG extension's implementation into here including the setting to change it to a button click.  Two current "issues" are
- The container app tree nodes don't have the spinner indicating they are deleting even though I call `deleteTreeItem`.  I have no idea why the UI doesn't reflect that since all of the `runWithTemporaryDescription` code is in `deleteTreeItemImpl`.
- The children of the node don't cache unless you expand the tree node which effects perf if you try to delete, cancel, and then delete again.  I don't foresee this being a _huge_ problem since I'd imagine most users would either just delete it or expand the environment first to see what apps are under it before deleting it.
---
## Creating KubeEnviornment

This has been changed to automatically create a new log workspace and resource group.  Potential issues right now are location availability between container app environment and log analytic workspaces, but I should be able to compare the available location lists.  It's just not convenient right now because I have to use the staging location to test certain features.

---

## Minor changes
- Fixed Open in Portal bug with Container Apps
- Use `uiUtils.listAllIterator` where possible (some require more parameters that I didn't account for which I can fix in the ui package)

### DIFF
--- a/src/tree/KubeEnvironmentTreeItem.ts
+++ b/src/tree/KubeEnvironmentTreeItem.ts
@@ -21,10 +21,6 @@ import { settingUtils } from "../utils/settingUtils";
 import { treeUtils } from "../utils/treeUtils";
 import { ContainerAppTreeItem } from "./ContainerAppTreeItem";
 import { IAzureResourceTreeItem } from './IAzureResourceTreeItem';
-<<<<<<< HEAD
-=======
-import { AzExtParentTreeItem, AzExtTreeItem, AzureWizard, AzureWizardExecuteStep, AzureWizardPromptStep, DialogResponses, IActionContext, ICreateChildImplContext, LocationListStep, parseError, TreeItemIconPath, VerifyProvidersStep } from "vscode-azureextensionui";
->>>>>>> efefe1438395e2d9f254b82be51b6651da50cdab
 
 export class KubeEnvironmentTreeItem extends AzExtParentTreeItem implements IAzureResourceTreeItem {
     public static contextValue: string = 'kubeEnvironment|azResource';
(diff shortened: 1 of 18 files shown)


bwateratmsft (approved):

src/tree/KubeEnvironmentTreeItem.ts
```diff
@@ -17,9 +17,14 @@ import { createWebSiteClient } from "../utils/azureClients";
 import { getResourceGroupFromId } from "../utils/azureUtils";
 import { localize } from "../utils/localize";
 import { nonNullProp } from "../utils/nonNull";
+import { settingUtils } from "../utils/settingUtils";
 import { treeUtils } from "../utils/treeUtils";
 import { ContainerAppTreeItem } from "./ContainerAppTreeItem";
 import { IAzureResourceTreeItem } from './IAzureResourceTreeItem';
+<<<<<<< HEAD
```
bwateratmsft, line 24: Still a merge conflict here as well

Issue #32, opened 2022-03-14
Add environment variables during deployment


anthonychu:
Maybe ask for the location of a file in .env format. Should be easy to skip this step.

PR #49, opened 2022-03-15
Add a environment var step in container app creation

nturinski:
Fixes https://github.com/microsoft/vscode-azurecontainerapps/issues/32

The UI for the step:
image
image

### DIFF
--- a/src/commands/createContainerApp/EnvironmentVariablesListStep.ts
+++ b/src/commands/createContainerApp/EnvironmentVariablesListStep.ts
@@ -9,7 +9,7 @@ import { localize } from "../../utils/localize";
 import { selectWorkspaceFile } from "../../utils/workspaceUtils";
 import { IContainerAppContext } from "./IContainerAppContext";
 
-const skipForNowLabel: string = '$(clock) Skip for now';
+const skipForNowLabel: string = localize('skipForNow', '$(clock) Skip for now');
 
 export class EnvironmentVariablesListStep extends AzureWizardPromptStep<IContainerAppContext> {
     public async prompt(context: IContainerAppContext): Promise<void> {
--- a/src/commands/createContainerApp/EnvironmentVariablesListStep.ts
+++ b/src/commands/createContainerApp/EnvironmentVariablesListStep.ts
@@ -28,7 +28,7 @@ export class EnvironmentVariablesListStep extends AzureWizardPromptStep<IContain
     }
 
     private async selectEnvironmentSettings(context: IContainerAppContext) {
-        const envFileFsPath: string = await selectWorkspaceFile(context, 'Select a .env file', { filters: { 'env file': ['env'] } }, 'env');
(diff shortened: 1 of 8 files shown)


src/commands/createContainerApp/EnvironmentVariablesListStep.ts
```diff
@@ -0,0 +1,35 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { parse } from "dotenv";
+import { AzExtFsExtra, AzureWizardPromptStep } from "vscode-azureextensionui";
+import { localize } from "../../utils/localize";
+import { selectWorkspaceFile } from "../../utils/workspaceUtils";
+import { IContainerAppContext } from "./IContainerAppContext";
+
+const skipForNowLabel: string = '$(clock) Skip for now';
```
bwateratmsft, line 12: `localize`?

src/commands/createContainerApp/EnvironmentVariablesListStep.ts
```diff
@@ -0,0 +1,35 @@
...
+
+        if (input.label !== skipForNowLabel) {
+            const envData = await this.selectEnvironmentSettings(context);
+            context.environmentVariables = Object.keys(envData).map(name => { return { name, value: envData[name] } });
+        }
+
+    }
+
+    public shouldPrompt(context: IContainerAppContext): boolean {
+        return context.environmentVariables === undefined
```
bwateratmsft, line 27: Nit
```suggestion
        return context.environmentVariables === undefined;
```

bwateratmsft:
Proposal to make this a codicon we can use in the "Skip for now" step:
image

bwateratmsft (approved):

src/commands/createContainerApp/EnvironmentVariablesListStep.ts
```diff
@@ -0,0 +1,35 @@
...
+import { AzExtFsExtra, AzureWizardPromptStep } from "vscode-azureextensionui";
+import { localize } from "../../utils/localize";
+import { selectWorkspaceFile } from "../../utils/workspaceUtils";
+import { IContainerAppContext } from "./IContainerAppContext";
+
+const skipForNowLabel: string = localize('skipForNow', '$(clock) Skip for now');
+
+export class EnvironmentVariablesListStep extends AzureWizardPromptStep<IContainerAppContext> {
+    public async prompt(context: IContainerAppContext): Promise<void> {
+        const input = await context.ui.showQuickPick([{ label: localize('set', 'Set with environment variable file') }, { label: skipForNowLabel }],
```
anthonychu, line 16: Should this be “environment variable file” or “environment variables file”? At first glance I thought it might be “variables” because it tells me I can have more than 1 variable in there but not sure about grammatically.

nturinski: I think you're correct, but it might actually be better to just go with `.env file` ala https://www.ibm.com/docs/en/aix/7.2?topic=files-env-file

Issue #47, opened 2022-03-14
Error: ContainerRegistryClientOptions.audience must be set to initialize ContainerRegistryClient


anthonychu:
Not sure if this was just a problem with my local build of the extension or if it's a real problem. Got this error when deploying from an ACR.

![image](https://user-images.githubusercontent.com/3982077/158113123-448ad3f0-4432-47b6-99b9-73a42d2c11d2.png)

anthonychu:
Still getting it with the latest build from main I got from @fiveisprime.

![image](https://user-images.githubusercontent.com/3982077/160019845-1fd2ea15-d04a-4f39-8808-08c9d7cd53de.png)

Not getting much more info in the output:

```
3:20:40 PM: Error: ContainerRegistryClientOptions.audience must be set to initialize ContainerRegistryClient.
```

PR #52, opened 2022-03-29
Fix ADAL tokens, change UI for single/multiple revisions

nturinski:
Fixes https://github.com/microsoft/vscode-azurecontainerapps/issues/47

If the Container App is in single revision mode, the revision tree item shouldn't be displayed because it's just confusing.  Based on whether or not it's enabled, where the user can swap RevisionMode context action exists on different tree items (which seems like it could be potentially confusing).

The wonky stuff I'm doing with the credentials... Basically, the T2 package for registry will ignore the credentials and options that I pass it if `signRequest` is defined.  Since that command is only for backwards compatibility, the easiest way to fix this is to just make it undefined when we need to use it.   Since `parseContext` returns a new object, messing with credentials shouldn't delete `signRequest` everywhere.

### DIFF
--- a/main.js
+++ b/main.js
@@ -16,10 +16,10 @@ let perfStats = {
 
 Object.defineProperty(exports, "__esModule", { value: true });
 
-const extension = require('./out/src/extension');
+const extension = require('./dist/extension.bundle');
 
 async function activate(ctx) {
-    return await extension.activateInternal(ctx, perfStats, true /* ignoreBundle */);
+    return await extension.activateInternal(ctx, perfStats);
 }
 
 async function deactivate(ctx) {
--- a/package.nls.json
+++ b/package.nls.json
@@ -10,7 +10,7 @@
     "containerApps.browse": "Browse",
     "containerApps.openInPortal": "Open in Portal",
     "containerApps.createContainerApp": "Create Container App...",
-    "containerApps.deployImage": "Deploy Image to Container App...",
+    "containerApps.deployImage": "Update Image in Container App...",
     "containerApps.deleteContainerApp": "Delete Container App...",
     "containerApps.openLogs": "Open Logs",
     "containerApps.disableIngress": "Disable Ingress for Container App..",
--- a/src/commands/chooseRevisionMode.ts
+++ b/src/commands/chooseRevisionMode.ts
@@ -8,30 +8,35 @@ import { ProgressLocation, window } from "vscode";
 import { IActionContext, IAzureQuickPickItem } from "vscode-azureextensionui";
 import { RevisionConstants } from "../constants";
 import { ext } from "../extensionVariables";
+import { ContainerAppTreeItem } from "../tree/ContainerAppTreeItem";
 import { RevisionsTreeItem } from "../tree/RevisionsTreeItem";
 import { createContainerAppsAPIClient } from "../utils/azureClients";
 import { localize } from "../utils/localize";
 import { nonNullValue } from "../utils/nonNull";
 
-export async function chooseRevisionMode(context: IActionContext, node?: RevisionsTreeItem): Promise<void> {
+export async function chooseRevisionMode(context: IActionContext, node?: ContainerAppTreeItem): Promise<void> {
     if (!node) {
-        node = await ext.tree.showTreeItemPicker<RevisionsTreeItem>(RevisionsTreeItem.contextValue, context);
+        node = await ext.tree.showTreeItemPicker<ContainerAppTreeItem>(new RegExp(ContainerAppTreeItem.contextValue), context);
+    }
+
+    if (node instanceof RevisionsTreeItem) {
+        node = node.parent;
     }
 
     const picks: IAzureQuickPickItem<string>[] = [RevisionConstants.single, RevisionConstants.multiple];
     const placeHolder = localize('chooseRevision', 'Choose revision mode');
 
-    const currentModeQp = picks.find(mode => mode.data === node?.parent.data.configuration?.activeRevisionsMode?.toLowerCase());
-    currentModeQp ? currentModeQp.description += localize('current', ' (current)') : undefined;
+    const currentModeQp = picks.find(mode => mode.data === node?.data.configuration?.activeRevisionsMode?.toLowerCase());
+    currentModeQp ? currentModeQp.detail = localize('current', ' current') : undefined;
 
     const result = await context.ui.showQuickPick(picks, { placeHolder, suppressPersistence: true });
 
     if (currentModeQp !== result) {
         // only update it if it's actually different
         const appClient: ContainerAppsAPIClient = await createContainerAppsAPIClient([context, node]);
-        const containerAppEnvelope = await node.parent.getContainerEnvelopeWithSecrets(context);
-        const updating = localize('updatingRevision', 'Updating revision mode of "{0}" to "{1}"...', node.parent.name, result.label.toLowerCase());
-        const updated = localize('updatedRevision', 'Updated revision mode of "{0}" to "{1}".', node.parent.name, result.label.toLowerCase());
+        const containerAppEnvelope = await node.getContainerEnvelopeWithSecrets(context);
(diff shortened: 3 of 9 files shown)


package.nls.json
```diff
@@ -10,7 +10,7 @@
     "containerApps.browse": "Browse",
     "containerApps.openInPortal": "Open in Portal",
     "containerApps.createContainerApp": "Create Container App...",
-    "containerApps.deployImage": "Deploy Image to Container App...",
+    "containerApps.deployImage": "Update Image in Container App...",
```
wwlorey, line 13: What was the reasoning for this verbiage change?

wwlorey (commented):
> The wonky stuff I'm doing with the credentials... Basically, the T2 package for registry will ignore the credentials and options that I pass it if signRequest is defined.

Did you file an issue on the SDK for this? This is a fine workaround for now but I could see other extensions having to do this same workaround.

I'm wondering if we made a mistake in keeping signRequest around for backwards compatibility 🤔  If this is just a bug on the SDK we're likely fine but maybe their behavior is intentional.

nturinski:
> Did you file an issue on the SDK for this? This is a fine workaround for now but I could see other extensions having to do this same workaround.

Not yet, but I get the impression that it was pretty intentional.  They use a `coreAuth.isTokenCredential` command to check if signRequest is defined or not.  But this SDK also hard codes the default scope (with no option to add more) so I'm not really sure what they're trying to enforce here.

> I'm wondering if we made a mistake in keeping signRequest around for backwards compatibility 🤔 If this is just a bug on the SDK we're likely fine but maybe their behavior is intentional.

I don't think so.  I can imagine the backlash of a bunch of users who got our token and were upset that the 0.10.0 update _also_ broke all T1 SDK users which I imagine is practically everybody besides us.  As far as I can tell, the access token is the exact same so Azure really shouldn't care as long as it's only using the default scope.

nturinski:
I do think that the ADAL token not working is a bug though.

```
if (coreAuth.isTokenCredential(credentialOrOptions)) {
            credential = credentialOrOptions;
            options = clientOptions;
        }
        else {
            options = credentialOrOptions !== null && credentialOrOptions !== void 0 ? credentialOrOptions : {};
        }
```

The SDK has a path for when the credential isn't a `TokenCredential` which I assume is only when it's a `ServiceClientCredential`.  But when it goes through this path, it never sets `credential` so it has no auth.  Also, options passed are ignored because it assumes the credential object are the options... so other parts break.  This part seems like a bug to me which I'll file.

package.nls.json
nturinski: The command kind of changed.  It used to add a new image to the container app.  Each container app could have 4 images running simultaneously, so the verb deployed made more sense.

However, we decided that's a pretty edge use case, so we are now removing the old image and replacing it with the new image, so it functions more like an update.

wwlorey:
> I don't think so. I can imagine the backlash of a bunch of users who got our token and were upset that the 0.10.0 update also broke all T1 SDK users which I imagine is practically everybody besides us. As far as I can tell, the access token is the exact same so Azure really shouldn't care as long as it's only using the default scope.

There are ways of avoiding fallout from removing `signRequest` especially now that we have a versioned API. But it's easier to leave it in and I think we should especially since it sounds like this is an SDK bug

src/tree/DaprTreeItem.ts
```diff
@@ -22,7 +22,7 @@ export class DaprTreeItem extends AzExtTreeItem implements IAzureResourceTreeIte
         super(parent);
         this.label = localize('dapr', 'Dapr');
         this.data = data || {};
-        this.description = this.parent.data.template?.dapr ? 'Enabled' : 'Disabled';
+        // this.description = this.parent.data.template?.dapr ? 'Enabled' : 'Disabled';
```
wwlorey, line 25: Why is the description excluded here?

src/tree/ContainerAppTreeItem.ts
```diff
@@ -143,18 +148,28 @@ export class ContainerAppTreeItem extends AzExtParentTreeItem implements IAzureR
...
-            method: 'POST',
-            queryParameters: { 'api-version': '2021-03-01' },
-            pathTemplate: `${this.id}/listSecrets`,
-        };
-
-        const response = await sendRequestWithTimeout(context, options, 5000, this.subscription.credentials);
-        // if 204, needs to be an empty []
-        concreteContainerAppEnvelope.configuration.secrets = response.status === 204 ? [] : <Secret[]>response.parsedBody;
+
+        // bug that returns 204 if there are no secrets
```
wwlorey, line 152: Has a bug been filed on the SDK for this?

nturinski:
> > I don't think so. I can imagine the backlash of a bunch of users who got our token and were upset that the 0.10.0 update also broke all T1 SDK users which I imagine is practically everybody besides us. As far as I can tell, the access token is the exact same so Azure really shouldn't care as long as it's only using the default scope.
> 
> There are ways of avoiding fallout from removing `signRequest` especially now that we have a versioned API. But it's easier to leave it in and I think we should especially since it sounds like this is an SDK bug

Yeah, that's true.  We could have two separate token responses and users can pick which one that they need.  I'm not opposed to doing that either and it's not too late since users wouldn't know that we did that unless they had already adopted T2 packages.

We would just have to change all of our extensions to use this new credentials2 that we return.

src/tree/DaprTreeItem.ts
nturinski: They changed where the dapr stuff lives now and I was too lazy to figure that out atm and fix it properly.

If they're going to change it on me, I'll just remove the tree entirely! (which I did.)

src/tree/ContainerAppTreeItem.ts
nturinski: It does now 😐

src/commands/chooseRevisionMode.ts
```diff
@@ -8,45 +8,51 @@ import { ProgressLocation, window } from "vscode";
...
 import { RevisionConstants } from "../constants";
 import { ext } from "../extensionVariables";
+import { ContainerAppTreeItem } from "../tree/ContainerAppTreeItem";
 import { RevisionsTreeItem } from "../tree/RevisionsTreeItem";
 import { createContainerAppsAPIClient } from "../utils/azureClients";
 import { localize } from "../utils/localize";
 import { nonNullValue } from "../utils/nonNull";
 
-export async function chooseRevisionMode(context: IActionContext, node?: RevisionsTreeItem): Promise<void> {
+export async function chooseRevisionMode(context: IActionContext, node?: ContainerAppTreeItem): Promise<void> {
```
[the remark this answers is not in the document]
nturinski: sigh.... yeeeeeaah.  But I changed it because in this scope:

image

 Typescript forgets that I already determined that this is a `ContainerAppItem` and throws typing errors.

package.nls.json
nturinski: Would you be okay with holding off a bit? @anthonychu and I are still kind of trying to figure out the best way to convey what this command does.  I was planning to rename everything (the file name, localize labels, etc.) but I didn't want to do that multiple times so was just waiting until we landed on something

wwlorey: Sure thing. Definitely not a blocker for this 👍

src/tree/DaprTreeItem.ts
wwlorey: Got it. I'm not sure how you're tracking issues like this one but it might be worth logging a GitHub issue on this repo  about dapr so comments like this don't get lost in the sauce

main.js
```diff
@@ -16,10 +16,10 @@ let perfStats = {
 
 Object.defineProperty(exports, "__esModule", { value: true });
 
-const extension = require('./out/src/extension');
+const extension = require('./dist/extension.bundle');
 
 async function activate(ctx) {
-    return await extension.activateInternal(ctx, perfStats, true /* ignoreBundle */);
+    return await extension.activateInternal(ctx, perfStats);
```
wwlorey, line 22: nit: revert this file

wwlorey (approved):

PR #53, opened 2022-03-30
Install the new sdk

nturinski:
Forgot to PR this one, but just using the new SDK that was published

### DIFF
--- a/package.json
+++ b/package.json
@@ -359,6 +359,7 @@
         "webpack-cli": "^4.6.0"
     },
     "dependencies": {
+        "@azure/arm-app": "^1.0.0-beta.1",
         "@azure/arm-containerregistry": "^8.1.1",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -368,7 +369,7 @@
         "dotenv": "^16.0.0",
         "open": "^8.0.4",
         "semver": "^7.3.5",
-        "vscode-azureextensionui": "file:../vscode-azuretools/ui/vscode-azureextensionui-0.50.0.tgz",
+        "vscode-azureextensionui": "^0.50.0",
```
bwateratmsft, line 372: This is the newish old SDK, do we want to be using the split one instead?

bwateratmsft (approved):

package.json
nturinski: Yeah, I'm working on that on another branch now

PR #58, opened 2022-04-04
Browse after revision update

nturinski:
Also marked the "@azure/container-registry": "1.0.0-beta.5" package.  There's a change in 1.0.0 that causes ADAL tokens to not work again.  This is a temporary fix and I'll file an issue against the sdk

### DIFF
--- a/src/tree/ContainerAppTreeItem.ts
+++ b/src/tree/ContainerAppTreeItem.ts
@@ -63,7 +63,7 @@ export class ContainerAppTreeItem extends AzExtParentTreeItem implements IAzureR
             children.push(this.revisionsTreeItem);
         }
 
-        this.data.configuration?.ingress ? children.push(new IngressTreeItem(this, this.data.configuration?.ingress)) : children.push(new IngressDisabledTreeItem(this));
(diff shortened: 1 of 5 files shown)


src/tree/ContainerAppTreeItem.ts
```diff
@@ -78,11 +78,11 @@ export class ContainerAppTreeItem extends AzExtParentTreeItem implements IAzureR
 
     public async browse(): Promise<void> {
         // make sure that ingress is enabled
-        if (!this.data.latestRevisionFqdn) {
+        if (this.ingressEnabled()) {
             throw new Error(localize('enableIngress', 'Enable ingress to perform this action.'));
         }
 
-        await openUrl(`https://${this.data.latestRevisionFqdn}`);
+        await openUrl(`https://${this.data.configuration?.ingress?.fqdn}`);
```
bwateratmsft, line 85: If any of this chain is `undefined`, it'll try to open `https://` which is probably going to have weird results. If we are confident that the parts can never be `undefined`, I'd use `!` to assert it; otherwise I'd guard this with an `if`.

bwateratmsft (approved):

Issue #76, opened 2022-04-06
There is no command "Open in Portal" in the context menu of the container app environment


v-xinda:
**OS:** All
**Build Version:** 20220405.6
**Regression:** Not a regression

**Repro Steps:**
1. F1 -> Azure Container Apps: Open in Portal -> Select a subscription -> Select a container app environment -> Open the container app environment in portal.
![image](https://user-images.githubusercontent.com/57662131/161958183-2e18b944-8cb9-4ab2-83b8-e352c4a5b998.png)
2. Right click the container app environment.
3. Check whether there is the command "Open in Portal" in the context menu.

**Expect:**
There is the command "Open in Portal" in the context menu.

**Actual:**
There is no command "Open in Portal" in the context menu.
![image](https://user-images.githubusercontent.com/57662131/161958100-9a69a69a-f0a1-403d-9e40-eae70e73432f.png)

Issue #74, opened 2022-04-06
"Resource not found" shows in portal after clicking the "Logs" node


v-ruizh:
**OS:** All
**Build Version:** 20220405.6
**Regression:** Not a regression

**Repro Steps:**
1. Create a container app.
2. Click "Logs" node in this container app.
3. Check whether opens the log in portal or not. 

**Expect:**
Open the log in portal successfully.

**Actual:**
"Resource not found" shows in portal.
![image](https://user-images.githubusercontent.com/45614674/161962177-0b0120c5-b6a2-4b7f-99e5-0ab7d3029c97.png)

PR #91, opened 2022-04-07
Fix open in portal for environment and logs

nturinski:
Fixes https://github.com/microsoft/vscode-azurecontainerapps/issues/76
Fixes https://github.com/microsoft/vscode-azurecontainerapps/issues/74


package.json
```diff
@@ -266,7 +266,7 @@
                 },
                 {
                     "command": "containerApps.openInPortal",
-                    "when": "view == containerApps && viewItem =~ /ManagedEnvironment|containerApp/",
+                    "when": "view == containerApps && viewItem =~ /managedEnvironment|containerApp/",
```
bwateratmsft, line 269: Do the regex flags work here? e.g. would this work?
```suggestion
                    "when": "view == containerApps && viewItem =~ /managedEnvironment|containerApp/i",
```

bwateratmsft (approved):

src/commands/openInPortal.ts
```diff
@@ -16,7 +17,10 @@ export async function openInPortal(context: IActionContext, node?: AzExtTreeItem
     if (node instanceof ContainerAppTreeItem) {
         // ContainerApp's id don't include the ManagedEnvironment (ie the parent) so don't use fullId
         await azUtil.openInPortal(node, <string>node.data.id);
-    } else {
+    } else if (node instanceof LogsTreeItem) {
+        await azUtil.openInPortal(node, `${node.parent.id}/logs`)
```
wwlorey, line 21: I think it would be better to appropriately set each tree item's id in the tree item class rather than add special cases here

package.json
nturinski: Actually, it does.  Neat.

src/commands/openInPortal.ts
nturinski: Some context-- a lot of azure resource IDs build on each other (such as web app/slots or static app/environment build) so we were using full id for most of those cases.  Container Apps only has 3 resources to open in portal and container apps do not include their environment as part of the Azure ID.  Anyway, the ids were already properly configured to work in the portal so this change should be good

wwlorey (approved):

PR #103, opened 2022-04-12
Add activate/deactivate/restart to container apps in single revision mode

nturinski:
All these commands are revision specific, so when we removed the "revisions" node, users could no longer run any of these commands on containers in single revision.

Single revision:
image

Multiple revision:
image

Also removed the tool tip that used to be there because it was more for dev purposes.


src/commands/revisionCommands/changeRevisionActiveState.ts
```diff
@@ -4,18 +4,22 @@
...
-export async function changeRevisionActiveState(context: IActionContext, command: 'activate' | 'deactivate' | 'restart', node?: RevisionTreeItem): Promise<void> {
+export async function changeRevisionActiveState(context: IActionContext, command: 'activate' | 'deactivate' | 'restart', node?: ContainerAppTreeItem | RevisionTreeItem): Promise<void> {
     if (!node) {
-        node = await ext.tree.showTreeItemPicker<RevisionTreeItem>(RevisionTreeItem.contextValue, context);
+        node = await ext.tree.showTreeItemPicker<ContainerAppTreeItem | RevisionTreeItem>(ContainerAppTreeItem.contextValue, context);
     }
 
+    const containerAppName: string = node instanceof RevisionTreeItem ? node.parent.parent.name : node.name;
+    const revisionName: string = node instanceof RevisionTreeItem ? node.name : nonNullProp(node.data, 'latestRevisionName');
+    const resourceGroupName: string = node instanceof RevisionTreeItem ? node.parent.parent.resourceGroupName : node.resourceGroupName;
```
wwlorey, line 21: I think it would be cleaner to do something like this:

```ts
let containerAppName: string;
...
if (node instanceof RevisionTreeItem) {
    containerAppName = ...
    ...
} else {
    ...
}
```

But I know you're trying to release and this isn't a blocker.

wwlorey (approved):

src/commands/revisionCommands/changeRevisionActiveState.ts
nturinski: I actually went back and forth on it.  That way is more clear what is happening, but it takes like 3x more lines of code XD

Issue #43, opened 2022-03-14
Support other anonymous registries


anthonychu:
Support registries other than Docker Hub and ACR.

- Allow freeform entry of container image name (supports anonymous images from non-Docker Hub registries like MCR)
- Ideally the next time we update the app, the image is defaulted to the previous value

PR #114, opened 2022-05-18
Add RegistryImageInputStep so people can just paste the url of image

nturinski:
Fixes https://github.com/microsoft/vscode-azurecontainerapps/issues/43

An example of the image name with tag is `mcr.microsoft.com/azuredocs/containerapps-helloworld:latest`


src/commands/createContainerApp/ContainerAppCreateStep.ts
```diff
@@ -63,6 +63,9 @@ export class ContainerAppCreateStep extends AzureWizardExecuteStep<IContainerApp
         progress.report({ message: creatingSwa });
         ext.outputChannel.appendLog(creatingSwa);
 
+        context.image ||= `${getLoginServer(context)}/${context.repositoryName}:${context.tag}`;
+        const name = context.image.substring(context.image.lastIndexOf('/') + 1).replace(':', '-');
```
bwateratmsft, line 67: Just to double check the logic here--some full image names will have several slashes even after the registry, e.g.:

`mcr.microsoft.com/dotnet/aspnet:5.0`

In that, the registry is `mcr.microsoft.com`, the repository is `dotnet/aspnet`, and the tag `5.0`. The name chosen would be `aspnet-5.0`, instead of `dotnet/aspnet-5.0`.

I'm not saying this is wrong, just want to make sure this is the behavior you are aiming for.

src/commands/deployImage/RegistryImageInputStep.ts
```diff
@@ -0,0 +1,25 @@
...
+
+import { AzureWizardPromptStep } from "@microsoft/vscode-azext-utils";
+import { localize } from "../../utils/localize";
+import { IContainerAppContext } from "../createContainerApp/IContainerAppContext";
+
+export class RegistryImageInputStep extends AzureWizardPromptStep<IContainerAppContext> {
+    public async prompt(context: IContainerAppContext): Promise<void> {
+        const prompt: string = localize('registryImagePrompt', 'Enter the container image with tag');
+        const placeHolder: string = localize('registryImagePlaceHolder', 'For example: `mcr.microsoft.com/azuredocs/containerapps-helloworld:latest`')
+        context.image = (await context.ui.showInputBox({
```
bwateratmsft, line 14: No validation needed, or just not worth it ATM? Maybe put in a `// TODO` comment to note it?

nturinski: Yeah, talking to Anthony about it, he said he didn't feel like it was needed.  The error you get back from container apps isn't great, but good enough

```
The following field(s) are either invalid or missing. Invalid value: "mcr.microsoft.com": Invalid Name: Per RFC 1123 name must consist of lower case alphanumeric characters or '-', and must start and end with an alphanumeric character (e.g. 'my-name', or '123-abc', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?'): template.containers.
```

Also, people tend to just copy/paste these image names anyway

src/commands/createContainerApp/ContainerAppCreateStep.ts
nturinski: Yeah, the name doesn't really matter, it just has to be unique to that container so just wanted to grab something + the tag to make it unique.

Issue #118, opened 2022-05-20
There is no validation when editing scale rule setting with "0-0" range


v-mengwe:
**OS:** Win10/M1
**Build Version:** 20220520.7
**Regression:** Not a Regression

**Repro Steps:**

1. Expand a container app -> Right click the "Scaling" -> Edit Scale Rule Setting -> Type "0-0" for the range.
2. Check whether there is a validation when typing "0-0".

**Expect:**
There is a validation when typing "0-0".

**Actual:**
1. There is no validation when typing "0-0".
    <img width="601" alt="Screen Shot 2022-05-20 at 17 39 05" src="https://user-images.githubusercontent.com/62053665/169500771-c8ba08b2-45d5-490e-a6c6-10142b830972.png">
2. Pop up an error when editing scale rule setting with "0-0" range.
    <img width="450" alt="Screen Shot 2022-05-20 at 17 40 26" src="https://user-images.githubusercontent.com/62053665/169501013-3527c0aa-1280-4bbc-8f43-ffd585c91088.png">

PR #125, opened 2022-05-20
Check that max replicas is greater than 0

nturinski:
Fixes https://github.com/microsoft/vscode-azurecontainerapps/issues/118

### DIFF
--- a/src/commands/scaling/editScalingRange.ts
+++ b/src/commands/scaling/editScalingRange.ts
@@ -56,6 +56,8 @@ async function validateInput(range: string | undefined): Promise<string | undefi
         return localize('maxRangeExceeded', 'The maximum number of replicas is 10.');
     } else if (min > max) {
         return localize('minHigherThanMax', 'The minimum range cannot be larger than the maximum range.');
+    } else if (max === 0) {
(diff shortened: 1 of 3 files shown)


src/commands/scaling/editScalingRange.ts
```diff
@@ -56,6 +56,8 @@ async function validateInput(range: string | undefined): Promise<string | undefi
         return localize('maxRangeExceeded', 'The maximum number of replicas is 10.');
     } else if (min > max) {
         return localize('minHigherThanMax', 'The minimum range cannot be larger than the maximum range.');
+    } else if (max === 0) {
```
bwateratmsft, line 59: Should we try to handle negative numbers or rely on the `formatRegex` to take care of that?

bwateratmsft (approved):

src/commands/scaling/editScalingRange.ts
nturinski: The format regex should take care of it since it has to start with a digit and there can only be one "-" before the next digit.

Side node: Why does node.js use RegExp as shorthand for regular expression when the rest of the world uses regex........

Issue #122, opened 2022-05-20
Fail to update image with "Other public registry" in container app


v-mengwe:
**OS:** Win10/M1
**Build Version:** 20220520.7
**Regression:** Not a Regression

**Repro Steps:**

1. Right click a container app -> Update Image in Container App -> Select "Other public registry" -> Type "hello-world:latest" -> Skip for now.
2. Check whether succeeds to update image with "Other public registry".

**Expect:**
Succeed to update image with "Other public registry".

**Actual:**
Fail to update image with "Other public registry".
<img width="444" alt="Screen Shot 2022-05-20 at 19 07 22" src="https://user-images.githubusercontent.com/62053665/169515519-8ebff478-2987-4dab-b601-b84990390fd2.png">

Issue #19, opened 2022-01-27
Error creating new Container App


BigMorty:
**Repro Steps**
- Right clicked on my new CA env I just created and selected 'Create Container App...'
- Name - 'first-app'
- Cont Reg - ACR - 'mikemortACR'
- Repository - 'orderproc'
- Tag = '3.0'
- HTTP endpoint - 'Enable'
- Traffic - 'External'
- Port - '80'

**Received the following error...**
Error: The following field(s) are either invalid or missing. Invalid value: "orderproc-3.0": Invalid Name: Per RFC 1123 name must consist of lower case alphanumeric characters or '-', and must start and end with an alphanumeric character (e.g. 'my-name',  or '123-abc', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?'): template.containers.

I tried a few other attempts, the pattern I see is it fails if the tag I select is something like 3.0 or 2.2. If I select a tag without a period in it, it works.

anthonychu:
Likely because the `.` creates an invalid hostname.

Also need to verify other name requirements. For instance, there's a limit of 2–20 characters enforced in VS Code but longer names are allowed.

PR #127, opened 2022-05-20
Correctly parse container image name and fix name regex

nturinski:
Fixes https://github.com/microsoft/vscode-azurecontainerapps/issues/122
Fixes https://github.com/microsoft/vscode-azurecontainerapps/issues/19

`.` are common in tags, but not allowed for container app names.  Since this is automatically generated, it was blocking people from creating with certain tags.

### DIFF
--- a/src/commands/createContainerApp/ContainerAppCreateStep.ts
+++ b/src/commands/createContainerApp/ContainerAppCreateStep.ts
@@ -64,7 +64,7 @@ export class ContainerAppCreateStep extends AzureWizardExecuteStep<IContainerApp
         ext.outputChannel.appendLog(creatingSwa);
 
         context.image ||= `${getLoginServer(context)}/${context.repositoryName}:${context.tag}`;
-        const name = context.image.substring(context.image.lastIndexOf('/') + 1).replace(':', '-');
+        const name = context.image.substring(context.image.lastIndexOf('/') + 1).replace(/[^0-9a-zA-Z-]/g, '-');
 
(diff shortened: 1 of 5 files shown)


src/commands/createContainerApp/ContainerAppCreateStep.ts
```diff
@@ -64,7 +64,7 @@ export class ContainerAppCreateStep extends AzureWizardExecuteStep<IContainerApp
         ext.outputChannel.appendLog(creatingSwa);
 
         context.image ||= `${getLoginServer(context)}/${context.repositoryName}:${context.tag}`;
-        const name = context.image.substring(context.image.lastIndexOf('/') + 1).replace(':', '-');
+        const name = context.image.substring(context.image.lastIndexOf('/') + 1).replace(/[^0-9a-zA-Z-]/g, '-');
```
bwateratmsft, line 67: I think this should be a function, since this regex is repeated twice, and they should always be kept in sync (right?).

bwateratmsft: Maybe something like `getContainerNameForImage(image: string): string`?

nturinski: Yeah, good point.  I thought the same thing, but figured I'd just overhaul this entire extension whenever we do the unified work.  This one has become a mess because of some last minute pivots in design 😬

bwateratmsft: If it's helpful, we have some similar [stuff](https://github.com/microsoft/vscode-docker/blob/main/src/utils/getValidImageName.ts) in the Docker extension that has been stable for years.

bwateratmsft (approved):

PR #136, opened 2022-06-25
Refactor TreeItems to Resource and provide to ResourceGroups

nturinski:
I tried to model this after what we have been discussing with the ResourceGroup Extension.  I made a mock BranchProvider that would accept the `ContainerAppsExtResourceBase` as the data model rather than the `AzExtTreeItem`. I ultimately have to make tree items anyway to make it compatible with RG's current implementation, but the idea would to get rid of that at some point and pass up the branch provider to RG rather than resolving all of the apps.


bwateratmsft:
> I ultimately have to make tree items anyway to make it compatible with RG's current implementation

I think we definitely need to do the upgrades to the RG extension that @philliphoff is working on beforehand

nturinski:
> > I ultimately have to make tree items anyway to make it compatible with RG's current implementation
> 
> I think we definitely need to do the upgrades to the RG extension that @philliphoff is working on beforehand

Definitely.  This refactor is still relevant though because it does plug CA into RG and we do want to release that sooner rather than later.

alexweininger:
I'm getting this when the extension is installed. We need to add the `x-azResources` property to the package json with at least the activation events ([example](https://github.com/microsoft/vscode-azurestaticwebapps/blob/90c29ddf3ed6c45a324af73c320ad4d4a7b15bc5/package.json#L64)).

image

src/utils/createTreeItemsWithErrorHandling.ts
```diff
@@ -0,0 +1,60 @@
...
+): Promise<AzExtTreeItem[]> {
+
+    const treeItems: AzExtTreeItem[] = [];
+    let lastUnknownItemError: unknown;
+    sourceArray ||= [];
+    await Promise.all(sourceArray.map(async (source: TSource) => {
+        try {
+            const item: AzExtTreeItem | undefined = await createTreeItem(source);
+            if (item) {
+                // Verify at least the following properties can be accessed without an error
```
bwateratmsft, line 20: To do?

src/tree/IngressResource.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    constructor(data: Ingress | undefined, containerApp: ContainerAppResource, parentId: string) {
+        super(containerApp);
+        this.data = data ?? {};
+        this.id = `${parentId}/${this.idSuffix}`
+
+        this.label = label;
+        this.description = data ? undefined : localize('disabled', 'Disabled');
+
+        this.isParent = !!data;
```
bwateratmsft, line 34: I was thinking some about the quick pick logic, and I was thinking--should we instead do `isLeaf`? i.e. is it more interesting whether something actually has children, or whether it's even possible for it to have children?

src/tree/ContainerAppsBranchDataProvider.ts
```diff
@@ -0,0 +1,49 @@
...
+
+import { AzExtParentTreeItem, AzExtTreeItem, callWithTelemetryAndErrorHandling, IActionContext } from "@microsoft/vscode-azext-utils";
+import { TreeDataProvider, TreeItem } from "vscode";
+import { ContainerAppsExtResourceBase } from "../resolver/ContainerAppsExtResourceBase";
+import { createTreeItemsWithErrorHandling } from "../utils/createTreeItemsWithErrorHandling";
+import { ContainerAppExtParentTreeItem } from "./ContainerAppExtParentTreeItem";
+import { ContainerAppExtTreeItem } from "./ContainerAppExtTreeItem";
+
+export class ContainerAppsBranchDataProvider implements Partial<TreeDataProvider<ContainerAppsExtResourceBase<unknown>>> {
+    public getTreeItem(treeItem: ContainerAppsExtResourceBase<unknown>): TreeItem {
```
bwateratmsft, line 14: Probably needs `collapsibleState` too right?

src/tree/ContainerAppExtParentTreeItem.ts
```diff
@@ -0,0 +1,39 @@
...
+    public label: string;
+    public contextValue: string;
+    // ideally this would be the branch provider, but that won't work right now
+    treeDataProvider = ext.rgApi.appResourceTree;
+
+    public resource: T;
+
+    constructor(parent: AzExtParentTreeItem | undefined, resource: T) {
+        super(parent);
+        this.label = resource.label;
```
bwateratmsft, line 20: This section where we copy all the properties in a constructor gets repeated a lot. We should try to refactor in such a way that uses generics or base classes to deal with these properties just once.

src/tree/IngressResource.ts
nturinski: I'm pretty impartial to either approach. I just went with `isParent` because if I leave it undefined, it'll default to `false`. My thought was that leaf nodes are more common than parents, but I feel like both convey the same information if you just ! it

src/tree/ContainerAppsBranchDataProvider.ts
nturinski: Yeah though right now, this branchProvider functionally doesn't do anything except call `getChildren`. I figured once the RG implementation was done, I'd go back and implement it those required functions.

src/tree/ContainerAppExtParentTreeItem.ts
nturinski: I'm having a pretty tough time thinking up a way to refactor this, namely because one needs to extend from `AzExtTreeItem` and the other has to extend from `AzExtParentTreeItem`.

If I refactor them into a base class, I can't have the parent nodes extend that base class because it won't have `AzParentTreeItem`  which leaves me having to implement it.

Maybe just a util function that does all of that initializing so there's less repeated code?

bwateratmsft: It's in a lot more than two places (or at least, it was when I put the comment), right? Just two would be much better.

Issue #21, opened 2022-01-27
Not all tags are shown for a repo


BigMorty:
- Right click on existing app and select 'Deploy image to Container App'
- Registry - 'ACR' - 'mikemortACR'
- Repository - Choose one with multiple tags

**Issue: Only most recent tag is shown** - Should see 4 tags for this repo

![image](https://user-images.githubusercontent.com/947510/151305254-7fc30443-ee4d-485e-9c08-37dcd1053a73.png)

nturinski:
Heeey @BigMorty!

Do you happen to know if the other tags have some invalid characters or some other pattern (maybe associated with a different repo or anything else) that could be causing this behavior?  I haven't run into this myself and have registries with multiple tags, so just wondering if you had any more insight.

Could you list the other tags for me?

BigMorty:
Here is an example of what I see in the Docker explorer for the orderproc image in my ACR:
![image](https://user-images.githubusercontent.com/947510/153108169-68d912e5-9aea-41af-9e83-b8a1b5d6c3cf.png)

Here is what I see when I go to create a container app using orderproc:
![image](https://user-images.githubusercontent.com/947510/153108220-b6b04dfe-98b1-4065-a420-131454159269.png)

PR #137, opened 2022-07-05
List all tags in all manifests and display time it was created

nturinski:
Fixes https://github.com/microsoft/vscode-azurecontainerapps/issues/21

Turns out the problem was the response of the manifest is grouped by the updated date. Previously, I was always grabbing the latest (index 0) manifest and only showing its tags.  Now I reduce it to one array of tags to display all of the tags.

I also went ahead and added descriptions about the last time it was updated. Seemed useful and is modeled after Docker. I believe the timestamps are different because Docker gets the timestamp from the [Docker repository](https://github.com/microsoft/vscode-docker/blob/8746a82b92abf68ee5ddc5a4137b38ff09504dcd/src/tree/registries/dockerV2/DockerV2RepositoryTreeItem.ts#L67-L72) whereas I'm using the ACR response.

image

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,6 +18,7 @@
                 "@azure/ms-rest-js": "^2.2.1",
                 "@microsoft/vscode-azext-azureutils": "^0.1.3",
                 "@microsoft/vscode-azext-utils": "^0.1.0",
+                "dayjs": "^1.11.3",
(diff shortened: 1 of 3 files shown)


alexweininger (approved):
I like the descriptions

Issue #133, opened 2022-06-16
The Ingress node is still enabled after executing "Disable Ingress for Container App" action


v-ruizh:
**OS:** Win 10
**Build Version:** 20220615.1
**Regression:** Previous Release

**Repro Steps:**
1. Right click Ingress node -> Select "Disable Ingress for Container App".
2. Check whether the Ingress node is disabled or not.

**Expect:**
The Ingress node is disabled.
![image](https://user-images.githubusercontent.com/45614674/174028131-e71713f0-a29c-4455-9bad-54e791cf5c9a.png)

**Actual:**
The Ingress node is still enabled.
![image](https://user-images.githubusercontent.com/45614674/174029181-5629eb7f-2362-4add-86e0-2d8b18323696.png)

**More Info:**
1. Refresh can't fix this issue.
2. This issue does not reproduce on the previous released Container App extension 0.1.1.

Issue #104, opened 2022-04-12
Unable to change revision mode if ingress is disabled


nturinski:
1. Right-click a multiple revision app with ingress disabled
1. Change revision mode...
2. Run into this error `Ingress section of the container app must have a targetPort specified. TargetPort must be in the range of [1,65535].`

PR #153, opened 2022-07-07
Pass in null rather than undefined

nturinski:
Fixes https://github.com/microsoft/vscode-azurecontainerapps/issues/133

Since I'm using a PATCH instead of a POST, the undefined doesn't get picked up and so ingress wasn't being disabled. The request needs to include ingress as null rather than undefined in order for it to be picked up.

Edit: I also noticed that this issue is fixed as well:
Fixes https://github.com/microsoft/vscode-azurecontainerapps/issues/104

Probably unrelated to this and instead because the package updating to use PATCH rather than POST, but will document that here.

### DIFF
--- a/src/commands/ingressCommands.ts
+++ b/src/commands/ingressCommands.ts
@@ -110,7 +110,7 @@ export async function toggleIngressVisibility(context: IActionContext, node?: In
 
 async function updateIngressSettings(context: IActionContext,
     options: {
-        ingress: Ingress | undefined,
+        ingress: Ingress | null,
         node: IngressTreeItem | IngressDisabledTreeItem,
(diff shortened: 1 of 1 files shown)


bwateratmsft (approved):
Maybe add a comment explaining why `null` instead of `undefined`?

bwateratmsft:
> Might be a good thing to include a comment explaining

Lol, me and Alex fifteen seconds apart with the same feedback

bwateratmsft (approved):

nturinski:
Too many issues with undefined vs. null that the SDK was complaining about. Just easier to cast it at the end (when I call the SDK command)

Issue #65, opened 2022-04-06
Fail to create a container app environment with name "ni"


v-xinda:
**OS:** All
**Build Version:** 20220405.6
**Regression:** Not a regression

**Repro Steps:**
1. F1 -> Azure Container Apps: Create Container App Environment... -> Type "ni" as the name.
2. Check whether succeeds to create a container app environment with name "ni".

**Expect:**
Succeed to create a container app environment with name "ni".

**Actual:**
Fail to create a container app environment with name "ni".
![image](https://user-images.githubusercontent.com/57662131/161913624-facd7570-1103-41fa-a27e-121d9de5dc2b.png)

PR #162, opened 2022-07-14
Fix name steps

nturinski:
Fixes https://github.com/microsoft/vscode-azurecontainerapps/issues/65

Also got rid of the `checkNameLength`.  It wasn't working properly and would let invalid names slip in.  Not worth the slight UI enhancements it brought.

### DIFF
--- a/src/commands/createContainerApp/ContainerAppNameStep.ts
+++ b/src/commands/createContainerApp/ContainerAppNameStep.ts
@@ -10,7 +10,6 @@ import { getResourceGroupFromId } from "../../utils/azureUtils";
 import { localize } from "../../utils/localize";
 import { IContainerAppContext } from './IContainerAppContext';
 
-let checkNameLength: boolean = false;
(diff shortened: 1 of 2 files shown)


bwateratmsft (approved):
Does the error go away immediately upon typing the second character, or is it a little delayed? (i.e. is there any perceivable delay)

nturinski:
It goes away immediately for container app names, but for the environment, it won't go away until they have typed 4 characters. It's annoying from a UI perspective, but if I leave checkNameLength as it was, it wouldn't screen names that are too short.

I think what I will end up doing is allowing us to set validation checks for `inputBox.onDidAccept` since currently the validation for `onDidChange` and `onDidAccept` have the same criteria which is unhelpful.

@MicroFish91 had a good idea about putting our API call validation code to run only `onDidAccept` since currently, we will call it every time a keystroke changes.

bwateratmsft:
I like that `onDidAccept` idea. IIRC that's a pretty recent addition to the API.

PR #166, opened 2022-07-21
Add Scale Rules Feature

MicroFish91:
Feature support for adding scale rules.  Currently only supporting `HTTP` and `Azure Queue` rules, no custom.

image


src/commands/scaling/addScaleRule/AddNewScaleRule.ts
```diff
@@ -0,0 +1,86 @@
...
+import { ScaleRuleTypes } from "../../../constants";
+import { ext } from "../../../extensionVariables";
+import { ContainerAppTreeItem } from "../../../tree/ContainerAppTreeItem";
+import { RevisionTreeItem } from "../../../tree/RevisionTreeItem";
+import { ScaleRuleGroupTreeItem } from '../../../tree/ScaleRuleGroupTreeItem';
+import { localize } from "../../../utils/localize";
+import { updateContainerApp } from "../../updateContainerApp";
+import { IAddScaleRuleWizardContext } from "./IAddScaleRuleWizardContext";
+
+export class AddNewScaleRule extends AzureWizardExecuteStep<IAddScaleRuleWizardContext> {
```
nturinski, line 18: nit: Usually we include Step in the class name. Also, I think Add and New are somewhat redundant, so maybe just name it `AddScaleRuleStep`?

src/commands/scaling/addScaleRule/AddNewScaleRule.ts
```diff
@@ -0,0 +1,86 @@
...
+    public priority: number = 100;
+
+    public async execute(context: IAddScaleRuleWizardContext, _progress: Progress<{ message?: string | undefined; increment?: number | undefined }>): Promise<void> {
+        const node: ScaleRuleGroupTreeItem = nonNullProp(context, "treeItem");
+        const containerApp: ContainerAppTreeItem = node.parent.parent instanceof RevisionTreeItem ? node.parent.parent.parent.parent : node.parent.parent;
+
+        const adding = localize('addingScaleRule', 'Adding scale rule setting to "{0}"...', containerApp.name);
```
nturinski, line 25: May be helpful to include the type of scale rule being added.

src/commands/scaling/addScaleRule/AddNewScaleRule.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    private integrateRule(context: IAddScaleRuleWizardContext, scaleRules: ScaleRule[], scaleRule: ScaleRule): void {
+        switch (context.ruleType) {
+            case ScaleRuleTypes.HTTP:
+                const idx: number = scaleRules.findIndex((rule) => rule.http);
+                if (idx !== -1) {
+                    scaleRules[idx] = scaleRule;
+                } else {
+                    scaleRules.push(scaleRule);
```
nturinski, line 77: I think it makes sense to just have either a default case that always pushes the scaleRule in and instead of replacing the HTTP rule by overwriting `scaleRules[ind]`, to just splice it out.  That way you could do something more like

```
switch (context.ruleType) {
            case ScaleRuleTypes.HTTP:
                const idx: number = scaleRules.findIndex((rule) => rule.http);
                if (idx) {
                    scaleRules.splice(idx, 1)
                }
            case ScaleRuleTypes.Queue:
            default:
                   scaleRules.push(scaleRule);
}

```

src/commands/scaling/addScaleRule/GetScaleRuleNameStep.ts
```diff
@@ -0,0 +1,54 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { ContainerApp, ContainerAppsAPIClient, ScaleRule } from '@azure/arm-appcontainers';
+import { AzureWizardPromptStep } from '@microsoft/vscode-azext-utils';
+import { createContainerAppsAPIClient } from '../../../utils/azureClients';
+import { getResourceGroupFromId } from '../../../utils/azureUtils';
+import { localize } from '../../../utils/localize';
+import { IAddScaleRuleWizardContext } from './IAddScaleRuleWizardContext';
+
+export class GetScaleRuleNameStep extends AzureWizardPromptStep<IAddScaleRuleWizardContext> {
```
nturinski, line 13: Following convention, this would just be `ScaleRuleNameStep`.

src/commands/scaling/addScaleRule/GetScaleRuleNameStep.ts
```diff
@@ -0,0 +1,54 @@
...
+    public hideStepCount: boolean = true;
+    containerApp: ContainerApp | undefined;
```
nturinski, line 15: You get the containerApp in multiple places. Seems like you could just save it on the context.

src/commands/scaling/addScaleRule/GetScaleRuleNameStep.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    public shouldPrompt(): boolean {
+        return true;
+    }
+
+    private async initContainerAppClient(context: IAddScaleRuleWizardContext): Promise<void> {
+        try {
+            const client: ContainerAppsAPIClient = await createContainerAppsAPIClient([context, context.treeItem]);
+            const containerAppRg: string = getResourceGroupFromId(context.treeItem.fullId);
+            this.containerApp = await client.containerApps.get(containerAppRg, context.treeItem.parent.parent.label);
```
nturinski, line 33: You should already have this data (it lives in the `ContainerAppTreeItem`) so you shouldn't have to create a new client and get this from Azure.

src/commands/scaling/addScaleRule/GetScaleRuleNameStep.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    public async prompt(context: IAddScaleRuleWizardContext): Promise<void> {
+        context.ruleName = (await context.ui.showInputBox({
+            prompt: localize('scaleRuleNamePrompt', 'Enter a name for the new scale rule.'),
+            validateInput: async (value: string | undefined): Promise<string | undefined> => await this.validateInput(context, value)
+        })).trim();
+    }
+
+    public shouldPrompt(): boolean {
+        return true;
```
nturinski, line 26: Typically, we check the value to see if it should prompt or not. Something like `return context.ruleName !== undefined`. The reason being, if we ever make a "basic" version of this command and set the `ruleName` to some default value without user input, the prompt step would be skipped.

src/commands/scaling/addScaleRule/GetScaleRuleNameStep.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    private async validateInput(context: IAddScaleRuleWizardContext, name: string | undefined): Promise<string | undefined> {
+        if (!this.containerApp || !this.scaleRules) { await this.initContainerAppClient(context) }
+        name = name ? name.trim() : '';
+
+        if (!/^[a-z0-9]([-a-z0-9]*[a-z0-9])?$/.test(name)) {
+            return localize('invalidChar', `A name must consist of lower case alphanumeric characters or '-', and must start and end with an alphanumeric character.`);
+        }
+
+        const scaleRuleExists = !!(this.scaleRules?.filter((rule) => {
```
nturinski, line 46: You can do a this.scaleRules.find instead of filtering and checking the length.

src/commands/scaling/addScaleRule/GetScaleRuleTypeStep.ts
```diff
@@ -0,0 +1,46 @@
...
+export class GetScaleRuleTypeStep extends AzureWizardPromptStep<IAddScaleRuleWizardContext> {
+    public hideStepCount: boolean = true;
+
+    public async prompt(context: IAddScaleRuleWizardContext): Promise<void> {
+        const qpItems: QuickPickItem[] = [];
+        for (const ruleType in ScaleRuleTypes) { qpItems.push({ label: ScaleRuleTypes[ruleType as keyof typeof ScaleRuleTypes] }); }
+        context.ruleType = (await context.ui.showQuickPick(qpItems, {})).label;
+    }
+
+    public shouldPrompt(): boolean {
```
nturinski, line 25: Refer to comment above ^

src/commands/scaling/addScaleRule/GetScaleRuleTypeStep.ts
```diff
@@ -0,0 +1,46 @@
...
+import { QuickPickItem } from 'vscode';
+import { ScaleRuleTypes } from '../../../constants';
+import { GetHttpConcurrentRequestsStep } from './http/GetHttpConcurrentRequestsStep';
+import { IAddScaleRuleWizardContext } from './IAddScaleRuleWizardContext';
+import { GetQueueAuthSecretStep } from './queue/GetQueueAuthStep';
+import { GetQueueAuthTriggerStep } from './queue/GetQueueAuthTriggerStep';
+import { GetQueueLengthStep } from './queue/GetQueueLengthStep';
+import { GetQueueNameStep } from './queue/GetQueueNameStep';
+
+export class GetScaleRuleTypeStep extends AzureWizardPromptStep<IAddScaleRuleWizardContext> {
```
nturinski, line 16: I think we can just call it `ScaleRuleTypeStep`. I think a promptStep is kind of assumed to be a "get" of some sort.

src/commands/scaling/addScaleRule/addScaleRule.ts
```diff
@@ -0,0 +1,32 @@
...
+import { GetScaleRuleNameStep } from "./GetScaleRuleNameStep";
+import { GetScaleRuleTypeStep } from "./GetScaleRuleTypeStep";
+import { IAddScaleRuleWizardContext } from "./IAddScaleRuleWizardContext";
+
+export async function addScaleRule(context: IActionContext, node?: ScaleRuleGroupTreeItem): Promise<void> {
+    if (!node) {
+        node = await ext.tree.showTreeItemPicker<ScaleRuleGroupTreeItem>(new RegExp(ScaleRuleGroupTreeItem.contextValue), context);
+    }
+
+    const title: string = localize('addScaleRuleTitle', 'Create Scale Rule');
```
nturinski, line 20: Kind of weird that it's called "Add Scale Rule" everywhere, but the title is "Create Scale Rule".

src/commands/scaling/addScaleRule/AddNewScaleRule.ts
nturinski, line 18: Some of this logic could be moved to `ScaleRuleGroupTreeItem.createChildImpl`. If you look at `ManagedEnvironmentTreeItem`, you'll see what I mean, but the benefit of doing this is that it'll automatically handle adding the child to the parent in the UI. You will also see the "Creating..." node under the parent as it's creating.

If you need more context/have questions, let's talk about it offline.

src/commands/scaling/addScaleRule/queue/GetQueueLengthStep.ts
```diff
@@ -0,0 +1,34 @@
...
+    public shouldPrompt(): boolean {
+        return true;
+    }
+
+    private async validateInput(length: string | undefined): Promise<string | undefined> {
+        length = length ? length.trim() : '';
+        if (!/^[1-9]+[0-9]*$/.test(length)) {
+            return localize('invalidQueueLength', 'The number of requests must be a whole number greater than or equal to 1.');
+        }
+        if (Number(length) > thirtyTwoBitMaxSafeInteger) {
```
nturinski, line 28: You can probably leave `thirtyTwoBitMaxSafeInteger` in this file since it's only being used in this one spot.

src/commands/scaling/addScaleRule/IAddScaleRuleWizardContext.ts
```diff
@@ -0,0 +1,18 @@
...
+import { IActionContext } from "@microsoft/vscode-azext-utils";
+import { ScaleRuleGroupTreeItem } from "../../../tree/ScaleRuleGroupTreeItem";
+
+export interface IAddScaleRuleWizardContext extends IActionContext {
+    treeItem: ScaleRuleGroupTreeItem;
+    ruleName?: string;
+    ruleType?: string;
+    concurrentRequests?: string;
+    queueName?: string;
+    queueLength?: string;
```
nturinski, line 15: It's a matter of preference, but I usually define the context values to match the "true" value is going to be at the time of the API call. What I mean is, queueLength will come in as a string because of user input, but when being sent in the request, it should be a number.

Therefore, I would do the conversion in the prompt step rather than at the time of the API call.  The benefit being, if we ever use that step again for something else, we wouldn't have to remember to convert it to a number.

src/commands/scaling/addScaleRule/addScaleRule.ts
MicroFish91: Yeah, I actually consciously made that choice just because it felt like a more accurate description of those specific steps, but I totally see why the lack of consistency would feel weird so I'll change it

src/commands/scaling/addScaleRule/GetScaleRuleNameStep.ts
bwateratmsft: I'd do `return !context.ruleName`, so any falsy value will prompt.

src/commands/scaling/addScaleRule/GetScaleRuleNameStep.ts
bwateratmsft: `.some()` is meant for this. `this.scaleRules?.some((rule) => rule.name === name)`

src/commands/scaling/addScaleRule/GetScaleRuleNameStep.ts
nturinski: Ah sorry, yeah, `context.ruleName === undefined` is what I meant. I tend to be specific about this incase a "" is a valid value or if the step deals with boolean values, but generally `!context.ruleName` works as well.

MicroFish91: It's all good, I already understood what you meant :)

src/commands/scaling/addScaleRule/addScaleRule.ts
```diff
@@ -0,0 +1,23 @@
...
+
+
+export async function addScaleRule(context: IActionContext, node?: ScaleRuleGroupTreeItem): Promise<void> {
+    if (!node) {
+        node = await ext.tree.showTreeItemPicker<ScaleRuleGroupTreeItem>(new RegExp(ScaleRuleGroupTreeItem.contextValue), context);
+    }
+
+    const srNode: ScaleRuleTreeItem = await node.createChild(context);
+    // void showContainerAppCreated(srNode);
+    console.log(srNode);
```
alexweininger, line 20: Should these be removed?

src/commands/scaling/addScaleRule/GetScaleRuleNameStep.ts
bwateratmsft: Yeah, I try to be deliberate in my thinking about strings w.r.t. `undefined` vs `''` vs `null` etc. _Most_ of the time (but certainly not _all_ of the time), any falsy value is semantically the same. Fortunately `null` rarely ever shows up (though argh the `undefined` vs `null` thing is a nightmare). It's up to you to decide if `''` is an acceptable value but for a resource name I would think it is not.

I also generally avoid doing equality checks against `undefined` _because_ of that `null` vs. `undefined` mess, unless I know with certainty that `undefined` is the thing I'm looking for.

src/commands/scaling/addScaleRule/addScaleRule.ts
MicroFish91: Yup, I that was from an unfinished WIP-commit :)

src/commands/scaling/addScaleRule/AddScaleRuleStep.ts
```diff
@@ -0,0 +1,84 @@
...
+                    queueLength: context.queueLength,
+                    auth: [{ secretRef: context.secretRef, triggerParameter: context.triggerParameter }]
+                }
+                break;
+            default:
+        }
+        return scaleRule;
+    }
+
+    private integrateRule(context: IAddScaleRuleWizardContext, scaleRules: ScaleRule[], scaleRule: ScaleRule): void {
```
alexweininger, line 69: Based on what this method does for HTTP scale rules, it looks like Container Apps can only have one HTTP scale rule.

Is that true? I couldn't find it quickly in the docs. If so we should add a comment here explaining why/what this code is doing.

MicroFish91: Yup, that's currently the behavior that the portal has implemented, so I was keeping it consistent with that.  I'll go and add a comment to explain that

src/commands/scaling/addScaleRule/queue/QueueLengthStep.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    public shouldPrompt(context: IAddScaleRuleWizardContext): boolean {
+        return context.queueLength === undefined;
+    }
+
+    private async validateInput(length: string | undefined): Promise<string | undefined> {
+        length = length ? length.trim() : '';
+
+        const thirtyTwoBitMaxSafeInteger = 2147483647;
+        if (!/^[1-9]+[0-9]*$/.test(length)) {
```
nturinski, line 26: Could probably make this regex a constant. 

Or if `HttpConcurrentRequestStep` actually follows `thirtyTwoBitMaxSafeInteger` as well, maybe you could make a base step that has this  implementation of `validateInput`

src/tree/ScaleRuleGroupTreeItem.ts
```diff
@@ -30,14 +36,39 @@ export class ScaleRuleGroupTreeItem extends AzExtParentTreeItem implements IAzur
...
+            promptSteps: [new ScaleRuleNameStep(), new ScaleRuleTypeStep()],
+            executeSteps: [new AddScaleRuleStep()],
+            showLoadingPrompt: true
+        };
+        const wizard: AzureWizard<IAddScaleRuleWizardContext> = new AzureWizard(wizardContext, wizardOptions);
+        await wizard.prompt();
+        context.showCreatingTreeItem(nonNullProp(wizardContext, 'ruleName'));
+        await wizard.execute();
+
+        if (wizardContext.error !== undefined) {
```
nturinski, line 58: You should be able to just throw the error from the wizard step instead of tracking it and then throwing it after the execution.

src/commands/scaling/addScaleRule/GetScaleRuleNameStep.ts
MicroFish91: In our case here, would it make sense to do an input validation check instead then?  Maybe something like this

`
public shouldPrompt(context): boolean {
        return this.validateInput(context.ruleName);
}
`

MicroFish91: Would it make sense to leverage our `validateInput` function to help us here then?  Maybe something like...

`
function shouldPrompt(context): boolean {
  return this.validateInput(context.ruleName) !== undefined;
}
`

nturinski: That's not really how `shouldPrompt` is used. It gets called when the wizard is created to get the total number of steps. If anything returns false at that time, it will not be included in total steps. It is then called again _before_ prompt is run, to see if it should skip it or not. 

If you tried to use validate input in there, there'd be no context.ruleName at the time of creation.

src/commands/scaling/addScaleRule/AddScaleRuleStep.ts
```diff
@@ -0,0 +1,85 @@
...
+        const scaleRule: ScaleRule = this.buildRule(context);
+        this.integrateRule(context, template.scale.rules, scaleRule);
+
+        try {
+            ext.outputChannel.appendLog(adding);
+            await updateContainerApp(context, context.containerApp, { template });
+            void window.showInformationMessage(added);
+            ext.outputChannel.appendLog(added);
+            context.scaleRule = scaleRule;
+        } catch (error) {
```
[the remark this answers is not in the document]
MicroFish91: Oh I see, I should have checked how createChildImpl was called, it looks like it just swallows the error without display it which covers the issue I was trying to avoid by doing this (having two errors pop up because of it not being percolated up properly)

src/commands/scaling/addScaleRule/ScaleRuleNameStep.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import { ContainerApp, ScaleRule } from '@azure/arm-appcontainers';
+import { AzureWizardPromptStep } from '@microsoft/vscode-azext-utils';
+import { localize } from '../../../utils/localize';
+import { IAddScaleRuleWizardContext } from './IAddScaleRuleWizardContext';
+
+export class ScaleRuleNameStep extends AzureWizardPromptStep<IAddScaleRuleWizardContext> {
+    public hideStepCount: boolean = true;
+    containerApp: ContainerApp | undefined;
+    scaleRules: ScaleRule[] | undefined;
```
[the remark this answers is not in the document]
MicroFish91: I hid the step count so well you couldn't see it ;)

nturinski (commented):
Shoot, I think my suggestion may have introduced a bug in the back button. I can explain offline, but it's because you're using the httpProperties/queueProperties in the context now 😬

src/commands/scaling/addScaleRule/queue/QueueAuthSecretStep.ts
```diff
@@ -6,16 +6,24 @@
...
 import { QuickPickItem } from 'vscode';
+import { localize } from '../../../../utils/localize';
 import { IAddScaleRuleWizardContext } from '../IAddScaleRuleWizardContext';
 
 export class QueueAuthSecretStep extends AzureWizardPromptStep<IAddScaleRuleWizardContext> {
     public async prompt(context: IAddScaleRuleWizardContext): Promise<void> {
+        const noSecrets: string = localize('noSecretsFound', 'No secrets were found. Create a secret to proceed.');
+        const placeHolder: string = localize('chooseSecretRef', 'Choose a secret reference');
         const containerAppWithSecrets = await context.containerApp.getContainerEnvelopeWithSecrets(context);
         const secrets: Secret[] | undefined = containerAppWithSecrets.configuration.secrets;
```
nturinski, line 17: Couldn't you just check to see if `secrets.length` were 0 here and throw an error if that's the case?

src/commands/scaling/addScaleRule/ScaleRuleNameStep.ts
```diff
@@ -0,0 +1,44 @@
...
+
+        if (!/^[a-z0-9]([-a-z0-9]*[a-z0-9])?$/.test(name)) {
+            return localize('invalidChar', `A name must consist of lower case alphanumeric characters or '-', and must start and end with an alphanumeric character.`);
+        }
+
+        const scaleRuleExists: boolean = !!this.scaleRules?.some((rule) => {
+            return rule?.name?.length && rule?.name === name;
+        });
+        if (scaleRuleExists) {
+            return localize('scaleRuleExists', 'The scale rule "{0}" already exists in container app "{1}". Please enter a unique name.', name, this.containerApp?.name as string);
```
nturinski, line 40: nit: I don't think you need the `as string` typing at the end since localize doesn't care if it's undefined or not.

src/commands/scaling/addScaleRule/queue/QueueAuthSecretStep.ts
```diff
@@ -0,0 +1,29 @@
...
+
+import { Secret } from '@azure/arm-appcontainers';
+import { AzureWizardPromptStep, nonNullProp } from '@microsoft/vscode-azext-utils';
+import { QuickPickItem } from 'vscode';
+import { localize } from '../../../../utils/localize';
+import { IAddScaleRuleWizardContext } from '../IAddScaleRuleWizardContext';
+
+export class QueueAuthSecretStep extends AzureWizardPromptStep<IAddScaleRuleWizardContext> {
+    public async prompt(context: IAddScaleRuleWizardContext): Promise<void> {
+        const noSecrets: string = localize('noSecretsFound', 'No secrets were found. Create a secret to proceed.');
```
nturinski, line 14: nit: I personally prefer to initializing variables closer to where they are being used.

src/commands/scaling/addScaleRule/AddScaleRuleStep.ts
```diff
@@ -0,0 +1,78 @@
...
+        const template: Template = context.containerApp.data.template || {};
+        template.scale = context.scale.data || {};
+        template.scale.rules = context.scaleRuleGroup.data || [];
+
+        const scaleRule: ScaleRule = this.buildRule(context);
+        this.integrateRule(context, template.scale.rules, scaleRule);
+
+        ext.outputChannel.appendLog(adding);
+        await updateContainerApp(context, context.containerApp, { template });
+        context.scaleRule = scaleRule;
```
nturinski, line 31: nit: I don't think that there's really any point in doing this. You're not really using it anywhere else, as far as I can tell.

src/utils/treeUtils.ts
```diff
@@ -15,4 +15,18 @@ export namespace treeUtils {
...
+        let foundParent: boolean = false;
+        let currentNode: AzExtTreeItem = node;
+        while (currentNode.parent) {
+            if (currentNode.constructor.name === parentInstance) {
+                foundParent = true;
+                break;
+            }
+            currentNode = currentNode.parent;
+        }
+        return foundParent ? currentNode as T : null;
```
nturinski, line 30: I think you should throw some sort of error here if you don't get a parent. If it returns as `null`, you'll get some funky errors in other parts of code. That way you won't have to do any `as T` typings either.

nturinski (commented):
In general, anything I write with "nit:" means stylistically, it's something I felt like I could comment on, but it's not something that I would actually block you merging your PR in for (especially if you disagree)

src/commands/scaling/addScaleRule/AddScaleRuleStep.ts
MicroFish91: Yeah good catch, this had a specific use in a previous iteration, but now has gone the way of the appendix 😂

src/utils/treeUtils.ts
MicroFish91: Good idea, this looks way cleaner now

MicroFish91:
Okay, I think I fixed all relevant change suggestions, please review again when ready :)

src/commands/scaling/addScaleRule/AddScaleRuleStep.ts
nturinski: Whoops

src/commands/scaling/addScaleRule/addScaleRule.ts
```diff
@@ -0,0 +1,19 @@
...
+import { ScaleRuleGroupTreeItem } from "../../../tree/ScaleRuleGroupTreeItem";
+import { treeUtils } from "../../../utils/treeUtils";
+
+export async function addScaleRule(context: IActionContext, node?: ScaleRuleGroupTreeItem): Promise<void> {
+    if (!node) {
+        node = await ext.tree.showTreeItemPicker<ScaleRuleGroupTreeItem>(new RegExp(ScaleRuleGroupTreeItem.contextValue), context);
+    }
+    const containerApp: ContainerAppTreeItem = treeUtils.findNearestParent(node, ContainerAppTreeItem.prototype);
+    await node.createChild(context);
+    await containerApp?.refresh(context);
```
nturinski, line 18: nit: Probably don't need the ? anymore

nturinski (approved):
🎉🎉🎉~~Congrats on your first big feature PR! 53 total comments isn't too bad ~~🎉🎉🎉

DanceDancingGIF

Issue #150, opened 2022-07-06
[Suggestion] It would be better to remove the extra "Container Apps" node


v-xinda:
**OS:** All
**Build Version:** 20220706.3

**Repro Steps:**
1. Make sure the resource groups by Resource Type.
2. Install the Container app extension -> Expand the subscription.
3. Check whether there is no extra "Container Apps" node. 

**Actual:**
There is an extra "Container Apps" node. 
![image](https://user-images.githubusercontent.com/57662131/177529011-8bc35850-ecb2-47dd-a6ac-5b8d2ad163b3.png)

**Suggestion:**
It would be better to remove the extra "Container Apps" node. 

**More Info:**
The "Container Apps Environment" node contains the container apps.

nturinski:
I plan on adding a setting that will take care of this problem.

By default, the Environment node will be displayed but the user can also display the Container Apps node as well.

Issue #148, opened 2022-07-06
There is no response when clicking "Open Log Stream/Open Logs" in Logs node


v-ruizh:
**OS:** All
**Build Version:** 20220706.3

**Repro Steps:**
1. Expand a container app -> Logs node.
2. Click "Open Log Stream/Open Logs".
3. Check whether the corresponding page is opened or not.

**Expect:**
The corresponding page is opened.

**Actual:**
There is no response.

Issue #146, opened 2022-07-06
There is no "Deleting..." label next to the container app when deleting a container app from command palette


v-xinda:
**OS:** All
**Build Version:** 20220706.3
**Regression:** Previous Release

**Repro Steps:**
1. F1 -> Click "Azure Container Apps: Delete Container App..." -> Select a subscription -> Select a container app.
2. Check whether there is a "Deleting..." label next to the container app.

**Expect:**
There is a "Deleting..." label next to the container app.
![image](https://user-images.githubusercontent.com/57662131/177519880-cd95c79f-1d1d-4ced-a39c-045b37eb53c3.png)

**Actual:**
There is no "Deleting..." label next to the container app.
![image](https://user-images.githubusercontent.com/57662131/177520048-e1fca0d2-f2ab-4594-bee3-266a1786b944.png)

**More Info:**
This issue doesn't reproduce when deleting a container app from context menu.

Issue #145, opened 2022-07-06
The deleted container app still displays under the container apps environment after deleing it from command palette


v-xinda:
**OS:** All
**Build Version:** 20220706.3
**Regression:** Previous Release

**Repro Steps:**
1. F1 -> Click "Azure Container Apps: Delete Container App..." -> Select a subscription -> Select a container app.
2. Check whether the deleted container app disappears from the container apps environment.

**Expect:**
The deleted container app disappears from the container apps environment.

**Actual:**
The deleted container app still displays under the container apps environment.
![image](https://user-images.githubusercontent.com/57662131/177521881-d2c97f5f-e598-41f6-8dad-9fde9877a0de.png)

**More Info:**
1. This issue doesn't reproduce when deleting a container app from context menu.
2. Refresh can fix this issue.

v-xinda:
1. The deleted container apps environment still shows in the list when deleting the container apps environment from command palette.
![image](https://user-images.githubusercontent.com/57662131/177531002-0b3d46a6-3af0-43c2-80b4-5b8c72e1543c.png)
2. The deleted container apps environment can be deleted again successfully.
![image](https://user-images.githubusercontent.com/57662131/177531306-87aef5ab-2232-493b-9aad-3b59d075d1fb.png)

PR #170, opened 2022-08-02
Unified Integration

MicroFish91:
Related to:  https://github.com/microsoft/vscode-azureresourcegroups/pull/353

Fixes: #150 
Fixes: #148 
Fixes: #146 
Fixes: #145
Fixes: #140


src/tree/ContainerAppTreeItem.ts
```diff
@@ -31,6 +31,7 @@ export class ContainerAppTreeItem extends AzExtParentTreeItem implements IAzureR
 
     public name: string;
     public label: string;
+    public childTypeLabel: string = 'Container App setting';
```
alexweininger, line 34: localize

src/tree/LogsTreeItem.ts
```diff
@@ -10,10 +10,12 @@ import { ContainerAppTreeItem } from "./ContainerAppTreeItem";
 
 export class LogsTreeItem extends AzExtParentTreeItem {
     public static contextValue: string = 'log';
+    public static openLogsContext: string = 'openLog';
     public readonly contextValue: string = LogsTreeItem.contextValue;
     public readonly parent: ContainerAppTreeItem;
 
     public label: string;
+    public childTypeLabel: string = 'log view';
```
alexweininger, line 18: localize

package.json
```diff
@@ -169,142 +157,88 @@
...
-                },
                 {
                     "command": "containerApps.createManagedEnvironment",
-                    "when": "view == containerApps && viewItem =~ /azureSubscription/i",
+                    "when": "view == azureResourceGroups && viewItem =~ /azureResourceTypeGroup.*microsoft.app/managedenvironments/",
                     "group": "2@1"
                 },
                 {
                     "command": "containerApps.createContainerApp",
-                    "when": "view == containerApps && viewItem =~ /managedEnvironment/i",
```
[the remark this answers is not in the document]
MicroFish91: Thanks for bringing this one up.  This was actually meant to be a temporary fix but I had forgotten to come back to it.  I did it because of this overlap:

image

image

After talking with Nathan about it, I think I will end up changing the `contextValue` from `managedEnvironment` to something like `containerEnvironment` instead

package.json
```diff
@@ -54,39 +49,32 @@
     ],
     "main": "./main.js",
     "contributes": {
+        "x-azResources": {
+            "activation": {
+                "onFetch": [
```
bwateratmsft, line 54: Does this really need to be `onFetch` or can it be `onResolve` instead?

package.json
```diff
@@ -169,142 +157,88 @@
...
-                },
-                {
-                    "command": "containerApps.refresh",
-                    "when": "view == containerApps && viewItem =~ /azureSubscription|managedEnvironment|containerApp/i",
-                    "group": "z@2"
-                },
                 {
                     "command": "containerApps.createManagedEnvironment",
-                    "when": "view == containerApps && viewItem =~ /azureSubscription/i",
+                    "when": "view == azureResourceGroups && viewItem =~ /azureResourceTypeGroup.*microsoft.app/managedenvironments/",
```
bwateratmsft, line 164: I think this is a safer bet:
```suggestion
                    "when": "view == azureResourceGroups && viewItem =~ /azureResourceTypeGroup/i && viewItem =~ /microsoft.app/managedenvironments/i",
```
(And maybe not `i` depending on the below comment thread)

src/commands/openInPortal.ts
```diff
@@ -4,12 +4,16 @@
  *--------------------------------------------------------------------------------------------*/
 import * as azUtil from '@microsoft/vscode-azext-azureutils';
```
bwateratmsft, line 5: nit: rename file to `openLogsInPortal.ts` (and update references accordingly)

src/ContainerAppsResolver.ts
```diff
@@ -0,0 +1,32 @@
...
+import { createContainerAppsAPIClient } from "./utils/azureClients";
+import { getResourceGroupFromId } from "./utils/azureUtils";
+
+export class ContainerAppsResolver implements AppResourceResolver {
+    public async resolveResource(subContext: ISubscriptionContext, resource: AppResource): Promise<ResolvedAppResourceBase | null> {
+        return await callWithTelemetryAndErrorHandling('resolveResource', async (context: IActionContext) => {
+            try {
+                const client: ContainerAppsAPIClient = await createContainerAppsAPIClient([context, subContext]);
+                const rgName: string = getResourceGroupFromId(resource.id);
+                const me: ManagedEnvironment = await client.managedEnvironments.get(rgName, resource.name);
```
bwateratmsft, line 20: It doesn't have to be part of this PR, but if at all possible we should defer this call until it's _really_ needed, to make resolve very fast.

src/tree/ManagedEnvironmentTreeItem.ts
```diff
@@ -3,166 +3,60 @@
...
-import { treeUtils } from "../utils/treeUtils";
-import { ContainerAppTreeItem } from "./ContainerAppTreeItem";
 import { IAzureResourceTreeItem } from './IAzureResourceTreeItem';
+import { ResolvedContainerAppsResource } from "./ResolvedContainerAppsResource";
 
 export class ManagedEnvironmentTreeItem extends AzExtParentTreeItem implements IAzureResourceTreeItem {
-    public static contextValue: string = 'managedEnvironment';
-    public static contextValueRegExp: RegExp = new RegExp(ManagedEnvironmentTreeItem.contextValue);
-    public readonly contextValue: string = `${ManagedEnvironmentTreeItem.contextValue}|${azResourceContextValue}`;
+    public static contextValue: string;
```
bwateratmsft, line 13: I think we should leave the initialization of these static ones here instead of the constructor, i.e.
```suggestion
    public static contextValue: string = ResolvedContainerAppsResource.contextValue;
```

src/tree/ResolvedContainerAppsResource.ts
```diff
@@ -0,0 +1,179 @@
...
+
+        this.resourceGroupName = getResourceGroupFromId(this.id);
+        this.name = nonNullProp(this.data, 'name');
+        this.label = this.name;
+
+        this.contextValuesToAdd.push(this.resolvedContextValue);
+    }
+
+    public get iconPath(): TreeItemIconPath {
+        return treeUtils.getIconPath('Container App Environments placeholder icon icon');
```
bwateratmsft, line 54: A `// TODO`, I'm guessing?

src/commands/createManagedEnvironment/createManagedEnvironment.ts
```diff
@@ -3,17 +3,16 @@
...
 import { IManagedEnvironmentContext } from "./IManagedEnvironmentContext";
 
-export async function createManagedEnvironment(context: IActionContext & Partial<ICreateChildImplContext> & Partial<IManagedEnvironmentContext>, node?: SubscriptionTreeItem): Promise<ManagedEnvironmentTreeItem> {
-    if (!node) {
-        node = await ext.tree.showTreeItemPicker<SubscriptionTreeItem>(SubscriptionTreeItem.contextValue, context);
-    }
-
-    const keNode: ManagedEnvironmentTreeItem = await node.createChild(context);
+export async function createManagedEnvironment(context: IActionContext, node?: SubscriptionTreeItemBase): Promise<ManagedEnvironmentTreeItem> {
+    node = await ext.rgApi.appResourceTree.showTreeItemPicker<SubscriptionTreeItemBase>(SubscriptionTreeItemBase.contextValue, context);
```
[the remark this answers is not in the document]
MicroFish91: So I wasn't sure how best to handle this, but basically my thought process was something like this.  If we use the if (!node) method, then the flow would be something like:

1. If you right click on the node to initiate a create, node will enter command as `GroupTreeItemBase`
2. If you initiate a create via QP, node will enter command as `undefined` and then turn into `SubscriptionTreeItemBase`

The command will work in both cases because the information it needs is on the treeItem; however, I thought it made more sense to have it so that the node was always a type that we could consistently expect - hence no if statement and just always re-initialize node to `SubscriptionTreeItemBase` in all cases.

Let me know if that makes sense or if I'm thinking about this in a weird way.  I'm thinking maybe it was a bad idea to label the incoming node type as `SubscriptionTreeItemBase` - maybe I can just say something like `_node?: AzExtTreeItem` and just initialize the subscription node reference as a separate variable to avoid any potential confusion down the road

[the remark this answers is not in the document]
alexweininger: The desired behavior is to avoid prompting for the subscription if at all possible. If the command has been run from the Container Apps type group tree item context menu, we shouldn't need to prompt for the subscription.

Although it's a `GroupTreeItemBase`, we can just cast it as a `SubscriptionTreeItem` since it has the `.subscription` property. It's not ideal, but we do that in every create command IIRC. [Checkout the Create Function App command as an example](https://github.com/microsoft/vscode-azurefunctions/blob/alex/mainjs/src/commands/createFunctionApp/createFunctionApp.ts#L13)

MicroFish91:
Oh yeah, I think I made all the changes recommended, let me know if you guys see anything else that might need changing :)

src/ContainerAppsResolver.ts
```diff
@@ -0,0 +1,27 @@
...
+import { managedEnvironmentProvider } from "./constants";
+import { ResolvedContainerAppsResource } from "./tree/ResolvedContainerAppsResource";
+import { createContainerAppsAPIClient } from "./utils/azureClients";
+import { getResourceGroupFromId } from "./utils/azureUtils";
+
+export class ContainerAppsResolver implements AppResourceResolver {
+    public async resolveResource(subContext: ISubscriptionContext, resource: AppResource): Promise<ResolvedAppResourceBase | null> {
+        return await callWithTelemetryAndErrorHandling('resolveResource', async (context: IActionContext) => {
+            const client: ContainerAppsAPIClient = await createContainerAppsAPIClient([context, subContext]);
+            const rgName: string = getResourceGroupFromId(resource.id);
```
nturinski, line 18: There should be a util function for this now in `vscode-azext-utils` that you can import (and remove utils/azureUtils) from here.

src/getExtensionApi.ts
```diff
@@ -0,0 +1,19 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { Extension, extensions } from "vscode";
+
+export async function getApiExport<T>(extensionId: string): Promise<T | undefined> {
```
nturinski, line 8: I think you should go ahead and implement the `getResourceGroupApi` funciton in here since you already have this file.

Just makes the `extension.ts` file cleaner.

https://github.com/microsoft/vscode-azurefunctions/blob/main/src/getExtensionApi.ts#L24-L31

src/tree/ContainerAppTreeItem.ts
```diff
@@ -31,6 +31,7 @@ export class ContainerAppTreeItem extends AzExtParentTreeItem implements IAzureR
 
     public name: string;
     public label: string;
+    public childTypeLabel: string = localize('containerAppSetting', 'Container App setting');
```
nturinski, line 34: What was the reason for this change?? I assume that it's for a quickpick thing, but not all of the children will be `Container App setting`

src/tree/ResolvedContainerAppsResource.ts
```diff
@@ -0,0 +1,176 @@
...
+import { createContainerAppsAPIClient } from "../utils/azureClients";
+import { getResourceGroupFromId } from "../utils/azureUtils";
+import { localize } from "../utils/localize";
+import { nonNullProp } from "../utils/nonNull";
+import { settingUtils } from "../utils/settingUtils";
+import { ContainerAppTreeItem } from "./ContainerAppTreeItem";
+import { IAzureResource } from "./IAzureResourceTreeItem";
+import { ManagedEnvironmentTreeItem } from "./ManagedEnvironmentTreeItem";
+
+export class ResolvedContainerAppsResource implements ResolvedAppResourceBase, IAzureResource {
```
nturinski, line 29: It's a little confusing that the class is a `ResolvedContainerAppsResource` but is actually a `containerEnvironment`. I think calling it a `ResolvedContainerEnvironmentResource` might make more sense in this case even if the ext is Container Apps.

src/tree/ContainerAppTreeItem.ts
MicroFish91: Yup it's a quick pick related thing. I believe walking through the openLogsInPortal command via the command palette made me want to add this. I didn't want the QP prompt showing up as undefined for the label. Is there a better label you'd prefer to use?

src/ContainerAppsResolver.ts
MicroFish91: Okay, I'm planning to change the other instances of this function call as well to sync up with this

nturinski (approved):
Very nicely done! 🎉

Even though all of this is about to be 🦖, it was definitely necessary and hopefully a good learning opportunity!

PR #179, opened 2022-08-25
Add `view properties` command back to all children below `containerAppEnvironment` level

MicroFish91:


src/commands/viewProperties.ts
```diff
@@ -0,0 +1,30 @@
...
+            filter: rootFilter,
+            expectedChildContextValue: azureResourceRegExp
+        });
+    }
+
+    if (!node.data) {
+        if (node.getDataImpl) {
+            await node.getDataImpl();
+        } else {
+            throw new Error(localize('No data exists on resource "{0}"', node.label));
```
bwateratmsft, line 25: `localize` needs an ID for each string;

```suggestion
            throw new Error(localize('viewProperties.noData', 'No data exists on resource "{0}"', node.label));
```

As written, it would think the "No data exists..." is the ID and `node.label` is the string to show

bwateratmsft (approved):

PR #181, opened 2022-09-01
Add Container App Delete Support to the Activity Log

MicroFish91:


src/commands/deleteContainerApp/DeleteContainerAppStep.ts
```diff
@@ -0,0 +1,40 @@
...
+            await webClient.containerApps.beginDeleteAndWait(context.containerApp.resourceGroupName, context.containerApp.name);
+        } catch (error) {
+            const pError = parseError(error);
+            // a 204 indicates a success, but sdk is catching it as an exception
+            // accept any 2xx reponse code
+            if (Number(pError.errorType) < 200 || Number(pError.errorType) >= 300) {
+                throw error;
+            }
+        }
+        if (!context.suppressNotification) { void window.showInformationMessage(deleteSucceeded); }
```
alexweininger, line 33: nit: I prefer if the if-statement body is on a newline.

src/tree/ContainerAppTreeItem.ts
```diff
@@ -91,32 +92,22 @@ export class ContainerAppTreeItem extends AzExtParentTreeItem implements IAzureR
...
+            activityTitle: deleteContainerApp,
+            containerApp: this,
+            ...context, ...(await createActivityContext())
+        };
+        const wizard: AzureWizard<IDeleteContainerAppWizardContext> = new AzureWizard(wizardContext, {
+            promptSteps: [new DeleteConfirmationStep(confirmMessage)],
+            executeSteps: [new DeleteContainerAppStep()]
         });
+
+        if (!context.suppressNotification) { await wizard.prompt(); }
```
alexweininger, line 109: nit: I prefer if the if-statement body is on a newline.

src/tree/ContainerAppTreeItem.ts
```diff
@@ -91,32 +92,22 @@ export class ContainerAppTreeItem extends AzExtParentTreeItem implements IAzureR
...
-            }
-
-            void window.showInformationMessage(deleteSucceeded);
-            ext.outputChannel.appendLog(deleteSucceeded);
+        const deleteContainerApp: string = localize('deleteContainerApp', 'Delete Container App "{0}"', this.name);
+
+        const wizardContext: IDeleteContainerAppWizardContext = {
+            activityTitle: deleteContainerApp,
+            containerApp: this,
+            ...context, ...(await createActivityContext())
```
alexweininger, line 102: nit: each property on a new line

src/commands/deleteContainerApp/IDeleteContainerAppWizardContext.ts
```diff
@@ -0,0 +1,11 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { ExecuteActivityContext, IActionContext } from "@microsoft/vscode-azext-utils";
+import { ContainerAppTreeItem } from "../../tree/ContainerAppTreeItem";
+
+export interface IDeleteContainerAppWizardContext extends IActionContext, ExecuteActivityContext {
+    containerApp: ContainerAppTreeItem;
```
alexweininger, line 10: Adding only the needed properties to the context rather than the entire tree item would be best. In this case I think it's `resourceGroupName`, `name`, and `subscription`. It's not a huge deal, but I think it's best not to put tree items in the context.

src/tree/ContainerAppTreeItem.ts
```diff
@@ -91,32 +92,22 @@ export class ContainerAppTreeItem extends AzExtParentTreeItem implements IAzureR
         await openUrl(`https://${this.data.configuration?.ingress?.fqdn}`);
     }
 
-    public async deleteTreeItemImpl(context: IActionContext & { suppressPrompt?: boolean }): Promise<void> {
+    public async deleteTreeItemImpl(context: IActionContext & { suppressNotification?: boolean }): Promise<void> {
```
alexweininger, line 95: Was this intentional?

MicroFish91: I made the assumption that `suppressPrompt` also meant `suppressNotification`.  I was referencing some other delete implementations and it looked like we also typically suppress showing output windows in the execute step.  Since we're checking `context.suppressNotification` in the execute step, I figured we needed to pass it that value when we initialized the `wizardContext` so that it aligns with this configuration (that way we can pull it by just doing `...context`):

Screen Shot 2022-09-01 at 11 51 02 AM

Thoughts?

alexweininger: `suppressNotification` is an activity log specific idea, which correlates to the user setting. This setting is on by default and makes it so that notifications don't appear if the same info could be gathered from the activity.

image

`suppressPrompt` is unrelated to `suppressMessage` and is just a flag that when `true`, means we shouldn't confirm prompt the user before completing the action. In this case that means we shouldn't prompt `DeleteConfirmationStep`.

MicroFish91: Okay, that makes sense, thanks for that clarification!

alexweininger: We could use some naming improvements for sure.

alexweininger (approved):

PR #182, opened 2022-09-02
Add Delete Managed Environment Support to the Activity Log

MicroFish91:
Note:  I refactored away from calling `containerApp.deleteTreeItem` directly in this implementation... reason being, calling this command in the execute steps of the `ManagedEnvironment` wizard causes multiple wizards to be run simultaneously because `containerApp.deleteTreeItem` also uses its own wizard.   The solution I initially put together was a bit hacky because it required the sharing of contexts across `deleteTreeItemImpl's`, so instead I am calling the Container Apps delete step directly in the `ManagedEnvironment` wizard and not the `deleteTreeItem` method.  

The main drawback to doing it this way is that each Container App child delete action doesn't show up as a separate node and we won't see delete spinners in the tree view... however, the user can still track the progress of each step in the delete cycle through `progress reports`.  Talking this over with @alexweininger, perhaps this is a reasonable alternative until we update the Activity Log API in the future such that it can elegantly handle shared `parent/child` activities.


src/commands/deleteContainerApp/DeleteAllContainerAppsStep.ts
```diff
@@ -0,0 +1,47 @@
...
+            } catch (error) {
+                const pError = parseError(error);
+                // a 204 indicates a success, but sdk is catching it as an exception
+                // accept any 2xx reponse code
+                if (Number(pError.errorType) < 200 || Number(pError.errorType) >= 300) {
+                    throw error;
+                }
+            }
+            if (!context.suppressNotification) {
+                void window.showInformationMessage(deleteSucceeded);
```
bwateratmsft, line 38: Do we want to show just one toast at the end instead of one per container app? Probably a good question for @fiveisprime.

bwateratmsft (approved):

PR #191, opened 2022-09-12
Add increased visibility to the creation of Managed Environments

### LINKED ISSUE #185: [Suggestion] It would be better to add a "View Output" button in the notification after creating a container apps environment
**OS:** All
**Build Version:** 20220908.8

**Repro Steps:**
1. Create a container apps environment.
2. Check whether there is a "View Output" button in the notification.

**Actual:**
There is no "View Output" button in the notification.
image

**Suggestion:**
It would be better to add a "View Output" button in the notification.

**More Info:**
This issue also reproduces when creating a container app.

MicroFish91:
Closes #185 

Added a `View Output` button to the success popup as well as an `activityResult` to the `Activity Log`.

Also did a bit of housekeeping in the `constants` file and the `createChild` method since I was already making changes there for this update.

### DIFF
--- a/src/commands/createManagedEnvironment/ManagedEnvironmentCreateStep.ts
+++ b/src/commands/createManagedEnvironment/ManagedEnvironmentCreateStep.ts
@@ -44,8 +44,9 @@ export class ManagedEnvironmentCreateStep extends AzureWizardExecuteStep<IManage
         );
 
         const createdKuEnv: string = localize('createKuEnv', 'Successfully created new Container Apps environment "{0}".', context.newManagedEnvironmentName);
+        const viewOutput: string = localize('viewOutput', 'View Output');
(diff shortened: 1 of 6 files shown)


bwateratmsft (approved):

src/commands/createManagedEnvironment/ManagedEnvironmentCreateStep.ts
```diff
@@ -43,9 +44,16 @@ export class ManagedEnvironmentCreateStep extends AzureWizardExecuteStep<IManage
             }
         );
 
+        context.activityResult = {
```
alexweininger, line 47: We can also do this now. We can change this after the release.

```suggestion
        context.activityResult = nonNullProp(context.managedEnvironment, 'id')
```

PR #207, opened 2022-09-20
Removed use of deprecated `Azure Log Analytics` package

MicroFish91:
The dependency has been deprecated in favor of Azure Monitor Query

Edit:  Will just remove the unused package.  Opened a related issue [here](https://github.com/microsoft/vscode-azurecontainerapps/issues/208) for when we add any query support in the future


bwateratmsft (approved):

nturinski:
We should honestly just remove the package altogether to keep the size down. 

Could you do that and create an issue that links this package and explains that we would need to use this to query the log analytics workspaces to display log information for the container app?

nturinski (approved):

PR #222, opened 2022-11-15
Fix name validation when creating a container app

### LINKED ISSUE #37: Update name validation
Check that environment and container app name validation matches the actual validation done by the service. For instance, VS Code has a 20 char max but the service doesn't appear to have the same issue.

### LINKED ISSUE #171: There is no validation if enter a name as "nicole-test--3" when creating a container app
**OS:** Win10
**Build Version:** 20220803.8

**Repro Steps:**
1. Right click a container apps environment -> Click "Create Container App..." -> Enter a name as "nicole-test--3".
2. Check whether there is a validation or not.

**Expect:**
There is a validation.
image

**Actual:**
There is no validation.
image

**More Info:**
Pop up an error as below after creating a container app with name "nicole-test--3".
image

motm32:
Fixes #171. 
Fixes #37. 

Validation done in the portal:
image

The error also does not seem to pop up when entering only one character.

### DIFF
--- a/src/commands/createContainerApp/ContainerAppNameStep.ts
+++ b/src/commands/createContainerApp/ContainerAppNameStep.ts
@@ -31,8 +31,8 @@ export class ContainerAppNameStep extends AzureWizardPromptStep<IContainerAppCon
         name = name ? name.trim() : '';
         // to prevent showing an error when the character types the first letter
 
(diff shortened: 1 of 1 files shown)


alexweininger (approved):
image

PR #229, opened 2022-12-12
`deployImageApi` implementation

MicroFish91:
Edit: Ready for review.


bwateratmsft:
> I wasn't sure how much string validation was necessary on the passed `imageName`. Right now it's pretty minimal - I just check to make sure there a matching number of args after we split up the `imageName` using `.split('/')`. I could definitely add some more robust validation, but wasn't sure how deep to go on it.

We can break up the image name in the Docker extension to pretty much any degree required, so if a slightly different format for the passed-in data is easier, feel free to change it.

src/utils/imageNameUtils.ts
```diff
@@ -0,0 +1,33 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import type { ContainerRegistryManagementClient, ContainerRegistryManagementModels } from "@azure/arm-containerregistry";
+import { ISubscriptionActionContext } from "@microsoft/vscode-azext-utils";
+import { acrDomain, acrDomainRegExp, dockerHubDomain, dockerHubDomainRegExp, SupportedRegistries } from "../constants";
+import { createContainerRegistryManagementClient } from "./azureClients";
+import { localize } from "./localize";
+
+export namespace imageNameUtils {
+    export function detectRegistryDomain(imageName: string): SupportedRegistries | undefined {
```
bwateratmsft, line 13: I'm a little leery of the regex approach here, since there's potential for mistakes. For example, if I supply an image with the name "azurecr.io.foo/bar:baz", it would be incorrectly detected as belonging to ACR. True, this example is contrived, but my gut is telling me something else could happen.

We _can_ supply the registry domain specifically, as part of the contract. The Docker extension already has some fairly sophisticated image name parsers that we trust. Given the domain, a more exact regex could be applied, such as `/\.azurecr\.io$/i` (which demands "azurecr.io" be the end of the domain, and that it start with a dot, e.g. "myfoo.azurecr.io" passes, but "myfoo.azurecr.iot" does not). For Docker Hub, I think `/^docker\.io$/i` is sufficient.

MicroFish91: Let me know if you think something like this would be ok!

image

Instead of amending the contract, just parsing the `registryName` off of the `imageName` and doing the regex search on that value.

bwateratmsft: Instead of `loginServer`, could we just send `registry` (the domain), and for Docker Hub's case just map `loginServer` to the other value it wants? IIRC, for ACR and others the login server === registry domain right?

MicroFish91:
Latest changes contingent on the changes from [this PR](https://github.com/microsoft/vscode-docker/pull/3818)

bwateratmsft (approved):

PR #230, opened 2022-12-19
Fix `openInPortal`

MicroFish91:
I noticed `openInPortal` wasn't showing up for the `containerApp` node.  I think it's likely that this is an artifact of when I initially did the unified migration a few months ago.  I think these changes should fix that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "preview": true,
     "activationEvents": [
         "onCommand:containerApps.viewProperties",
-        "onCommand:containerApps.openLogsInPortal",
+        "onCommand:containerApps.openInPortal",
         "onCommand:containerApps.reportIssue",
         "onCommand:containerApps.createContainerApp",
         "onCommand:containerApps.deployImage",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -29,7 +29,7 @@
     "preview": true,
     "activationEvents": [
         "onCommand:containerApps.viewProperties",
-        "onCommand:containerApps.openLogsInPortal",
+        "onCommand:containerApps.openInPortal",
```
nturinski, line 32: How come you changed the command name to `openInPortal`?

MicroFish91: Because the command should be generic enough to handle both our logs and container app nodes.  

It's kind of a long story - originally it was `openInPortal`, but when I did the unified merge, I didn't completely understand the difference between the `openInPortal` command issued by Resource Groups vs by the individual extension.  In migrating, I missed that we lost the `openInPortal` for our Container App nodes.  I thought at the time that we only needed to keep one command for our `GenericTreeItem` log nodes so I made it a separate command.  Now that I looked back, I realized that I hadn't handled it properly since the `ContainerApps` nodes do not show `openInPortal` so I'm fixing all of the open logic with one generic `openInPortal` (probably how it should have been handled the first time?)

nturinski (approved):

bwateratmsft:
When this extension is upgraded to V2, we should be able to remove this command and just use the `portalUrl` property on model objects.

PR #234, opened 2023-01-18
Only show `Disable Ingress for Container App` command when ingress is already enabled

### LINKED ISSUE #201: There is an "Enable Ingress" option when executing "Disable Ingress" command for a disabled Ingress container app
**OS:** All
**Build Version:** 20220914.3

**Repro Steps:**
1. Disable Ingress for one container app.
2. F1 -> Azure Container Apps: Disable Ingress for Container App -> Select the container app environment -> Select the container app which Ingress is disabled.
3. Check whether an error occurs or not.

**Expect:**
An error occurs.
image

**Actual:**
No error occurs and the next step is "Enable Ingress".
image

motm32:
Fixes #201. 

Instead of showing an error I thought it may be better to only show the `Disable Ingress for Container App` command in the command palette when Ingress is enabled. This is consistent with the behavior that happens when you right click on the Ingress icon.


package.json
```diff
@@ -299,6 +303,11 @@
                             "%containerApps.deleteConfirmation.ClickButton%"
                         ],
                         "default": "EnterName"
+                    },
+                    "containerApps.enableIngress": {
```
bwateratmsft, line 307: Having ingress enabled is a property of an individual container app isn't it?

motm32: Yes, you are correct 😊

This change would disable the command for all container apps. In this case it is probably best to show the command but do the check after a specific container app is chosen.

bwateratmsft: This is a good use case for context values. We can apply a context value to note that ingress is enabled on those container apps for which it is the case, and show the "Disable Ingress" command only for those matching the context value.

src/utils/pickContainerApp.ts
```diff
@@ -9,8 +9,9 @@ import { ext } from "../extensionVariables";
 import { ContainerAppItem } from "../tree/ContainerAppItem";
 
 // TODO: support creating a new container app from picker
-export async function pickContainerApp(context: IActionContext): Promise<ContainerAppItem> {
+export async function pickContainerApp(context: IActionContext, exclude?: string | RegExp | (string | RegExp)[]): Promise<ContainerAppItem> {
     return await azureResourceExperience<ContainerAppItem>(context, ext.rgApiV2.resources.azureResourceTreeDataProvider, AzExtResourceType.ContainerAppsEnvironment, {
         include: ContainerAppItem.contextValueRegExp,
+        exclude,
```
alexweininger, line 15: This excludes items from being final picks, however in this case, the ideal experience is to exclude ingress disabled Container Apps from being ancestor picks, and final picks.

I think applying that change to the `exclude` property would be a breaking change. So I'm thinking about adding new ancestor-pick-specific filter options. They can be optional.

PR #235, opened 2023-01-21
Rewrite extension for v2 API

alexweininger:
Apart from the commands which don't have picking support, everything _should_ be working as expected. I will leave that for CTI to determine, but I've tested most things as I went pretty diligently.

Todos are marked with `// TODO` comments (I know I'm a genius 🤓)

The largest todos are:
1. Picking things for a revision. We need to either pick a container app in single revisions mode, or a revision of a container app in Multiple revisions mode. This is tricky since a revision items are (grand) children of container app items. Haven't put too much effort into this, should be relatively straight forward.
2. Implement a "runWithTemporaryChild" utility that allows us to show a "Creating {0}..." child on an item while the child is being created.


package.json
```diff
@@ -27,31 +27,17 @@
         "multi-root ready"
     ],
     "preview": true,
-    "activationEvents": [
```
bwateratmsft, line 30: What's the minimum VSCode version for this pattern for activation events? Might need to update here.

alexweininger: It was introduced in 1.74, the latest update Are we ok to require the latest? Or should I revert this change?

bwateratmsft: I'm OK either way

src/commands/ingress/editTargetPort.ts
```diff
@@ -0,0 +1,41 @@
...
+import { ContainerAppItem } from "../../tree/ContainerAppItem";
+import { createSubscriptionContext } from "../../tree/ContainerAppsBranchDataProvider";
+import { IngressItem } from "../../tree/IngressItem";
+import { localize } from "../../utils/localize";
+import { pickContainerApp } from "../../utils/pickContainerApp";
+import { IContainerAppContext } from "../createContainerApp/IContainerAppContext";
+import { TargetPortStep } from "../createContainerApp/TargetPortStep";
+import { updateIngressSettings } from "./updateIngressSettings";
+
+export async function editTargetPort(context: IActionContext, node?: IngressItem): Promise<void> {
```
bwateratmsft, line 16: I imagine a lot of this is just files that have been split up, renamed, etc. but aren't actually new code. Is it practical to try to denote which ones are actually new code--and should be reviewed closer?

alexweininger: I did split up and rename a lot of the commands, however the implementations _did_ change too. So I would review those closely. 

Rewriting the tree items was a lot more of "copy this, copy that". So I'd say you can gloss over the tree item stuff. Potential mistakes there will be visible on the UI and should be easily caught by us or CTI.

src/tree/TreeItemState.ts
```diff
@@ -0,0 +1,141 @@
...
+    spinner?: boolean;
+    /**
+     * Temporary children to be displayed
+     */
+    temporaryChildren?: TreeElementBase[];
+}
+
+type ResourceGroupsItem = TreeElementBase & { id: string };
+
+export class TreeItemStateStore implements vscode.Disposable {
```
bwateratmsft, line 27: Seems like we should do some code sharing with this class/file.

alexweininger: Yes. I will put an implementation in utils. Planning on making it cover our use cases, but allow for clients to extend if needed.

PR #267, opened 2023-02-23
Add Container App Environment refresh for deploy 

MicroFish91:
Children of the CA were updating, but view properties on the Container App wasn't, so switching the node we refresh to the container app environment


src/commands/deployImage/deployImage.ts
```diff
@@ -95,7 +96,7 @@ export async function deployImage(context: ITreeItemPickerContext & Partial<IDep
             void showContainerAppCreated(updatedContainerApp, true);
         });
 
-        refreshContainerApp(containerApp.id);
+        refreshContainerAppEnvironment(containerApp.managedEnvironmentId);
```
alexweininger, line 99: I think you can just change this to 

```ts
ext.state.notifyChildrenChanged(containerApp.managedEnvironmentId);
```

And avoid making all the other changes.

MicroFish91:
Alex has a PR ready instead

PR #274, opened 2023-02-27
Set values for quickstart image

### LINKED ISSUE #272: Implement the `Use quickstart image` method when creating a Container App

motm32:
Fixes #272.

### DIFF
--- a/src/commands/createContainerApp/setQuickStartImage.ts
+++ b/src/commands/createContainerApp/setQuickStartImage.ts
@@ -1,0 +1,13 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { IContainerAppContext } from "./IContainerAppContext";
+
+export function setQuickStartImage(context: IContainerAppContext): void {
+    context.image = 'mcr.microsoft.com/azuredocs/aci-helloworld';
+    context.enableIngress = true;
+    context.enableExternal = true;
+    context.targetPort = 80;
+}


bwateratmsft (approved):

src/commands/createContainerApp/setQuickStartImage.ts
```diff
@@ -0,0 +1,13 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { IContainerAppContext } from "./IContainerAppContext";
+
+export function setQuickStartImage(context: IContainerAppContext): void {
+    context.image = 'mcr.microsoft.com/azuredocs/aci-helloworld';
```
MicroFish91, line 9: Minor nit, but the portal uses: 'mcr.microsoft.com/azuredocs/containerapps-helloworld:latest'

Wondering if we want to try and keep the quickstart image sources the same 🤔

motm32: Yeah probably would be good to be consistent I'll change that!

bwateratmsft (approved):

src/commands/createContainerApp/IngressVisibilityStep.ts
```diff
@@ -17,6 +17,6 @@ export class IngressVisibilityStep extends AzureWizardPromptStep<IContainerAppCo
     }
 
     public shouldPrompt(context: IContainerAppContext): boolean {
-        return context.enableIngress === true;
+        return context.enableExternal === undefined;
```
MicroFish91, line 20: Another nit, but just to minimize the potential for side effects, I would probably keep the original statement as well.  Maybe something like:

`return context.enableIngress === true && context.enableExternal === undefined;`

nturinski (approved):

nturinski (commented):
Well approved after Matt's feedback

MicroFish91 (approved):
🚀🥳

PR #277, opened 2023-03-01
Added a more generic deploy command 

### LINKED ISSUE #275: Add a more generic deploy command
Add a more generic deploy command so that we essentially can leverage the following image sourcing options:

<image src="https://user-images.githubusercontent.com/40250218/224411631-dd990328-73ac-4f5c-b410-ba50229bdedd.png" height="400px" />

Note:  At the time of implementation, the current `deployImage` command should probably be renamed to something like `deployImageFromRegistry` or similar.  That will likely be fixed by this issue: https://github.com/microsoft/vscode-azurecontainerapps/issues/276

motm32:
Fixes #275. 

Added "Deploy Container App..." to the cloud button in the workspace. 
image

Also added a prompt for different deployment methods. 
image


MicroFish91:
- One thing I've been thinking about is if there's a way we can refactor the existing `ImageSourceListStep` to be reusable here.  It does what we want mostly, which is to prompt for and set `context.imageSource`.  The main issue is that it currently doesn't support a way to filter out `QuickStartImage`.... 
- Since right now we only support one option for deploy, we can probably auto set the `context.imageSource` option until we implement the other methods

src/commands/deployImage/deploy.ts
```diff
@@ -0,0 +1,34 @@
...
+    const promptSteps: AzureWizardPromptStep<IDeployImageContext>[] = [];
+    const wizardContext: IDeployImageContext = context;
+    const wizard: AzureWizard<IDeployImageContext> = new AzureWizard(wizardContext, {
+        promptSteps
+    });
+
+    context.targetContainer = await getContainerApp(context); //may need to change so it is more easy to add create
+
+    const useQuickStartImage: boolean = false
+    promptSteps.push(new ImageSourceListStep({ useQuickStartImage }));
```
bwateratmsft, line 23: I think this might technically work because of pass-by-reference, but it'd be clearer to assemble the prompt steps before constructing the wizard.

motm32:
Based on offline discussion a couple of changes were made:
- Deploy now utilizes `ImageSourceListStep` and the quickstart image pick is filtered out when deploying
- The first step in deploy is now choosing a container app instead of method of deployment to be consistent with our other extensions 
- Added a "Update an image in container app" pick when deploying if an image is detected

src/commands/createContainerApp/createContainerApp.ts
```diff
@@ -32,10 +32,11 @@ export async function createContainerApp(context: IActionContext & Partial<ICrea
     };
 
     const title: string = localize('createContainerApp', 'Create Container App');
+    const useQuickStartImage: boolean = true;
 
     const promptSteps: AzureWizardPromptStep<IContainerAppWithActivityContext>[] = [
         new ContainerAppNameStep(),
-        new ImageSourceListStep(),
+        new ImageSourceListStep({ useQuickStartImage }),
```
MicroFish91, line 39: I think it would be nice to not have to pass the optional parameter when trying to obtain the default behavior.  If we change the options naming to something like `hideQuickStartImage` instead of `useQuickStartImage`, we are then able to show the option by default even when no options parameter is passed

src/commands/createContainerApp/ImageSourceListStep.ts
```diff
@@ -11,21 +11,36 @@ import { EnvironmentVariablesListStep } from "./EnvironmentVariablesListStep";
 import { IContainerAppContext } from './IContainerAppContext';
 import { setQuickStartImage } from "./setQuickStartImage";
 
+interface ImageSourceOptions {
+    useQuickStartImage?: boolean;
+}
 export class ImageSourceListStep extends AzureWizardPromptStep<IContainerAppContext> {
```
MicroFish91, line 17: Nit:  add a space between the class and the interface

src/commands/deployImage/IDeployImageContext.ts
```diff
@@ -6,9 +6,11 @@
 import type { ContainerApp, EnvironmentVar } from "@azure/arm-appcontainers";
 import type { ContainerRegistryManagementModels } from '@azure/arm-containerregistry';
 import { ISubscriptionActionContext } from '@microsoft/vscode-azext-utils';
-import { SupportedRegistries } from '../../constants';
+import { ImageSourceValues, SupportedRegistries } from '../../constants';
 
 export interface IDeployImageContext extends ISubscriptionActionContext {
+    deployMethod?: ImageSourceValues;
```
MicroFish91, line 12: Wondering if we can just use `imageSource` instead of `deployMethod` since `ImageSourceListStep` already handles setting this value for us.

src/commands/createContainerApp/ImageSourceListStep.ts
```diff
@@ -11,21 +11,36 @@ import { EnvironmentVariablesListStep } from "./EnvironmentVariablesListStep";
...
+            localize('updateImage', 'Update image in Container App')
             // localize('buildFromProject', 'Build from project'),
         ];
 
         const placeHolder: string = localize('imageBuildSourcePrompt', 'Select an image source for the container app');
         const picks: IAzureQuickPickItem<ImageSourceValues | undefined>[] = [
-            { label: imageSourceLabels[0], data: ImageSource.QuickStartImage, suppressPersistence: true },
-            { label: imageSourceLabels[1], data: ImageSource.ExternalRegistry, suppressPersistence: true },
+            { label: imageSourceLabels[0], data: ImageSource.ExternalRegistry, suppressPersistence: true },
             // { label: imageSourceLabels[2], data: undefined, suppressPersistence: true },
```
MicroFish91, line 33: Nit:  Could you change my old comment's index from 2 to 3 please 🙂

src/commands/createContainerApp/ImageSourceListStep.ts
```diff
@@ -11,21 +11,36 @@ import { EnvironmentVariablesListStep } from "./EnvironmentVariablesListStep";
...
         ];
 
+        if (this.options?.useQuickStartImage) {
+            picks.unshift({ label: imageSourceLabels[1], data: ImageSource.QuickStartImage, suppressPersistence: true });
+        } else {
+            if (context.targetContainer?.template?.containers?.[0]?.image) {
+                picks.push({ label: imageSourceLabels[2], data: ImageSource.ExternalRegistry, suppressPersistence: true });
```
MicroFish91, line 40: I just created an issue, could you please link it as a TODO above the `imageSourceLabels[2]` pick 🙂

https://github.com/microsoft/vscode-azurecontainerapps/issues/280

src/commands/deployImage/deploy.ts
```diff
@@ -0,0 +1,34 @@
...
+export async function deployContainerApp(context: IDeployImageContext): Promise<void> {
+    const promptSteps: AzureWizardPromptStep<IDeployImageContext>[] = [];
+    const wizardContext: IDeployImageContext = context;
+    const wizard: AzureWizard<IDeployImageContext> = new AzureWizard(wizardContext, {
+        promptSteps
+    });
+
+    context.targetContainer = await getContainerApp(context); //may need to change so it is more easy to add create
+
+    const useQuickStartImage: boolean = false
```
MicroFish91, line 22: Nit: I prefer this written in one line :)

`new ImageSourceListStep({ useQuickStartImage: false });`

src/commands/deployImage/IDeployImageContext.ts
motm32: I think this is something I forgot to remove from when I had a fully separate deploy step from `ImageSourceListStep`.

src/commands/createContainerApp/ImageSourceListStep.ts
MicroFish91: Update: Based on Anthony's feedback, we can probably just remove this pick and the issue entirely

src/commands/createContainerApp/createContainerApp.ts
nturinski: I actually prefer the opposite way-- we should have to opt-in to showing the quickstart image. Right now, we show/hide it 50% of the time in either direction, but I feel like there are likely more scenarios where we would hide the image rather than display it.

But I think either way, it doesn't really impact the code that much. It's a half-full half-empty scenario 🥤

MicroFish91: Agreed, I had actually already talked with Megan and we decided on doing what you just mentioned.  I originally had the file under the create folder, but we ended up moving it to the deploy folder anyway, so it makes more sense to show instead of hide 🙂

bwateratmsft (approved):

src/commands/deploy/ImageSourceListStep.ts
```diff
@@ -15,18 +15,21 @@ import { DeployFromRegistryConfigureStep } from "./deployFromRegistry/DeployFrom
...
 
         const placeHolder: string = localize('imageBuildSourcePrompt', 'Select an image source for the container app');
         const picks: IAzureQuickPickItem<ImageSourceValues | undefined>[] = [
-            { label: imageSourceLabels[0], data: ImageSource.QuickStartImage, suppressPersistence: true },
-            { label: imageSourceLabels[1], data: ImageSource.ExternalRegistry, suppressPersistence: true },
+            { label: imageSourceLabels[0], data: ImageSource.ExternalRegistry, suppressPersistence: true },
             // { label: imageSourceLabels[2], data: undefined, suppressPersistence: true },
         ];
 
+        if (context.useQuickStartImage !== undefined) {
```
MicroFish91, line 29:
```suggestion
        if (context.useQuickStartImage) {
```

MicroFish91: otherwise a value of `false` may evaluate the expression to `true`

src/commands/deploy/IDeployBaseContext.ts
```diff
@@ -14,6 +14,7 @@ export interface IDeployBaseContext extends ISubscriptionActionContext {
     targetContainer?: ContainerAppModel;
 
     imageSource?: ImageSourceValues;
+    useQuickStartImage?: boolean;
```
MicroFish91, line 17: Are you open to using `showQuickStartImage` or similar instead of `useQuickStartImage`?  I feel like when I see the word "use" I'm expecting that we are defaulting to that choice rather that just offering the option as a pick.

MicroFish91 (approved):
I left 2 comments above to look at.

Otherwise, looks good 🙂🚀

src/commands/deploy/deploy.ts
```diff
@@ -0,0 +1,52 @@
...
+    const { subscription, containerApp } = node;
+
+    const wizardContext: IDeployBaseContext = {
+        ...context,
+        ...createSubscriptionContext(subscription),
+        subscription,
+        targetContainer: containerApp
+    };
+
+    const title: string = localize('deploy', 'Deploying to "{0}"', containerApp.name);
```
alexweininger, line 31: It looks like this variable is only used in one place below on line 44. Can we just create this value inline on line 44 instead of creating a constant for it here?

alexweininger (approved):

Issue #276, opened 2023-02-27
Refactor `deployImage`


MicroFish91:
Refactor `deployImage`:

1. Set up a pattern so that code can be re-used in both `Create...` and `Deploy...`
2. Make sure all image deploy code is encapsulated into wizard steps so they can also be reused during `createContainerApp`
3. Rename `deployImage` and associated files to something more akin to `deployImageFromRegistry`

MicroFish91:
Fixed by #283

PR #281, opened 2023-03-06
Consolidate `deployImage` related logic into wizard steps and introduce base deployment pattern

MicroFish91:
### Introduction

This is the first PR in a multiple part series to increase `deployImage` code re-usability.  This partially addresses:  #276

### Implementation

In this PR, I'm consolidating all `deployImage` logic into wizard steps and introducing a new `IDeployBaseContext` that all deploy scenarios can extend from. 

These attributes should be configured by the deploy steps such that the envelope can then be handed off to either the  `Create` or `Update` Container App steps.

image

### Notes

- The base context currently extends `ISubscriptionActionContext` and includes a `subscription` (Azure Subscription) attribute.  I didn't like including both because it felt a bit redundant, but the `VerifyProvidersStep` currently requires `ISubscriptionActionContext` while most of the Container Apps repository builds the subscription context using `AzureSubscription` as a passed in precursor.
- Right now I have `deployImage` and `deploy` are under two totally separate folders.  I plan to rename `deployImage` to something different and to stick all its sub-logic under the new `deploy` directory.  I refrained from doing so in this PR because I felt that it would pollute readability during review
- Expect at least two more PRs: 1) one for converting file names and general re-organizing, 2) another for refactoring the `createContainerApp` code to use the new pattern


bwateratmsft (approved):

src/commands/deployImage/getRegistryCredentialsAndSecrets.ts
```diff
@@ -3,24 +3,24 @@
...
 
-export async function getAcrCredentialsAndSecrets(context: IDeployImageContext, containerAppEnvelope: Required<ContainerApp>): Promise<RegistryCredentialsAndSecrets> {
+export async function getAcrCredentialsAndSecrets(context: IDeployImageContext, containerAppSettings?: RegistryCredentialsAndSecrets): Promise<RegistryCredentialsAndSecrets> {
     const registry = nonNullProp(context, 'registry');
     const { username, password } = await listCredentialsFromRegistry(context, registry);
     const passwordName = `${registry.name?.toLocaleLowerCase()}-${password?.name}`;
 
     // Remove duplicate registries
-    const registries: RegistryCredentials[] | undefined = containerAppEnvelope.configuration.registries?.filter(r => r.server !== registry.loginServer);
+    const registries: RegistryCredentials[] = containerAppSettings?.registries?.filter(r => r.server !== registry.loginServer) ?? [];
```
nturinski, line 23: Was this a necessary change? I don't know if `undefined` and `[]` behave the exact same in the context of a container app site envelope.

MicroFish91: I bet we could go either way.  It appears to me that the two are probably interchangeable, but the reason I added that is because there is a section in `ContainerAppCreateStep` that looks like this:

image

This would preserve that assignment behavior

MicroFish91:
> Most of this is just moving/refactoring code, right?

Yup, the main changes are the one you caught above, the new `IDeployBaseContext`, and the two steps:  `DeployImageConfigureStep`, `ContainerAppUpdateStep`

nturinski (approved):

Issue #286, opened 2023-03-13
Implement `Build from project remotely using Azure` command


motm32:
This is a choice when creating a container app: 

![image](https://user-images.githubusercontent.com/59709511/224838757-50ccbca9-f4a7-4907-810c-ce59c909f3fa.png)

PR #287, opened 2023-03-13
Integrate docker `buildImageInAzure` code

motm32:
Integrated code from [buildImageInAzure ](https://github.com/microsoft/vscode-docker/blob/main/src/commands/registries/azure/tasks/buildImageInAzure.ts) and [scheduleRunRequest ](https://github.com/microsoft/vscode-docker/blob/main/src/commands/registries/azure/tasks/scheduleRunRequest.ts) which will be used for the 'Build from project remotely using Azure' choice when creating a new container app.

Contributes to #286.


src/commands/buildImageInAzure/RunStep.ts
```diff
@@ -0,0 +1,40 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { DockerBuildRequest as AcrDockerBuildRequest } from "@azure/arm-containerregistry";
+import { AzureWizardExecuteStep } from "@microsoft/vscode-azext-utils";
+import * as fse from 'fs-extra';
+import { IBuildImageContext } from "./IBuildImageContext";
+import path = require("path");
```
MicroFish91, line 10:
```suggestion
import * as path from 'path';
```

src/commands/buildImageInAzure/IBuildImageContext.ts
```diff
@@ -0,0 +1,25 @@
...
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { ContainerRegistryManagementClient, Run as AcrRun } from '@azure/arm-containerregistry';
+import * as vscode from 'vscode';
+import { IContainerAppContext } from '../createContainerApp/IContainerAppContext';
+import { IDeployFromRegistryContext } from '../deploy/deployFromRegistry/IDeployFromRegistryContext';
+import { Item } from "./DockerFileItemStep";
+
+export interface IBuildImageContext extends IContainerAppContext, IDeployFromRegistryContext {
```
MicroFish91, line 12: Some potential suggestions:

1) Move `buildImageInAzure/` to sit under `deploy/` (related to all the deploy image code)

2) Instead of extending `IContainerAppContext` and `IDeployFromRegistryContext`, extend from `IDeployBaseContext`

3) Rename `IBuildImageContext` to match the directory name.  So something like: `IBuildImageInAzureContext`

src/commands/buildImageInAzure/DockerFileItemStep.ts
```diff
@@ -0,0 +1,41 @@
...
+        context.dockerFile = createFileItem(context.rootFolder, selectedDockerFile);
+    }
+
+    public shouldPrompt(context: IBuildImageContext): boolean {
+        return !context.dockerFile;
+    }
+}
+
+function createFileItem(rootFolder: vscode.WorkspaceFolder, selectedDockerFile: string): Item {
+    const relativeFilePath = path.join(".", selectedDockerFile.substr(rootFolder.uri.fsPath.length));
```
nturinski, line 32: Can we replace path with Utils?

Here's an example: https://github.com/microsoft/vscode-azurestaticwebapps/pull/801/files#diff-5d1a282c2932eed334e91bcc823026bd3ee0f4c259ca8be3a509d89aff82cfe0R59

bwateratmsft:
> Integrated code from [buildImageInAzure ](https://github.com/microsoft/vscode-docker/blob/main/src/commands/registries/azure/tasks/buildImageInAzure.ts) and [scheduleRunRequest ](https://github.com/microsoft/vscode-docker/blob/main/src/commands/registries/azure/tasks/scheduleRunRequest.ts) which will be used for the 'Build from project remotely using Azure' choice when creating a new container app.
> 
> Contributes to #286.

Why not just call into the Docker extension instead?

motm32:
> Why not just call into the Docker extension instead? Or is the aim to move that code to here? (Actually I rather like the idea of moving that logic here)

We originally decided it was best to migrate the code to Container Apps so that we could use it in .dev. We also wanted to put everything in prompt/execute steps so users can go back in steps.

src/commands/deploy/buildImageInAzure/tar.d.ts
```diff
@@ -0,0 +1,78 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See LICENSE.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+/**
+ * The `@types/tar` package is riddled with inaccuracies, so it's easier
+ * to just declare typings for it here
+ */
+declare module "tar" {
```
bwateratmsft, line 10: I'd include a comment with a link to note that this is originally from the Docker extension. Not worth the effort to make a package though, IMO.

src/constants.ts
```diff
@@ -68,3 +68,5 @@ export type QuickPicksCache = { cache: QuickPickItem[], next: string | null };
 
 export const azResourceContextValue: string = 'azResource';
 export const azResourceRegExp = new RegExp(azResourceContextValue, 'i');
+
+export const DOCKERFILE_GLOB_PATTERN = '**/{*.[dD][oO][cC][kK][eE][rR][fF][iI][lL][eE],[dD][oO][cC][kK][eE][rR][fF][iI][lL][eE],[dD][oO][cC][kK][eE][rR][fF][iI][lL][eE].*}';
```
bwateratmsft, line 72: I'd include a comment with a link to note that this is originally from the Docker extension. Not worth the effort to make a package though, IMO.

bwateratmsft (approved):

src/commands/deploy/buildImageInAzure/IBuildImageInAzureContext.ts
```diff
@@ -0,0 +1,23 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { Run as AcrRun, ContainerRegistryManagementClient } from '@azure/arm-containerregistry';
```
nturinski, line 6:
```suggestion
import type { Run as AcrRun, ContainerRegistryManagementClient } from '@azure/arm-containerregistry';
```

I think?

src/commands/deploy/buildImageInAzure/IBuildImageInAzureContext.ts
```diff
@@ -0,0 +1,23 @@
...
+import * as vscode from 'vscode';
+import { IDeployFromRegistryContext } from '../deployFromRegistry/IDeployFromRegistryContext';
+
+export interface IBuildImageInAzureContext extends IDeployFromRegistryContext {
+    rootFolder: vscode.WorkspaceFolder;
+    dockerFilePath: string;
+    imageName: string;
+    os: string;
```
nturinski, line 14: For additional type-safety/clarity, you could make this type:
```suggestion
    os: 'Windows' | 'Linux';
```
Doing this will only allow us to type one or the other, not just any string.

src/commands/deploy/buildImageInAzure/RootFolderStep.ts
```diff
@@ -0,0 +1,33 @@
...
+    public async prompt(context: IBuildImageInAzureContext): Promise<void> {
+        context.rootFolder = await getRootWorkSpaceFolder();
+    }
+
+    public shouldPrompt(context: IBuildImageInAzureContext): boolean {
+        return !context.rootFolder;
+    }
+}
+
+export async function getRootWorkSpaceFolder(): Promise<vscode.WorkspaceFolder> {
```
nturinski, line 20: Do you need to export this? I don't see it being imported anywhere.

If it is, I think we should probably move it to the `workspaceUtils`.

src/commands/deploy/buildImageInAzure/RunStep.ts
```diff
@@ -0,0 +1,38 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { DockerBuildRequest as AcrDockerBuildRequest } from "@azure/arm-containerregistry";
```
nturinski, line 6:
```suggestion
import type { DockerBuildRequest as AcrDockerBuildRequest } from "@azure/arm-containerregistry";
```

src/commands/deploy/buildImageInAzure/RunStep.ts
```diff
@@ -0,0 +1,38 @@
...
+            const runRequest: AcrDockerBuildRequest = {
+                type: 'DockerBuildRequest',
+                imageNames: [context.imageName],
+                isPushEnabled: true,
+                sourceLocation: context.uploadedSourceLocation,
+                platform: { os: context.os },
+                dockerFilePath: path.relative(rootUri.path, context.dockerFilePath)
+            };
+
+            context.run = await context.client.registries.beginScheduleRunAndWait(context.resourceGroupName, context.registryName, runRequest);
```
nturinski, line 27: How long is this operation? If it takes a few seconds+, I think we should have an activity message to update on the progress.

src/commands/deploy/buildImageInAzure/UploadSourceCodeStep.ts
```diff
@@ -0,0 +1,41 @@
...
+export class UploadSourceCodeStep extends AzureWizardExecuteStep<IBuildImageInAzureContext> {
+    public priority: number = 175;
+
+    public async execute(context: IBuildImageInAzureContext): Promise<void> {
+        context.registryName = nonNullValue(context.registry?.name);
+        context.resourceGroupName = getResourceGroupFromId(nonNullValue(context.registry?.id));
+        context.client = await createContainerRegistryManagementClient(context);
+
+        const source: string = context.rootFolder.uri.fsPath;
+        let items = await fse.readdir(source);
```
nturinski, line 23: There should be an equivalent function for `readdir` in `AzExtFsExtra`

src/commands/deploy/buildImageInAzure/UploadSourceCodeStep.ts
```diff
@@ -0,0 +1,41 @@
...
+        items = items.filter(i => !(i in vcsIgnoreList));
+        tar.c({ cwd: source }, items).pipe(fse.createWriteStream(context.tarFilePath));
```
nturinski, line 25: Not `createWriteStream` though. This is something that we'll have to figure out at one point though 😅

src/commands/deploy/buildImageInAzure/UploadSourceCodeStep.ts
```diff
@@ -0,0 +1,41 @@
...
+
+        const sourceUploadLocation = await context.client.registries.getBuildSourceUploadUrl(context.resourceGroupName, context.registryName);
+        const uploadUrl: string = nonNullValue(sourceUploadLocation.uploadUrl);
+        const relativePath: string = nonNullValue(sourceUploadLocation.relativePath);
+
+        const storageBlob = await import('@azure/storage-blob');
```
nturinski, line 31: I think it might be better to have this outside of the class declaration. I think if we leave it in here, it keeps getting instantiated and we'll have more unnecessary async calls to import the module.

src/commands/deploy/buildImageInAzure/buildImageInAzure.ts
```diff
@@ -0,0 +1,28 @@
...
+import { nonNullValue } from '@microsoft/vscode-azext-utils';
+import { IBuildImageInAzureContext } from "./IBuildImageInAzureContext";
+
+const WAIT_MS = 5000;
+
+export async function buildImageInAzure(context: IBuildImageInAzureContext): Promise<AcrRun | undefined> {
+    const getRun = async () => context.client.runs.get(context.resourceGroupName, context.registryName, nonNullValue(context.run.runId));
+
+    let run = await getRun();
+    const { KnownRunStatus } = await import('@azure/arm-containerregistry');
```
nturinski, line 17: This could probably be imported at the top as `import type`

src/commands/deploy/buildImageInAzure/BuildImageStep.ts
```diff
@@ -0,0 +1,28 @@
...
+export class BuildImageStep extends AzureWizardExecuteStep<IBuildImageInAzureContext> {
+    public priority: number = 225;
+
+    public async execute(context: IBuildImageInAzureContext): Promise<void> {
+        context.registryDomain = acrDomain;
+
+        const run = await buildImageInAzure(context);
+        const outputImages = run?.outputImages;
+        if (outputImages) {
+            const image = outputImages[0];
```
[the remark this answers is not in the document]
motm32: The run generates a list of all images generated during the run. I think in our case there should only be one image generated which was why I took the 0 index.

src/commands/deploy/buildImageInAzure/UploadSourceCodeStep.ts
```diff
@@ -2,20 +2,19 @@
...
-import { IBuildImageContext } from './IBuildImageContext';
+import { createContainerRegistryManagementClient } from '../../../utils/azureClients';
+import { IBuildImageInAzureContext } from './IBuildImageInAzureContext';
 
 const vcsIgnoreList = ['.git', '.gitignore', '.bzr', 'bzrignore', '.hg', '.hgignore', '.svn'];
 
-export class UploadSourceCodeStep extends AzureWizardExecuteStep<IBuildImageContext> {
-    public priority: number = 200;
+export class UploadSourceCodeStep extends AzureWizardExecuteStep<IBuildImageInAzureContext> {
+    public priority: number = 175;
```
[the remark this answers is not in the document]
motm32: Yeah this would be helpful especially when there are so many steps.

src/commands/deploy/buildImageInAzure/RootFolderStep.ts
motm32: Not using it anywhere else 😄 I'll remove the export.

src/commands/deploy/buildImageInAzure/buildImageInAzure.ts
motm32: It would be`import` not `import type` I think?

nturinski: Oh yes, I think you're right. It's an enum, not a typing.

src/commands/deploy/buildImageInAzure/UploadSourceCodeStep.ts
nturinski: Yeah sure. I didn't see this before the other comment

src/commands/deploy/buildImageInAzure/UploadSourceCodeStep.ts
nturinski: Actually, if I remember correctly, when you await an import, it lazy-loads it and only does this once. So that means when it comes to this line of code again, the promised will already be resolved. So disregard my comment.

src/commands/deploy/buildImageInAzure/BuildImageStep.ts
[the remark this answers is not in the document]
nturinski: This seems to be how the AKS team does it as well. I'm not super concerned about additional images being created at this time. But could you add some telemetry to track the length of outputImages? If we see there are a lot of users with more than 1, we may want to revisit this.

src/commands/deploy/buildImageInAzure/RunStep.ts
motm32: It takes a little bit so I'll add an activity message to update on progress.

alexweininger:
Have we already or are we planning to have an activity in the activity log for this?

motm32:
> Have we already or are we planning to have an activity in the activity log for this?

This doesn't have it's own activity log but I added this to the progress in the "Create Container App..." activity to signal that the image is being built. 
image

nturinski:
From your progress message, I'd just leave feedback it might be worth saying in registry "{1}" because the name doesn't make it immediately obvious what that name is.

alexweininger:
It'd be great to have some richer features in the activity log for some of the super-wizards we have. Maybe sub activities or something.

src/commands/deploy/buildImageInAzure/RunStep.ts
```diff
@@ -0,0 +1,43 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import type { DockerBuildRequest as AcrDockerBuildRequest } from "@azure/arm-containerregistry";
+import { AzExtFsExtra, AzureWizardExecuteStep } from "@microsoft/vscode-azext-utils";
+import * as path from 'path';
```
nturinski, line 8: I think that this is the only file still using path.

nturinski: Actually, don't worry about it. I don't think `dockerFilePath` is used to read any file which is the only reason we'd need to use Uris. We actually have a browser version of path, so `relative` should still work in this case.

If we ever wanted to implement our own relative in VSCode, here's the node implementation: https://github.com/nodejs/node/blob/main/lib/path.js#L491-L616

But doesn't feel worth it especially in this case.

nturinski (approved):

PR #290, opened 2023-03-17
Connect buildImageInAzure to Create Container App

motm32:
Connects code from #287 to the create path. 

image

Contributes to #286

### DIFF
--- a/src/commands/deploy/buildImageInAzure/BuildFromProjectListStep.ts
+++ b/src/commands/deploy/buildImageInAzure/BuildFromProjectListStep.ts
@@ -18,23 +18,25 @@ import { TarFileStep } from "./TarFileStep";
 import { UploadSourceCodeStep } from "./UploadSourceCodeStep";
 
 const buildImageInAzure = 'buildImageInAzure';
+const buildFromProjectLabels: string[] = [
+    localize('azure', 'Build from project remotely using Azure Container Registry')
+    //localize('docker', 'Build from project locally using Docker')
+];
+
 export class BuildFromProjectListStep extends AzureWizardPromptStep<IDeployBaseContext> {
     public async prompt(context: IDeployBaseContext): Promise<void> {
-        const buildFromProjectLabels: string[] = [
-            localize('azure', 'Build from project remotely using Azure Container Registry')
-            //localize('docker', 'Build from project locally using Docker')
+        const placeHolder: string = localize('buildType', 'Select how you want to build your project');
+        const picks: IAzureQuickPickItem<string | undefined>[] = [
+            { label: buildFromProjectLabels[0], data: buildImageInAzure, suppressPersistence: true },
+            //{ label: buildFromProjectLabels[1], data: buildFromProjectLabels[1], suppressPersistence: true }
         ];
 
+        context.buildType = (await context.ui.showQuickPick(picks, { placeHolder })).data;
+    }
+
+    public async configureBeforePrompt(context: IDeployBaseContext): Promise<void> {
         if (buildFromProjectLabels.length === 1) {
             context.buildType = buildImageInAzure;
-        } else {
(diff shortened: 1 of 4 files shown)


src/commands/deploy/buildImageInAzure/BuildFromProjectListStep.ts
```diff
@@ -0,0 +1,58 @@
...
+    public async getSubWizard(context: IBuildImageInAzureContext): Promise<IWizardOptions<IDeployBaseContext> | undefined> {
+        const promptSteps: AzureWizardPromptStep<IDeployBaseContext>[] = [];
+        const executeSteps: AzureWizardExecuteStep<IDeployBaseContext>[] = [];
+
+        switch (context.buildType) {
+            case buildImageInAzure:
+                promptSteps.push(new AcrListStep(), new RootFolderStep(), new DockerFileItemStep(), new ImageNameStep(), new OSPickStep());
+                executeSteps.push(new TarFileStep(), new UploadSourceCodeStep(), new RunStep(), new BuildImageStep());
+                break;
+            //TODO: case for 'Build from project locally using Docker'
```
bwateratmsft, line 54: For local build, I'd recommend calling into the Docker extension's build command. Users will expect customizations applied to the Docker extension to work, and there is _extensive_ customizability available for build.

src/commands/deploy/buildImageInAzure/BuildFromProjectListStep.ts
```diff
@@ -0,0 +1,58 @@
...
+import { OSPickStep } from "./OSPickStep";
+import { RootFolderStep } from "./RootFolderStep";
+import { RunStep } from "./RunStep";
+import { TarFileStep } from "./TarFileStep";
+import { UploadSourceCodeStep } from "./UploadSourceCodeStep";
+
+export class BuildFromProjectListStep extends AzureWizardPromptStep<IDeployBaseContext> {
+    public async prompt(context: IDeployBaseContext): Promise<void> {
+        const buildFromProjectLabels: string[] = [
+            localize('azure', 'Build from project remotely using Azure')
```
bwateratmsft, line 24:
```suggestion
            localize('azure', 'Build from project remotely using Azure Container Registry')
```

Maybe this?

src/constants.ts
```diff
@@ -61,6 +61,8 @@ export const acrDomain = 'azurecr.io';
 export const dockerHubDomain = 'docker.io';
 export const dockerHubRegistry = 'index.docker.io';
 
+export const buildImageInAzure = 'buildImageInAzure';
```
MicroFish91, line 64: Nit:  Is this constant used anywhere else outside of `BuildFromProjectListStep`?

If not, I would just declare this constant at the top of the file where it's used and remove it from here.  If we end up needing this in more than one file we can always move it back

src/commands/deploy/buildImageInAzure/BuildFromProjectListStep.ts
```diff
@@ -0,0 +1,58 @@
...
+
+const buildImageInAzure = 'buildImageInAzure';
+export class BuildFromProjectListStep extends AzureWizardPromptStep<IDeployBaseContext> {
+    public async prompt(context: IDeployBaseContext): Promise<void> {
+        const buildFromProjectLabels: string[] = [
+            localize('azure', 'Build from project remotely using Azure Container Registry')
+            //localize('docker', 'Build from project locally using Docker')
+        ];
+
+        if (buildFromProjectLabels.length === 1) {
```
nturinski, line 28: This should go in `configureBeforePrompt`. That way your `shouldPrompt` will make this prompt get skipped and we'll never go into `prompt`. This will mess up the total number of steps which isn't that weird, but that is what `configureBeforePrompt` is made for.

nturinski (approved):

PR #292, opened 2023-03-21
Final "Build Image in Azure"

### LINKED ISSUE #286: Implement `Build from project remotely using Azure` command

motm32:
Fixes #286.


src/commands/deploy/buildImageInAzure/BuildImageStep.ts
```diff
@@ -0,0 +1,30 @@
...
+    public priority: number = 225;
+
+    public async execute(context: IBuildImageInAzureContext): Promise<void> {
+        context.registryDomain = acrDomain;
+
+        const run = await buildImageInAzure(context);
+        const outputImages = run?.outputImages;
+        context.telemetry.properties.outputImages = outputImages?.length?.toString();
+
+        if (outputImages) {
```
nturinski, line 21: I think we should throw an error here if there's not any output image. The whole create scenario will break at the end anyway, so at least we could give them a more descriptive error?

nturinski (approved):

src/commands/deploy/IDeployBaseContext.ts
```diff
@@ -14,6 +14,7 @@ export interface IDeployBaseContext extends ISubscriptionActionContext {
     targetContainer?: ContainerAppModel;
 
     imageSource?: ImageSourceValues;
+    buildType?: string;
```
alexweininger, line 17: This should probably be a stronger type than just `string`. Either an enum or string union.

alexweininger, line 17: Also, we should try to document these context properties where necessary. How are these properties used and what do they control?

src/commands/deploy/buildImageInAzure/BuildFromProjectListStep.ts
```diff
@@ -0,0 +1,60 @@
...
+import { OSPickStep } from "./OSPickStep";
+import { RootFolderStep } from "./RootFolderStep";
+import { RunStep } from "./RunStep";
+import { TarFileStep } from "./TarFileStep";
+import { UploadSourceCodeStep } from "./UploadSourceCodeStep";
+
+const buildImageInAzure = 'buildImageInAzure';
+const buildFromProjectLabels: string[] = [
+    localize('azure', 'Build from project remotely using Azure Container Registry')
+    //localize('docker', 'Build from project locally using Docker')
```
alexweininger, line 23: I probably missed something, but what are we commenting this out until it's ready somewhere else?

motm32: This is related to another create CA scenario (#285). This scenario won't be in this release so just putting a placeholder for now.

nturinski (approved):

PR #293, opened 2023-03-23
Fix filtering in `UploadSourceCodeStep`

motm32:
Added link to logs to error message:
image

Also fixed filtering in `UploadSourceCodeStep`


src/commands/deploy/buildImageInAzure/BuildImageStep.ts
```diff
@@ -23,7 +23,9 @@ export class BuildImageStep extends AzureWizardExecuteStep<IBuildImageInAzureCon
             const image = outputImages[0];
             context.image = `${image.registry}/${image.repository}:${image.tag}`;
         } else {
-            throw new Error(localize('noImagesBuilt', 'Failed to build image.'));
+            const logSasUrl = (await context.client.runs.getLogSasUrl(context.resourceGroupName, context.registryName, nonNullValue(context.run.runId))).logLink;
+            const errorMessage = localize('noImagesBuilt', 'Failed to build image. View [logs]({0}) for more details.', logSasUrl);
```
bwateratmsft, line 27: Should we use a button instead of a link? I don't feel strongly one way or the other but it's worth considering both.

src/commands/deploy/buildImageInAzure/BuildImageStep.ts
```diff
@@ -23,7 +23,9 @@ export class BuildImageStep extends AzureWizardExecuteStep<IBuildImageInAzureCon
...
+            throw new Error(errorMessage);
```
bwateratmsft, line 28: You may want to also do `context.errorHandling.suppressReportIssue = true`.

bwateratmsft (approved):

nturinski (approved):
Don't hate me, but have you considered outputting the log in VS Code as a .log file? 😓If you want to explore that, a new PR would be fine.

EDIT: The idea being that way people don't need to leave VS Code.

nturinski (approved):

PR #303, opened 2023-03-27
Output build image error logs to .log file

motm32:
When an image fails to build logs are outputted to a .log file. 

image


src/commands/imageSource/buildImageInAzure/openImageErrorLog.ts
```diff
@@ -0,0 +1,41 @@
...
+import { nonNullValue } from "@microsoft/vscode-azext-utils";
+import { TextDocumentContentProvider, Uri, window, workspace } from "vscode";
+import { ext } from "../../../extensionVariables";
+import { IBuildImageInAzureContext } from "./IBuildImageInAzureContext";
+
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+const contentScheme: string = 'image-error-log';
```
alexweininger, line 11: I feel that the scheme should be a bit more unique to our extension than this to avoid conflicting with any other providers.

Maybe `azurecontainerapps-imagelogs`?

src/commands/imageSource/buildImageInAzure/openImageErrorLog.ts
```diff
@@ -0,0 +1,41 @@
...
+
+let _cachedContentProvider: ImageLogContentProvider | undefined;
```
alexweininger, line 13: nit: remove underscore

src/commands/imageSource/buildImageInAzure/openImageErrorLog.ts
```diff
@@ -0,0 +1,41 @@
...
+export function getContentProvider(): ImageLogContentProvider {
+    if (!_cachedContentProvider) {
+        _cachedContentProvider = new ImageLogContentProvider();
+        ext.context.subscriptions.push(workspace.registerTextDocumentContentProvider(contentScheme, _cachedContentProvider));
+    }
+    return _cachedContentProvider;
+}
+
+class ImageLogContentProvider implements TextDocumentContentProvider {
+    private _contentMap: Map<string, string> = new Map<string, string>();
```
alexweininger, line 23: nit: remove underscore

alexweininger, line 23: This can be `readonly` too.

src/commands/imageSource/buildImageInAzure/openImageErrorLog.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    public async openImageLog(context: IBuildImageInAzureContext): Promise<void> {
+        const timeout = 15000;
```
alexweininger, line 26: Was the timeout added to fix an issue with the request sometimes taking too long? We might want a comment explaining why a timeout was needed.

alexweininger, line 26: `timeout` is only used once below, I'd just use the value there instead of making a new variable here.

alexweininger (commented):
This could be used to get any image build log, not just error logs, right? If so, later on we might want to provide a way to view logs for an image build operation regardless of the build result.

src/commands/imageSource/buildImageInAzure/openImageErrorLog.ts
motm32: It wasn't added to fix an issue `sendRequestWithTimeout` requires a timeout value. I just used the default value that is used in functions.

nturinski:
I'm totally okay with this approached, but it's quite a bit more complex than I was anticipating...

Sorry for creating so much extra work. What I was suggesting could have been done with just an extra line (which is why I suggested it) but I think I directed you to much more complicated code than I remembered. 

The GitHub action log was made with a file provider because it has text folding incorporated into it. In order to support that feature, we had to make a content provider to handle it.

However, if we just want to open a file in a read-only editor, you can use our `openReadOnlyContent` util command.

That ends up looking something like

```
const logSasUrl = nonNullValue((await context.client.runs.getLogSasUrl(context.resourceGroupName, context.registryName, nonNullValue(context.run.runId))).logLink);
const content = (await sendRequestWithTimeout(context, { method: 'GET', url: logSasUrl }, 2500, undefined)).bodyAsText;
// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
await openReadOnlyContent({ label: context.run.name!, fullId: context.run.id! }, content!, '.log');
```

Since you already did everything, I'm okay to keep the more complicated approach though!

src/commands/imageSource/buildImageInAzure/BuildImageStep.ts
```diff
@@ -23,11 +24,16 @@ export class BuildImageStep extends AzureWizardExecuteStep<IBuildImageInAzureCon
             const image = outputImages[0];
             context.image = `${image.registry}/${image.repository}:${image.tag}`;
         } else {
-            throw new Error(localize('noImagesBuilt', 'Failed to build image.'));
+            const contentProvider = getContentProvider();
+            await contentProvider.openImageLog(context);
```
nturinski, line 28: We should add a button that says "View logs" that lets the user control if the log comes up or not. I think it'll be kind of jarring to have a log file just pop up immediately after an error happens.

Probably can't do this now, but it's a shame once they close the log (or if they don't click it) there's no way for the user to open it later. Definitely seems like a good use-case for the sub-activities thing.

alexweininger:
I'm glad we went with the easier approach 😄

src/commands/imageSource/buildImageInAzure/BuildImageStep.ts
```diff
@@ -19,15 +21,30 @@ export class BuildImageStep extends AzureWizardExecuteStep<IBuildImageInAzureCon
...
         if (outputImages) {
             const image = outputImages[0];
             context.image = `${image.registry}/${image.repository}:${image.tag}`;
         } else {
-            throw new Error(localize('noImagesBuilt', 'Failed to build image.'));
+            const logSasUrl = (await context.client.runs.getLogSasUrl(context.resourceGroupName, context.registryName, nonNullValue(context.run.runId))).logLink;
+            const contentTask = sendRequestWithTimeout(context, { method: 'GET', url: logSasUrl }, 2500, undefined)
+
+            const viewLogsButton: MessageItem = { title: localize('viewLogs', 'View Logs') };
+            void window.showErrorMessage(errorMessage, ...[viewLogsButton]).then(async result => {
```
alexweininger, line 34:
```suggestion
            void window.showErrorMessage(errorMessage, viewLogsButton).then(async result => {
```

alexweininger (approved):

bwateratmsft (approved):

src/commands/imageSource/buildImageInAzure/BuildImageStep.ts
```diff
@@ -23,11 +25,25 @@ export class BuildImageStep extends AzureWizardExecuteStep<IBuildImageInAzureCon
             const image = outputImages[0];
             context.image = `${image.registry}/${image.repository}:${image.tag}`;
         } else {
-            throw new Error(localize('noImagesBuilt', 'Failed to build image.'));
+            const logSasUrl = (await context.client.runs.getLogSasUrl(context.resourceGroupName, context.registryName, nonNullValue(context.run.runId))).logLink;
+            const contentTask = sendRequestWithTimeout(context, { method: 'GET', url: logSasUrl }, 2500, undefined)
+
+            const viewLogsButton: MessageItem = { title: localize('viewLogs', 'View Logs') };
+            const errorMessage = localize('noImagesBuilt', 'Failed to build image. View logs for more details.');
+            void window.showErrorMessage(errorMessage, viewLogsButton).then(async result => {
```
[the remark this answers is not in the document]
motm32: I don't think context.ui has a showErrorMessage only a showWarningMessage which I think could also work since I am throwing an error anyways.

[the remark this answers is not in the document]
alexweininger: The different here between the two will be the icon shown in the notification. Error will be the red ❌  and warning will be ⚠️.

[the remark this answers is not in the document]
motm32: I agree I think it is better to be consistent with the message even if the activity log shows it as an error.

PR #325, opened 2023-03-30
Only add `Build from project remotely...` pick if on Desktop

motm32:
We don't want this pick to show up when using .dev environment. I also tested it on virtual workspaces and that also seems to cause issues with the AcrRun so also not adding the pick when it is a virtual workspace. 

See [here](https://code.visualstudio.com/api/extension-guides/virtual-workspaces#detect-virtual-workspaces-programmatically) for details on detecting virtual workspaces.

### DIFF
--- a/src/commands/imageSource/ImageSourceListStep.ts
+++ b/src/commands/imageSource/ImageSourceListStep.ts
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { AzureWizardExecuteStep, AzureWizardPromptStep, IAzureQuickPickItem, IWizardOptions } from "@microsoft/vscode-azext-utils";
+import { UIKind, env, workspace } from "vscode";
 import { ImageSource, ImageSourceValues } from "../../constants";
 import { localize } from "../../utils/localize";
 import { setQuickStartImage } from "../createContainerApp/setQuickStartImage";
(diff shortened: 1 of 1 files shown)


alexweininger:
> See [here](https://code.visualstudio.com/api/extension-guides/virtual-workspaces#detect-virtual-workspaces-programmatically) for details on detecting virtual workspaces.

TIL. We should make this a utility at some point.

alexweininger (approved):
Do we have plans to try to get this working later on? From the surface I don't see why it can't work (if we put in effort to fixing the issues you encountered here).

motm32:
> Do we have plans to try to get this working later on? From the surface I don't see why it can't work (if we put in effort to fixing the issues you encountered here).

We do have plans to add this later! Just don't have the time right now to fix the current issues before the next release.

alexweininger:
> > Do we have plans to try to get this working later on? From the surface I don't see why it can't work (if we put in effort to fixing the issues you encountered here).
> 
> We do have plans to add this later! Just don't have the time right now to fix the current issues before the next release.

Cool 😎

src/commands/imageSource/ImageSourceListStep.ts
```diff
@@ -23,14 +24,18 @@ export class ImageSourceListStep extends AzureWizardPromptStep<IImageSourceBaseC
...
-            { label: imageSourceLabels[0], data: ImageSource.ContainerRegistry, suppressPersistence: true },
-            { label: imageSourceLabels[2], data: ImageSource.RemoteAcrBuild, suppressPersistence: true },
+            { label: imageSourceLabels[0], data: ImageSource.ContainerRegistry, suppressPersistence: true }
         ];
 
         if (context.showQuickStartImage) {
             picks.unshift({ label: imageSourceLabels[1], data: ImageSource.QuickStartImage, suppressPersistence: true });
         }
 
+        const isVirtualWorkspace = workspace.workspaceFolders && workspace.workspaceFolders.every(f => f.uri.scheme !== 'file');
```
bwateratmsft, line 34: Doesn't have to be addressed in this PR, but this is screaming for a utility method in utils to more generically test what sort of situation you're in--desktop, remote from desktop, remote from browser, just browser, etc.

bwateratmsft (approved):

PR #350, opened 2023-04-19
Add Log Streaming capabilities

### LINKED ISSUE #34: Add log stream
Select a replica and view its log stream. Related #33

motm32:
Fixes #34. 

When clicking `Stop Streaming Logs...` users are prompted to choose which stream they would like to stop. 

image

There is a pick option to stop all streams. I thought it could be helpful to have a warning message pop up to confirm they want to stop all logs since it will immediately stop all logs after this step. Would that be something we would want to add?


src/commands/logStream/ContainerListStep.ts
```diff
@@ -0,0 +1,26 @@
...
+    public async prompt(context: IStreamLogsContext): Promise<void> {
+        const placeHolder: string = localize('selectContainer', 'Select a Container');
+        context.container = (await context.ui.showQuickPick(this.getPicks(context), { placeHolder })).data;
+    }
+
+    public shouldPrompt(context: IStreamLogsContext): boolean {
+        return !context.container;
+    }
+
+    public async getPicks(context: IStreamLogsContext): Promise<IAzureQuickPickItem<ReplicaContainer>[]> {
```
bwateratmsft, line 21:
```suggestion
    private async getPicks(context: IStreamLogsContext): Promise<IAzureQuickPickItem<ReplicaContainer>[]> {
```

src/commands/logStream/ReplicaListStep.ts
```diff
@@ -0,0 +1,37 @@
...
+    public async prompt(context: IStreamLogsContext): Promise<void> {
+        const placeHolder: string = localize('selectReplica', 'Select a Replica');
+        context.replica = (await context.ui.showQuickPick(this.getPicks(context), { placeHolder })).data;
+    }
+
+    public shouldPrompt(context: IStreamLogsContext): boolean {
+        return !context.replica;
+    }
+
+    public async getPicks(context: IStreamLogsContext): Promise<IAzureQuickPickItem<Replica>[]> {
```
bwateratmsft, line 23:
```suggestion
    private async getPicks(context: IStreamLogsContext): Promise<IAzureQuickPickItem<Replica>[]> {
```

src/commands/logStream/RevisionListStep.ts
```diff
@@ -0,0 +1,43 @@
...
+                }
+            });
+        }
+    }
+
+    public shouldPrompt(context: IStreamLogsContext): boolean {
+        return !context.revision;
+    }
+
+    public async getPicks(context: IStreamLogsContext): Promise<IAzureQuickPickItem<Revision>[]> {
```
bwateratmsft, line 33:
```suggestion
    private async getPicks(context: IStreamLogsContext): Promise<IAzureQuickPickItem<Revision>[]> {
```

src/commands/logStream/StreamListStep.ts
```diff
@@ -0,0 +1,36 @@
...
+        const picks: IAzureQuickPickItem<ILogStream | undefined>[] = await this.getPicks(context);
+        picks.push({ label: localize('stopAll', 'Stop all Streams'), data: undefined })
+        context.logStreamToStop = (await context.ui.showQuickPick(picks, { placeHolder, enableGrouping: true })).data;
+    }
+
+    public shouldPrompt(context: IStreamLogsContext): boolean {
+        return !context.logStreamToStop;
+    }
+
+    public async getPicks(context: IStreamLogsContext): Promise<IAzureQuickPickItem<ILogStream>[]> {
```
bwateratmsft, line 23:
```suggestion
    private async getPicks(context: IStreamLogsContext): Promise<IAzureQuickPickItem<ILogStream>[]> {
```

bwateratmsft (approved):

src/commands/logStream/IStreamLogsContext.ts
```diff
@@ -0,0 +1,22 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { Replica, ReplicaContainer, Revision } from "@azure/arm-appcontainers";
```
nturinski, line 6:
```suggestion
import type { Replica, ReplicaContainer, Revision } from "@azure/arm-appcontainers";
```

src/commands/logStream/IStreamLogsContext.ts
```diff
@@ -0,0 +1,22 @@
...
+import { IActionContext } from "@microsoft/vscode-azext-utils";
+import { AzureSubscription } from "@microsoft/vscode-azureresources-api";
```
nturinski, line 8:
```suggestion
import type { AzureSubscription } from "@microsoft/vscode-azureresources-api";
```

nturinski: Just do it for all of them 🙄

src/commands/logStream/ReplicaListStep.ts
```diff
@@ -0,0 +1,37 @@
...
+import { ContainerAppsAPIClient, Replica } from "@azure/arm-appcontainers";
+import { AzureWizardPromptStep, IAzureQuickPickItem, createSubscriptionContext, nonNullProp, nonNullValue } from "@microsoft/vscode-azext-utils";
+import * as dayjs from 'dayjs';
+import { createContainerAppsAPIClient } from "../../utils/azureClients";
+import { localize } from "../../utils/localize";
+import { IStreamLogsContext } from "./IStreamLogsContext";
+
+export class ReplicaListStep extends AzureWizardPromptStep<IStreamLogsContext> {
+    public async prompt(context: IStreamLogsContext): Promise<void> {
+        const placeHolder: string = localize('selectReplica', 'Select a Replica');
```
nturinski, line 15: nit: lowercase replica. We typically only capitalize things that have Azure branding in it

src/commands/logStream/logStreamRequest.ts
```diff
@@ -0,0 +1,137 @@
...
+
+export async function disconnectLogStreaming(context: IStreamLogsContext): Promise<void> {
+    const logStream: ILogStream | undefined = context.logStreamToStop;
+    if (!logStream) {
+        const allStreams = getActiveLogStreams(context);
+        for (const streams of allStreams.values()) {
+            if (streams && streams.isConnected) {
+                streams.dispose();
+            } else {
+                await vscode.window.showWarningMessage(localize('alreadyDisconnected', 'The log-streaming service is already disconnected.'));
```
nturinski, line 127: Generally you should use the `ext.ui.showWarningMessage` since it has certain behaviors that are consistent amongst extensions.

src/commands/logStream/logStreamRequest.ts
```diff
@@ -0,0 +1,137 @@
...
+import { IStreamLogsContext } from './IStreamLogsContext';
+
+export interface ILogStream extends vscode.Disposable {
+    isConnected: boolean;
+    outputChannel: vscode.OutputChannel;
+    data: {
+        revision?: string;
+        replica?: string;
+        container?: string;
+        containerApp?: string;
```
nturinski, line 25: nit: Could you move this to the top so it makes more hierarchal sense?

src/commands/logStream/logStreamRequest.ts
```diff
@@ -0,0 +1,137 @@
...
+        });
+    }
+}
+
+export async function disconnectLogStreaming(context: IStreamLogsContext): Promise<void> {
+    const logStream: ILogStream | undefined = context.logStreamToStop;
+    if (!logStream) {
+        const allStreams = getActiveLogStreams(context);
+        for (const streams of allStreams.values()) {
+            if (streams && streams.isConnected) {
```
nturinski, line 124: I think the cleaner way to do this is to make this block of code a utility method and then just calling it on logStream or in each stream in `allStreams`.

You could also do something like

`const allStreams = context.logStreamToStop ? [context.logStreamToStop] : getActiveLogStreams(context);`

src/commands/logStream/logStreamRequest.ts
```diff
@@ -0,0 +1,129 @@
...
+                });
+            });
+        });
+    }
+}
+
+export async function disconnectLogStreaming(context: IStreamLogsContext): Promise<void> {
+    const allStreams = context.logStreamToStop ? [context.logStreamToStop] : getActiveLogStreams(context);
+
+    for (const streams of allStreams.values()) {
```
nturinski, line 122: What happens if there are no active log streams? We should probably throw an error. There's another spot `getActiveLogStreams`  is used (and an error is thrown) so maybe it makes sense to just have the error be thrown from that function?

src/commands/logStream/RevisionListStep.ts
```diff
@@ -0,0 +1,40 @@
...
+import { localize } from "../../utils/localize";
+import { IStreamLogsContext } from "./IStreamLogsContext";
+
+export class RevisionListStep extends AzureWizardPromptStep<IStreamLogsContext> {
+    public async prompt(context: IStreamLogsContext): Promise<void> {
+        if (context.containerApp.revisionsMode === 'Multiple') {
+            const placeHolder: string = localize('selectRevision', 'Select a revision');
+            context.revision = (await context.ui.showQuickPick(this.getPicks(context), { placeHolder })).data;
+        } else {
+            (await this.getPicks(context)).forEach(revision => {
```
nturinski, line 20: You can do this to get the latest/active revision. Single revision's latest one should always be the active, AFAIK.

`const revisionData = await client.containerAppsRevisions.getRevision(this.resourceGroup, this.name, nonNullProp(this.containerApp, 'latestRevisionName'));`

src/commands/logStream/RevisionListStep.ts
```diff
@@ -0,0 +1,40 @@
...
+                if (revision.data.active === true) {
+                    context.revision = revision.data;
+                }
+            });
+        }
+    }
+
+    public shouldPrompt(context: IStreamLogsContext): boolean {
```
nturinski, line 28: I think you could use the `configureBeforePrompt` here effectively.

If the revision mode is single, then just set `context.revision` to the latest.
If the revision mode is multiple, but there is only 1 revision, then we can automatically set `context.revision` to that one.

src/commands/logStream/StreamListStep.ts
```diff
@@ -0,0 +1,38 @@
...
+            picks.push({ label: localize('stopAll', 'Stop all Streams'), data: undefined });
+        }
+        context.logStreamToStop = (await context.ui.showQuickPick(picks, { placeHolder, enableGrouping: true, suppressPersistence: true })).data;
+    }
+
+    public shouldPrompt(context: IStreamLogsContext): boolean {
+        return !context.logStreamToStop;
+    }
+
+    private async getPicks(context: IStreamLogsContext): Promise<IAzureQuickPickItem<ILogStream>[]> {
```
nturinski, line 25: I'm pretty sure that this doesn't need to be async.

src/commands/logStream/logStreamRequest.ts
motm32: `StreamListStep` currently throws an error if there are no active streams. This is called before `disconnectLogStreaming ` so it should already be thrown if there are no active streams.

nturinski: That makes sense. 

I'd personally still prefer to have the error be thrown in `getActiveLogStreams` though.  I know it's probably unlikely, but in the event that we ever have to call `disconnectLogStreaming` from somewhere else, it'd be good to have the error behavior consistent.

src/commands/logStream/RevisionListStep.ts
```diff
@@ -13,15 +13,18 @@ import { IStreamLogsContext } from "./IStreamLogsContext";
...
-                if (revision.data.active === true) {
-                    context.revision = revision.data;
-                }
-            });
+        const placeHolder: string = localize('selectRevision', 'Select a revision');
+        context.revision = (await context.ui.showQuickPick(this.getPicks(context), { placeHolder })).data;
+    }
+
+    public async configureBeforePrompt(context: IStreamLogsContext): Promise<void> {
+        const picks = await this.getPicks(context);
```
nturinski, line 21: We don't want to call this first. In the event that this is single revision mode, this call was pointless and makes the wizard slightly more sluggish.

I'd suggestion that you first check if the revisionsMode is single, then in your else clause, do the logic to see how many revisions there are.

nturinski: You may also want to cache the results of `this.getPicks` in the wizard step, otherwise we will call it again needlessly if the event we actually do prompt the users.

Since it's pretty rare to actually have multiple revision mode, I wouldn't mind too much if you don't cache it, but I thought I'd mention it for best practice :)

src/commands/logStream/RevisionListStep.ts
```diff
@@ -12,19 +12,23 @@ import { localize } from "../../utils/localize";
 import { IStreamLogsContext } from "./IStreamLogsContext";
 
 export class RevisionListStep extends AzureWizardPromptStep<IStreamLogsContext> {
+    private revisions: Revision[] | undefined;
+
     public async prompt(context: IStreamLogsContext): Promise<void> {
         const placeHolder: string = localize('selectRevision', 'Select a revision');
-        context.revision = (await context.ui.showQuickPick(this.getPicks(context), { placeHolder })).data;
+        context.revision = (await context.ui.showQuickPick(await this.getPicks(context), { placeHolder })).data;
```
nturinski, line 19: Generally you don't want to await `this.getPicks` in `showQuickPick`. `showQuickPick` is designed to handle promises so that it will show a loading bar while the promise resolves. If you await it, there'll be a small gap where the quickpick UI doesn't show up while it's loading the picks.

motm32: whoops accidently added that ahaha.

src/commands/logStream/StreamListStep.ts
```diff
@@ -0,0 +1,36 @@
...
+import { localize } from "../../utils/localize";
+import { IStreamLogsContext } from "./IStreamLogsContext";
+import { ILogStream, getActiveLogStreams } from "./logStreamRequest";
+
+export class StreamListStep extends AzureWizardPromptStep<IStreamLogsContext> {
+    public async prompt(context: IStreamLogsContext): Promise<void> {
+        const placeHolder: string = localize('selectStream', 'Select a stream');
+        const picks: IAzureQuickPickItem<ILogStream | undefined>[] = this.getPicks(context);
+        if (picks.length > 1) {
+            picks.push({ label: localize('stopAll', 'Stop all Streams'), data: undefined });
```
nturinski, line 16: lol.... nit: I'm not sure if it should just be `Stop all streams` since it's not exactly a title.

nturinski (approved):

PR #352, opened 2023-04-20
Fix error with `Deploy to Container App...`

motm32:
This error started showing up after updating the API version to add log streams. 
image

We basically should allow null values but still want to make sure values are not undefined.

### DIFF
--- a/src/tree/ContainerAppItem.ts
+++ b/src/tree/ContainerAppItem.ts
@@ -5,7 +5,7 @@
 
 import { ContainerApp, ContainerAppsAPIClient, KnownActiveRevisionsMode } from "@azure/arm-appcontainers";
 import { getResourceGroupFromId, uiUtils } from "@microsoft/vscode-azext-azureutils";
-import { AzureWizard, DeleteConfirmationStep, IActionContext, callWithTelemetryAndErrorHandling, createSubscriptionContext, nonNullProp } from "@microsoft/vscode-azext-utils";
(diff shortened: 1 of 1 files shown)


src/tree/ContainerAppItem.ts
```diff
@@ -154,7 +154,12 @@ export async function getContainerEnvelopeWithSecrets(context: IActionContext, s
 
     // verify all top-level properties
     for (const key of Object.keys(containerAppEnvelope)) {
-        containerAppEnvelope[key] = nonNullProp(containerAppEnvelope, <keyof ContainerApp>key);
+        // We know it's undefined but we want it to throw the error
+        if (containerAppEnvelope[key] === undefined) {
```
bwateratmsft, line 158: Are there other falsy values that we should check for? `null`, `''` (empty string), `0`, etc.? Or is `undefined` the only one we expect to deal with here?

bwateratmsft (approved):

src/tree/ContainerAppItem.ts
motm32: Currently the error is occurring because the workloadProfileType property is set to null which is why we changed it to just check for undefined. I don't think we would want to check for '' or 0 in this case either.

nturinski: Yeah, that function is only here to typecheck all of the properties because when dealing with them in other commands, it's annoying to always do =|| or assert that they're non-null.

Issue #313, opened 2023-03-29
Deploy to a container app via connected GitHub repository and enable CI


nturinski:
Similar to how to SWA does it, we want to hook up the user's GitHub repository so that whenever they push a new commit to it, it will automatically update the container app.

MicroFish91:
Related tracking: https://github.com/microsoft/azure-container-apps/issues/530

MicroFish91:
PR roadmap:
- [ ] Scaffold `connectToGitHub` and add wizard steps to obtain the GitHub org, repository, and branch <b>(current)</b>
- [ ] Finish the rest of the `non-Octokit-related` support for the command
- [ ] Add `Actions` tree node & children, similar to SWA
- [ ] Add a disconnect command
- [ ] Add more robust service principal selection/creation
- [ ] Convert GitHub work to shared package

MicroFish91:
Have been blocked for a little while on waiting for the OIDC related SDK updates, see checklist above for more context**

PR #361, opened 2023-05-04
Add GitHub Actions tree node and children

MicroFish91:
Part 3 of https://github.com/microsoft/vscode-azurecontainerapps/issues/313#issuecomment-1518418209


MicroFish91:
Side-note:  I'm not super satisfied with the current tree item context names used, I will probably do some overall renaming in a later PR, but didn't want to get carried away doing that in this one as it might touch a couple un-related tree items as well.

src/utils/actionUtils.ts
```diff
@@ -0,0 +1,133 @@
...
+                colorId = 'testing.iconPassed';
+                break;
+        }
+    }
+
+    return new ThemeIcon(id, colorId ? new ThemeColor(colorId) : undefined);
+}
+
+// Status
+export function ensureStatus(data: { status: string | null }): Status {
```
bwateratmsft, line 90: This isn't too different from `nonNullProp`, can that be used instead?

src/utils/actionUtils.ts
```diff
@@ -0,0 +1,133 @@
...
+    if (Object.values(Status).includes(<Status>data.status)) {
+        return <Status>data.status;
+    } else {
+        throw new RangeError(localize('invalidStatus', 'Invalid status "{0}".', data.status));
+    }
+}
+
+// Helpers...
+function ensureConclusion(data: { conclusion: string | null }): Conclusion {
```
bwateratmsft, line 99: This isn't too different from `nonNullProp`, can that be used instead?

src/tree/gitHub/ActionTreeItem.ts
```diff
@@ -0,0 +1,58 @@
...
+            const { owner, name: repo } = gitUrlParse(this.actionWorkflowRuns.repository.html_url);
+            const getJobsParams: GetJobsParams = {
+                owner,
+                repo,
+                run_id: this.actionWorkflowRuns.id
+            };
+            return await getJobs(context, getJobsParams);
+        });
+
+        if (jobsData && jobsData.total_count > 0) {
```
bwateratmsft, line 52: What you have is valid, but another way to write it is this:

```suggestion
        if (jobsData?.total_count) {
```

You can keep whichever format you like better.

src/utils/actionUtils.ts
MicroFish91: Good catch, I should be able to just use that + typecast

package.json
```diff
@@ -384,6 +384,7 @@
         "dayjs": "^1.11.3",
         "dotenv": "^16.0.0",
         "fs-extra": "^8.1.0",
+        "git-url-parse": "^13.1.0",
```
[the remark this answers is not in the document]
MicroFish91: Yeah sure, I'll look into adding our own parsing utility

src/tree/gitHub/ActionTreeItem.ts
MicroFish91: Definitely prefer your suggestion

PR #376, opened 2023-05-18
Update `toggleIngressVisibility` to utilize `IngressContext` and the activity log

MicroFish91:
Last of the ingress related updates 🙂


bwateratmsft (approved):

src/commands/ingress/toggleIngressVisibility/ToggleIngressVisibilityStep.ts
```diff
@@ -0,0 +1,33 @@
...
+import type { Progress } from "vscode";
+import { IngressConstants } from "../../../constants";
+import { localize } from "../../../utils/localize";
+import type { IngressContext } from "../IngressContext";
+import { IngressUpdateBaseStep } from "../IngressUpdateBaseStep";
+
+export class ToggleIngressVisibilityStep extends IngressUpdateBaseStep<IngressContext> {
+    public priority: number = 300;
+
+    public async execute(context: IngressContext, progress: Progress<{ message?: string | undefined; increment?: number | undefined }>): Promise<void> {
```
nturinski, line 16: In this case, it's kind of nit-picky, but I really don't like having an execute step having a prompt. It kind of goes against our typical wizard patterns.

If there ever is a case where we want to use this with other wizards/commands, it'll be super jarring to have a prompt in the middle of your `wizard.execute()` sequence.

I won't insist on changing it now because of its generally low impact (I doubt it'll be used anywhere else), but please keep that in mind for any other future execute steps.

src/commands/ingress/toggleIngressVisibility/ToggleIngressVisibilityStep.ts
```diff
@@ -0,0 +1,33 @@
...
+        ingress.external = !ingress.external;
+
+        const working: string = localize('updatingVisibility', 'Updating ingress visibility...');
+        const workCompleted: string = localize('updatedVisibility', 'Updated container app "{0}" ingress visibility to "{1}"', containerApp.name, ingress.external ? IngressConstants.external : IngressConstants.internal);
+
+        await this.updateIngressSettings(context, progress, { ingress, working, workCompleted });
+    }
+
+    public shouldExecute(): boolean {
+        return true;
```
nturinski, line 31: I've also noticed other steps where `shouldExecute` just returns true.  I think this is another practice that goes against our typical wizard design patterns. 

Typically, we want to think about the proper conditions for these steps to execute. Again, since this step isn't super reusable,it doesn't matter very much in this case, but the context for properties that the user set should be validated to ensure that there weren't any errors with prompting. I think in general, it will make code less error prone and debugging easier.

A bit of a tangent, but I kind of want to execute steps to log when they don't execute and which values were missing that caused them not to execute for further clarity.

src/commands/ingress/toggleIngressVisibility/toggleIngressVisibility.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    const wizard: AzureWizard<IngressContext> = new AzureWizard(wizardContext, {
+        title,
+        executeSteps,
+        showLoadingPrompt: true
+    });
+
+    // Title normally gets set during prompt phase... since no prompt steps are provided we must set the 'activityTitle' manually
+    wizardContext.activityTitle = title;
+    await wizard.execute();
```
nturinski, line 38: Just as a sidenote, I prefer having the `promptStep` actually be the wizard step, and then just having the "execute" code just be part of the command.

src/commands/ingress/toggleIngressVisibility/ToggleIngressVisibilityStep.ts
MicroFish91: That's fair, it felt weird adding just that section as its own promptStep though.  I'm happy to migrate that section out of the execute step though.  We still need it to be an execute step to utilize the setIngressSettings method.

src/commands/ingress/toggleIngressVisibility/ToggleIngressVisibilityStep.ts
MicroFish91: Noted, though in this case it wasn't super clear to me if we needed to even check for anything here since it's just a toggle.  This did make me reconsider some of the other sections though... it's perhaps relevant to add a more specific check when we edit the target port.

src/commands/ingress/toggleIngressVisibility/toggleIngressVisibility.ts
MicroFish91: See section above, but I created a base execute step that allowed all ingress steps to inherit a method for updating ingress settings.  I think that this case was a bit of an outlier in that it still had a smidge of a prompt inside the execution, but normally it doesn't... and in this case I can just move the warning prompt out to keep it a bit more consistent.

PR #378, opened 2023-05-22
Add more robust `Dockerfile` selection step

MicroFish91:
Addresses next part of [this](https://github.com/microsoft/vscode-azurecontainerapps/issues/313#issuecomment-1518418209).

Added a reusable wizard step that allows a user to traverse a GitHub repository through quick pick steps and search for a designated file (in this case `Dockerfile`).

GitHub_remote_repo


src/commands/gitHub/connectToGitHub/connectToGitHub.ts
```diff
@@ -48,7 +48,11 @@ export async function connectToGitHub(context: ITreeItemPickerContext & Partial<
         new GitHubOrgListStep(),
         new GitHubRepositoryListStep(),
         new GitHubBranchListStep(),
-        new DockerfileLocationInputStep(),
+        new GitHubRepositoryFileSelectStep(
+            'Dockerfile',  // fileName
```
MicroFish91, line 52: The action only works properly when this case-sensitive version of the `Dockerfile` is provided.

src/gitHub/GitHubRepositoryFileSelectStep.ts
```diff
@@ -0,0 +1,113 @@
...
+        this.promptPlaceHolder = prompt;
+    }
+
+    public async prompt(context: IGitHubContext): Promise<void> {
+        // Reset cache when prompting in case the user has made breaking changes via the back button
+        this.cachedPicks = new Map();
+
+        let path: string = '';
+        while (true) {
+            const contentPickData = (await context.ui.showQuickPick(this.getPicks(context, path), { placeHolder: this.promptPlaceHolder })).data;
```
nturinski, line 42: Rather than` suppressPersistence` on every pick (down below) why not just set it as an option for the entire quickpick?

MicroFish91: I actually didn't realize there was a suppressPersistence option there 😂

src/gitHub/GitHubRepositoryFileSelectStep.ts
```diff
@@ -0,0 +1,113 @@
...
+
+            if (contentPickData.traverse === 'Up') {
+                path = path.split('/').slice(0, -1).join('/');
```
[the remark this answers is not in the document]
MicroFish91: Yeah good call, that's a much more efficient solution

MicroFish91:
I got a request to look into implementing a version of this using an FS provider and `openFileDialog`.  Looks like it would be possible, but I haven't gotten around to doing it yet as I'm still prioritizing finishing up the batch revision work.  Perhaps the way forward is to revisit and compare these implementations in the near future?

PR #381, opened 2023-06-06
Add Service Connector

motm32:
Related to https://github.com/microsoft/vscode-azuretools/issues/1419

A couple of notes: 

Currently just for testing the create, delete, and validate service connector commands are accessible from right clicking on a container app. 
image
These commands will be moved to the service connector node once that is added in tools. In order to view if the service connector was created/deleted correctly you will have to look at the portal.


src/commands/serviceConnector/createServiceConnector.ts
```diff
@@ -0,0 +1,43 @@
...
+        new ContainerPickStep(),
+    ];
+
+    const wizard: AzureWizard<IServiceConnectorWithActivityContext> = new AzureWizard(wizardContext, {
+        title: localize('createServiceConnector', 'Create Service Connector'),
+        promptSteps,
+    });
+
+    await wizard.prompt();
+    await createLinker(wizardContext, nonNullValue(item));
```
alexweininger, line 42: The item here only needs to fulfill this interface. While `item` here fulfills this, I prefer to only pass in exactly what is needed where we can. 
```ts
interface LinkerItem {
    subscription: AzureSubscription,
    id: string,
}
```

```suggestion
    await createLinker(wizardContext, { id, subscription });
```

MicroFish91, line 42: I see `createLinker` builds its own wizard.  Is the intent that whenever we need to create a link that we must first add a step to define the scope?  

If so, I'm wondering if it makes more sense to use a different pattern here.  An example solution would be to build these functions such that you can pass them the precursor step(s), and they incorporate it into the wizard for you.  Depending on the answer above, there are probably some other patterns that could work as well... mostly just positing for discussion purposes since I don't have a strong opinion on the solution yet.

(Main reason I brought this was up... something felt weird to me about having two wizards going like this... especially since the first one builds for the activity log, but `ICreateServiceContext` doesn't type/build for an activity context.  I was thinking it made more sense to try and design around building just one continuous wizard process if possible in these instances, whether that be through using a helper to build the wizard like suggested above, or through importing the steps into the wizard used in the client extension... )

src/commands/serviceConnector/ContainerPickStep.ts
```diff
@@ -0,0 +1,26 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { AzureWizardPromptStep, IActionContext, IAzureQuickPickItem, nonNullValue } from "@microsoft/vscode-azext-utils";
+import { localize } from "../../utils/localize";
+import { IServiceConnectorContext } from "./IServiceConnectorContext";
+
+export class ContainerPickStep extends AzureWizardPromptStep<IActionContext> {
```
MicroFish91, line 10:
```suggestion
export class ContainerPickStep extends AzureWizardPromptStep<IServiceConnectorContext> {
```

src/commands/serviceConnector/ContainerPickStep.ts
```diff
@@ -0,0 +1,26 @@
...
+    public async prompt(context: IServiceConnectorContext): Promise<void> {
```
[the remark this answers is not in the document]
bwateratmsft: I prefer keeping `public` to keep things explicit. Time for a cage match!

src/commands/serviceConnector/ContainerPickStep.ts
```diff
@@ -0,0 +1,26 @@
...
+    }
+
+    public shouldPrompt(context: IServiceConnectorContext): boolean {
+        return !context.scope
+    }
+
+    private async getPicks(context: IServiceConnectorContext): Promise<IAzureQuickPickItem<string>[]> {
+        const containers = nonNullValue(context.containerApp.template?.containers);
+        return containers.map(c => {
+            return { label: nonNullValue(c.name), data: nonNullValue(c.name) }
```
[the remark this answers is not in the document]
bwateratmsft: I do too unless it's very short. On a line this length I could go either way.

src/commands/serviceConnector/validateServiceConnector.ts
```diff
@@ -0,0 +1,41 @@
...
+        containerApp: containerApp,
+        ...(await createActivityContext())
+    }
+
+    const promptSteps: AzureWizardPromptStep<IServiceConnectorWithActivityContext>[] = [
+        new ContainerPickStep(),
+    ];
+
+    const wizard: AzureWizard<IServiceConnectorWithActivityContext> = new AzureWizard(wizardContext, {
+        title: localize('deleteServiceConnector', 'Delete Service Connector'),
```
bwateratmsft, line 35: Is this supposed to be "Delete" or is that a copypasta error?

src/commands/serviceConnector/ContainerPickStep.ts
```diff
@@ -0,0 +1,36 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { AzureWizardPromptStep, IAzureQuickPickItem, nonNullValue } from "@microsoft/vscode-azext-utils";
+import { localize } from "../../utils/localize";
+import { IServiceConnectorContext } from "./IServiceConnectorContext";
+
+export class ContainerPickStep extends AzureWizardPromptStep<IServiceConnectorContext> {
```
nturinski, line 10: Generally this is called a `ListStep` not a `PickStep`. 

Though I get that there is already a `ContainerListStep`, I think we should rename things to be more explicit. The `ContainerListStep` today is technically the `ReplicaContainerListStep` and has a hard requirement of having a `replica` defined on the context (which happens in the previous step). So maybe just rename that one and then call this `ContainerListStep` or `TemplateContainerListStep`?

src/commands/serviceConnector/IServiceConnectorContext.ts
```diff
@@ -0,0 +1,14 @@
...
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { ICreateLinkerContext } from "@microsoft/vscode-azext-serviceconnector";
+import { ExecuteActivityContext } from "@microsoft/vscode-azext-utils";
+import { ContainerAppModel } from "../../tree/ContainerAppItem";
+
+export interface IServiceConnectorContext extends ICreateLinkerContext {
+    containerApp: ContainerAppModel;
```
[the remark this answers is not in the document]
MicroFish91: > I extended from the ICreateLinkerContext since I needed the scope.

Does it not make sense to extend from both?

> Yeah, `IStreamLogsContext` and `IAddScaleRuleWizardContext` could as well though `IContainerAppContext` has `containerApp` as an optional property. Not sure if that gunks things up a bit.

Yeah, there are a few contexts that still need to be migrated to extend from `IContainerAppContext` since it was only recently added.  Scale rules is definitely one of those contexts.

src/utils/azureClients.ts
```diff
@@ -30,8 +30,6 @@ export function createContainerRegistryClient(context: AzExtClientContext, regis
     const clientContext = parseClientContext(context);
     // @azure/container-registry doesn't support ADAL tokens at all.  If it sees `signRequest` is defined
     // it errors, but we don't actually need `signRequest` because this is a T2 package
```
alexweininger, line 32: You can remove this comment now 😃

src/commands/serviceConnector/IServiceConnectorContext.ts
```diff
@@ -0,0 +1,10 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { ICreateLinkerContext } from "@microsoft/vscode-azext-serviceconnector";
+import { IContainerAppContext } from "../createContainerApp/IContainerAppContext";
+
+export interface IServiceConnectorContext extends ICreateLinkerContext, IContainerAppContext {
```
MicroFish91, line 9: Nit: A possible cleaner looking alternative (if you don't end up adding any properties):

`export type IServiceConnectorContext = ICreateLinkerContext & IContainerAppContext;`

On a side note, I'm more and more agreeing with Alex on not liking the `I` prefix for context interfaces/types (that's not a change request, more of just a general musing 🙂).

PR #382, opened 2023-06-08
Add `createGraphClient`

MicroFish91:
Will need to utilize Microsoft Graph to add support for CRUD operations on Service Principals, see: https://github.com/microsoft/vscode-azurecontainerapps/issues/313#issuecomment-1518418209

Shoutout to @nturinski for helping me get unstuck with proper scopes and setup for the client.

Here's how the login process looks the first time `createGraphClient` is called:
image

image

### DIFF
--- a/src/utils/graphClient.ts
+++ b/src/utils/graphClient.ts
@@ -1,0 +1,33 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AuthenticationProvider, Client, ClientOptions } from "@microsoft/microsoft-graph-client";
+import "isomorphic-fetch"; // Requires importing this library or a fetch polyfill to use the graph client
+import { AuthenticationGetSessionOptions, AuthenticationSession, authentication } from "vscode";
+
+export async function createGraphClient(): Promise<Client> {
+    const scopes: string[] = ['https://graph.microsoft.com/.default'];
+    const sessionOptions: AuthenticationGetSessionOptions = {
+        clearSessionPreference: true,
+        createIfNone: true
+    };
+    const session: AuthenticationSession | undefined = await authentication.getSession('microsoft', scopes, sessionOptions);
+
+    if (!session) {
+        throw new Error('You are not signed in to a valid Microsoft Graph account. Please sign in and try again.');
+    }
+
+    const authProvider: AuthenticationProvider = {
+        // AuthenticationProvider expects 'getAccessToken' to return a Promise
+        getAccessToken: (): Promise<string> => {
+            return Promise.resolve(session.accessToken);
+        }
+    };
+    const clientOptions: ClientOptions = {
+        authProvider
+    };
+
+    return Client.initWithMiddleware(clientOptions);
+}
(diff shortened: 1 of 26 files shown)


tsconfig.json
```diff
@@ -4,7 +4,8 @@
         "target": "es6",
         "outDir": "out",
         "lib": [
-            "es6"
+            "es6",
+            "dom"
```
MicroFish91, line 8: Graph client uses some DOM type references underneath the hood... this solved the related compile errors I was getting

alexweininger: Oh that's unfortunate. I wish there was a way to only include the exact part of the DOM lib we actually need.

alexweininger (commented):
Just for clarity, users have to login to this separately although they might already be logged into Azure since this needs different scopes.

Is it possible to request _all_ the scopes we need including these ones at the beginning or do we want to avoid that?

MicroFish91:
> Just for clarity, users have to login to this separately although they might already be logged into Azure since this needs different scopes.
> 
> Is it possible to request _all_ the scopes we need including these ones at the beginning or do we want to avoid that?

Hmm, yeah I'm curious what the consensus is for this.  I'd have to look into it, but I was originally planning to just go with the defaults for now and add onto the scope as the need arose.  I've only tested so far with the `/me` endpoint, so I don't know which all scopes I will be needing that the defaults don't already cover.

tsconfig.json
bwateratmsft: I'd at least leave a comment in the tsconfig about it, but this feels bad.

src/utils/graphClient.ts
```diff
@@ -0,0 +1,42 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { AuthenticationProvider, Client, ClientOptions } from "@microsoft/microsoft-graph-client";
```
bwateratmsft, line 6:
```suggestion
import { AuthenticationProvider, Client as GraphClient, ClientOptions as GraphClientOptions } from "@microsoft/microsoft-graph-client";
```

And corresponding changes below

MicroFish91:
> Is it plausible to pull it off with direct HTTP requests rather than adding a new dependency (especially with the DOM thing), or is it complicated enough to justify using it?

Sure, let me investigate making the request without the dependency - I'll get back to you on this one

alexweininger:
> > Just for clarity, users have to login to this separately although they might already be logged into Azure since this needs different scopes.
> > Is it possible to request _all_ the scopes we need including these ones at the beginning or do we want to avoid that?
> 
> When you say beginning, are you referring to initializing this client earlier on during like the initial extension activation? Or is the question more focused on maximizing our scopes and we just create the first instance of the client whenever we need to actually use it?

My assumptions are: a new user who wants to use this feature will have to sign in twice. Once to sign into Azure, and again when we need to create this graph client.

My question is: Is it feasible to somehow still only require a single login?

How I _think_ it might be possible: Request the scopes needed to create the graph client on the first sign in, that way we can create the graph client later on without an additional sign in.

With that said, even if we can do that, I'm not sure it's a great idea to request a lot of additional scopes if we don't necessarily plan to always use those scopes. But I think it's worth investigating.

MicroFish91:
Everything works now with the direct HTTP request and with removal of the Microsoft Graph SDK... however, I'm still investigating how to get expand the scopes to operate with Service Principals.

MicroFish91:
Update: Will probably end up closing this out for a different solution... waiting for some final confirmations before doing so

PR #390, opened 2023-06-14
Reconfigure files and commands to fit the new revision management UI

MicroFish91:
Builds on top of this work: #379 

Closes #380 

Summary of work:

- Reorganized some files
- Moved some of the revision commands off of the `ContainerAppItem` to the `RevisionItem` and `RevisionsItem`
- Renamed/restructured revision context values
- Add view properties to the newer `ConfigurationItem`
- Add new `Revision Management` icon
- Some minor related prep work for the upcoming revision draft PRs

PR #397, opened 2023-06-23
Update tree item context values and `viewItem` patterns

MicroFish91:
I reworked the context values to (hopefully) be more intuitive.  Each item's base context now appears more like the name of the node itself and ends with 'item'.  This alone surprisingly simplifies our RegExp expressions quite a bit and looks a lot more readable (in my opinion).  When going through the `package.json`, it's now pretty easy to quickly parse which commands are on which nodes.  

I've also organized all the commands by tree item groupings in the `package.json`.  They appear in roughly the same order as in the tree view for quick parsing.  Hopefully you all find this a little bit easier to go through.  I personally find it helps me locate the commands more quickly.

Note:  This PR skips the GitHub-related tree items and commands because I'm already planning to revisit those again when I migrate to the upcoming GitHub shared tools package.


src/tree/ContainerAppItem.ts
```diff
@@ -82,7 +82,7 @@ export class ContainerAppItem implements ContainerAppsItem {
             id: this.id,
             label: nonNullProp(this.containerApp, 'name'),
             iconPath: treeUtils.getIconPath('azure-containerapps'),
-            contextValue: `containerApp|revisionmode:${this.containerApp.revisionsMode}`,
```
MicroFish91, line 85: An old artifact that could have been removed in this [PR](https://github.com/microsoft/vscode-azurecontainerapps/pull/390)

src/tree/configurations/IngressItem.ts
```diff
@@ -7,17 +7,20 @@ import { ContainerApp, Ingress } from "@azure/arm-appcontainers";
...
 import { ContainerAppModel } from "../ContainerAppItem";
 import { ContainerAppsItem, TreeElementBase } from "../ContainerAppsBranchDataProvider";
 
 const label: string = localize('ingress', 'Ingress');
 
-export class IngressItem implements ContainerAppsItem {
-    static contextValue: string = 'ingress|enabled';
-    readonly contextValue: string = `${IngressItem.contextValue}|${azResourceContextValue}`;
+export const targetPortItemContextValue: string = 'targetPortItem';
+export const visibilityItemContextValue: string = 'visibilityItem';
```
MicroFish91, line 19: I've been exporting the `item`-based context values to convey reusability even though some are only used in the one containing file.  I'm totally cool with removing the exports if people prefer that.

src/tree/revisionManagement/RevisionItem.ts
```diff
@@ -17,9 +17,14 @@ export interface RevisionsItemModel extends ContainerAppsItem {
     revision: Revision;
 }
 
+const revisionStateActiveContextValue: string = 'revisionState:active';
+const revisionStateInactiveContextValue: string = 'revisionState:inactive';
+const revisionModeSingleContextValue: string = 'revisionMode:single';
+const revisionModeMultipleContextValue: string = 'revisionMode:multiple';
```
MicroFish91, line 23: Put at the top here like this for readability/discoverability, but totally happy to re-write this inline again if preferred

bwateratmsft: image

src/tree/configurations/IngressItem.ts
bwateratmsft: I'd export only what is used outside the file

bwateratmsft (approved):

src/tree/LogsItem.ts
```diff
@@ -11,14 +11,21 @@ import { createPortalUrl } from "../utils/createPortalUrl";
 import { localize } from "../utils/localize";
 import { TreeElementBase } from "./ContainerAppsBranchDataProvider";
 
-export class LogsItem implements TreeElementBase {
```
bwateratmsft, line 14: Since this class is renamed, consider also renaming the file to match

src/tree/configurations/IngressItem.ts
```diff
@@ -7,17 +7,20 @@ import { ContainerApp, Ingress } from "@azure/arm-appcontainers";
...
 
 const label: string = localize('ingress', 'Ingress');
 
-export class IngressItem implements ContainerAppsItem {
-    static contextValue: string = 'ingress|enabled';
-    readonly contextValue: string = `${IngressItem.contextValue}|${azResourceContextValue}`;
+const targetPortItemContextValue: string = 'targetPortItem';
+const visibilityItemContextValue: string = 'visibilityItem';
+
+export class IngressEnabledItem implements ContainerAppsItem {
```
bwateratmsft, line 21: Since this class is renamed, consider also renaming the file to match

bwateratmsft (approved):

src/tree/LogsItem.ts
MicroFish91: Good catch, totally missed that

src/tree/configurations/IngressItem.ts
MicroFish91: I'm considering keeping the file name generic as just `IngressItem` here because it houses both the enabled and disabled versions of the tree item.  Just wanted it to match the same naming convention we were using for the Dapr tree items.

bwateratmsft: Fair enough

PR #400, opened 2023-06-27
Remove `Latest` in single revision mode

MicroFish91:
Based on the new changes we discussed, remove the `Latest` node and have `Scaling` sit up one level.  Re-arrange commands to match.

image


src/tree/ContainerAppItem.ts
```diff
@@ -57,16 +58,22 @@ export class ContainerAppItem implements ContainerAppsItem {
 
     portalUrl: Uri = createPortalUrl(this.subscription, this.containerApp.id);
 
+    get contextValue(): string {
```
alexweininger, line 61: Can this be private?

alexweininger (approved):

PR #403, opened 2023-07-06
Reorganizing commands in preparation for upcoming revision draft work

MicroFish91:
In preparation for the upcoming revision draft changes, I wanted to reorganize some of the commands ahead of time.

Summary of work:

- Remove workspace actions for container apps since we'll only have one create command to display anyway
- Get rid of the generic deploy command as we will be replacing with a revision draft-based version instead
- Add back `deployImage` and rename command tooltip to ~Update Container App Image..., plan to move this command to images node once we add that tree item to the UI


package.nls.json
```diff
@@ -6,7 +6,7 @@
     "containerApps.enableOutputTimestamps": "Prepends each line displayed in the output channel with a timestamp.",
     "containerApps.browse": "Browse",
     "containerApps.createContainerApp": "Create Container App...",
-    "containerApps.deploy": "Deploy to Container App...",
+    "containerApps.deployImage": "Update Image in Container App...",
     "containerApps.deployImageApi": "Update Image in Container App (API)...",
```
alexweininger, line 10: I'm fine leaving this as-is for now, but I'd like to discuss changing the wording for this command in a future PR.

I think "Update Container App Image..." is a better title for this command. I understand that there can be multiple images in a Container App, but if a user isn't aware of that "Update Image in Container App..." sounds very odd. I think "Update Container App Image..." makes sense in both scenarios, whether the CA is running a single image or multiple images.

Also, whatever we decide, we should make sure the command id is changed to match the command title. If possible, I really prefer for the command ids to match the command title. This makes development a bit less confusing for those who aren't working on the extension as frequently.

alexweininger (approved):

package.nls.json
MicroFish91: I prefer that title as well, I even accidentally wrote it as that in my note above 😂
I think that's probably something I can change here before we merge if no objections.

As for changing the `deployImage` and `deployImageApi` command ids - I think we can look into changing those when we revisit adding the images node and re-associating those commands with the batch revision work.  We would need to have the Docker extension change the command name it calls as well since it currently leverages `deployImageApi`.  I'll talk with Brandon about changing the name when we make that update

nturinski: > I'm fine leaving this as-is for now, but I'd like to discuss changing the wording for this command in a future PR.
> 
> I think "Update Container App Image..." is a better title for this command. I understand that there can be multiple images in a Container App, but if a user isn't aware of that "Update Image in Container App..." sounds very odd. I think "Update Container App Image..." makes sense in both scenarios, whether the CA is running a single image or multiple images.
> 
> Also, whatever we decide, we should make sure the command id is changed to match the command title. If possible, I really prefer for the command ids to match the command title. This makes development a bit less confusing for those who aren't working on the extension as frequently.

My idea would be under the container app node, we'd have an image node.

- Right-clicking the images node gives you the option to "Add an image"
- Expanding the image node reveals the individual images in the container app.
- Right-clicking the image gives you the option to "Update image" or "Delete image"

All three of these commands would be hooked into the "batch revision" editing.

That we we just avoid using any vernacular regarding "container apps in any of these commands since the container app doesn't really have that much to do with the images anyway.

Issue #311, opened 2023-03-29
Enable revision updates through editing container app envelope JSON


nturinski:
## What it is
Today, users can only update one property at a time which causes a new revision to be made. We want to enable a batch update. The most effective way to do this right now is to open the current container app envelope in the VS Code editor, allow them to make edits to whatever properties that they want, and then PUT that to their container app, creating a new revision.

### Features
- Intellisense for editing properties/values in the JSON envelope
- Empowers users to make advanced configurations that we do not natively expose through the UI
- Allows for batch revisions
- Allows for previewing local changes before creating a new revision

### UI Flow
- Right-click container app/revision
- Select "Create new revision..."
- Opens container app envelope in editor
- User makes edits, and if saves or clicks deploy button in editor
- Prompts user to create new revision
- New revision is created

### Implementation
- Intellisense - we can use the ARM package type defintions
  - Investigate how to do this: https://github.com/microsoft/vscode-azurefunctions/blob/main/package.json#L714-L727
- Figure out the entry points for uploading
  - On save?
    - Do we have a never show again or always upload setting?
  - UI node with deploy button
    - Be under the revision nodes (when in multiple)
    - Be under the container app node (when in single)
    - Possibly put it as part of the Local Workspace View?

### Additional Feature Work
- Include draft node in revision UI
- Would include deploy button in the draft node
  - NOTE: Simple UI first, don't show all of these nodes
![image](https://user-images.githubusercontent.com/5290572/228690013-6abb347d-10a6-4086-b40e-a3ad03065bac.png)

PR #405, opened 2023-07-07
Add revision draft support (single revision mode)

MicroFish91:
**Background**:

Based on team feedback surrounding the initial revision draft PR (v1), here is the new and improved v2!  Major changes include:

- Removal of the "Latest" node in single revisions mode.  We now try to hide the concept of revisions a little more in single revisions mode.  The children of the "Latest" node have now come up one level (right now that's really mainly the `scaleItem` ("Scaling")
- Instead of creating a draft node in single revisions mode, we now show "unsaved changes" indicators
- The edit container app json no longer opens up by default and is less easily found, now labeled as an advanced editing feature
- No longer persist revision drafts in Memento storage (in-memory only)

**Purpose**:
- See #311 

**What is covered in this PR**:

- Revision draft support for single revision mode

**What is coming down the line**:

- Deploy revision draft command
- Revision draft support for multiple revisions mode

**Demo of the full feature**:
- [Demo](https://microsoft-my.sharepoint.com/:v:/p/matfisher/IQBFu96qLyiJRbgNBIHUEt8rAZwOtVW_H595XC9F_IUqwYg?e=VOK9c5) (Not a super polished demo, but good enough to give some context)

I had to include a little bit of the multiple revisions mode code (for example, `RevisionDraftItem`) because of some loose coupling, but overall I tried to split it up so it's a little bit less code to review overall.

If you see missing command spots in the `package.json`, you can see here for what the full thing will eventually look like: https://github.com/microsoft/vscode-azurecontainerapps/pull/405/commits/a58ee13a3184cc7a4f18425e6e8350965a757e19


src/tree/scaling/ScaleItem.ts
```diff
@@ -43,22 +47,48 @@ export class ScaleItem implements ContainerAppsItem {
...
+    private hasUnsavedChanges(): boolean {
+        const scaleDraftTemplate = ext.revisionDraftFileSystem.parseRevisionDraft(this)?.scale;
+        if (!scaleDraftTemplate) {
+            return false;
+        }
+
+        if (this.containerApp.revisionsMode === KnownActiveRevisionsMode.Single) {
+            return !!this.containerApp.template?.scale && !isDeepEqual(this.containerApp.template.scale, scaleDraftTemplate);
+        } else {
+            // We only care about showing changes to descendants of the revision draft item when in multiple revisions mode
```
MicroFish91, line 88: This code segment will be introduced in the multiple revisions PR, commented out for now with a return placeholder

src/commands/revisionDraft/RevisionDraftFileSystem.ts
```diff
@@ -0,0 +1,197 @@
+import type { Template } from "@azure/arm-appcontainers";
```
nturinski, line 1: Missing license header

src/commands/revisionDraft/RevisionDraftFileSystem.ts
```diff
@@ -0,0 +1,197 @@
...
+ * Enforces a policy of one revision draft per container app
+ */
+export class RevisionDraftFileSystem implements FileSystemProvider {
+    static readonly scheme: string = 'containerAppsRevisionDraft';
+
+    private readonly emitter: EventEmitter<FileChangeEvent[]> = new EventEmitter<FileChangeEvent[]>();
+    private readonly bufferedEvents: FileChangeEvent[] = [];
+    private fireSoonHandle?: NodeJS.Timer;
+
+    private draftStore: Record<string, RevisionDraftFile> = {};
```
nturinski, line 42: It's not necessary, but I think you could simplify some stuff by making this a `Map<string, RevisionDraftFile>` type since it has built in methods for `has` and `delete`,

src/constants.ts
```diff
@@ -71,3 +71,12 @@ export type QuickPicksCache = { cache: QuickPickItem[], next: string | null };
 
 // Originally from the Docker extension: https://github.com/microsoft/vscode-docker/blob/main/src/constants.ts
 export const DOCKERFILE_GLOB_PATTERN = '**/{*.[dD][oO][cC][kK][eE][rR][fF][iI][lL][eE],[dD][oO][cC][kK][eE][rR][fF][iI][lL][eE],[dD][oO][cC][kK][eE][rR][fF][iI][lL][eE].*}';
+
+export const unsavedChangesTrueContextValue: string = 'unsavedChanges:true';
+export const unsavedChangesFalseContextValue: string = 'unsavedChanges:false';
+
+export const revisionDraftTrueContextValue: string = 'revisionDraft:true';
```
nturinski, line 78: This isn't imported by anything. Happening in a follow-up PR?

src/constants.ts
```diff
@@ -71,3 +71,12 @@ export type QuickPicksCache = { cache: QuickPickItem[], next: string | null };
 
 // Originally from the Docker extension: https://github.com/microsoft/vscode-docker/blob/main/src/constants.ts
 export const DOCKERFILE_GLOB_PATTERN = '**/{*.[dD][oO][cC][kK][eE][rR][fF][iI][lL][eE],[dD][oO][cC][kK][eE][rR][fF][iI][lL][eE],[dD][oO][cC][kK][eE][rR][fF][iI][lL][eE].*}';
+
+export const unsavedChangesTrueContextValue: string = 'unsavedChanges:true';
```
nturinski, line 75: Seems like it's only used in `ContainerAppItem` right now. Is that going to change or was there a different reason for having it in this file?

src/tree/revisionManagement/RevisionDraftItem.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    get revisionName(): string {
+        return nonNullProp(this.revision, 'name');
+    }
+
+    /**
+     * @example gets "rev-1" of "my-app--rev-1"
+     */
+    private get baseRevisionName(): string {
+        return this.revisionName.substring(this.revisionName.lastIndexOf('--') + 2);
```
nturinski, line 38: Another way to do this is `this.revisionName.split('--').pop()`  which I think it's a little bit cleaner.

Mostly sharing just as an FYI, not as a requested change. Both work just fine.

src/tree/revisionManagement/RevisionDraftItem.ts
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    static hasDescendant(item: RevisionsItemModel): boolean {
```
nturinski, line 41: What's this for? I see the code that uses it is currently commented out, but I don't totally understand what that's doing either.

src/tree/revisionManagement/RevisionDraftItem.ts
```diff
@@ -0,0 +1,64 @@
...
+            iconPath: treeUtils.getIconPath('revision-draft'),
+            description: this.containerApp.latestRevisionName === this.revisionName ?
+                localize('basedOnLatestRevision', 'Based on "{0}" (Latest)', this.baseRevisionName) :
+                localize('basedOnRevision', 'Based on "{0}"', this.baseRevisionName),
+            contextValue: RevisionDraftItem.contextValue,
+            collapsibleState: TreeItemCollapsibleState.Expanded
+        };
+    }
+
+    getChildren(): TreeElementBase[] {
```
nturinski, line 59: This being the same as the `ContainerAppItem` kind of makes me want this class to be extended from or based off of a an abstract `ContainerAppItemBase`. I think it'll be kind of annoying when we change/add children or the container app item itself.

src/tree/scaling/ScaleItem.ts
```diff
@@ -43,22 +47,48 @@ export class ScaleItem implements ContainerAppsItem {
...
             iconPath: treeUtils.getIconPath('scaling'),
             collapsibleState: TreeItemCollapsibleState.Collapsed,
         }
     }
 
-    async getChildren?(): Promise<TreeElementBase[]> {
+    getChildren(): TreeElementBase[] {
+        let scale: Scale | undefined;
+
+        if (this.hasUnsavedChanges()) {
```
nturinski, line 60: Don't need it now, but maybe just a side-note. Anytime we have an item that is part of the "batch changes", we could have a base model for it that has the same implementation for how to get the value/children and check for unsavedChanges, etc.

src/commands/editContainerApp.ts
```diff
@@ -0,0 +1,21 @@
...
+import { ext } from "../extensionVariables";
+import type { ContainerAppItem } from "../tree/ContainerAppItem";
+import { localize } from "../utils/localize";
+import { pickContainerApp } from "../utils/pickContainerApp";
+
+export async function editContainerApp(context: IActionContext, node?: ContainerAppItem): Promise<void> {
+    node ??= await pickContainerApp(context);
+
+    if (node.containerApp.revisionsMode !== KnownActiveRevisionsMode.Single) {
+        throw new Error(localize('revisionModeError', 'The issued command can only be executed in single revision mode.'));
```
alexweininger, line 17: Is it possible for a user to encounter this error? Can we setup the context values so that this is unreachable if the container app is in multi revision mode?

src/tree/ContainerAppItem.ts
```diff
@@ -60,9 +61,22 @@ export class ContainerAppItem implements ContainerAppsItem {
     private get contextValue(): string {
         const values: string[] = [ContainerAppItem.contextValue];
         values.push(this.containerApp.revisionsMode === KnownActiveRevisionsMode.Single ? revisionModeSingleContextValue : revisionModeMultipleContextValue);
+        values.push(this.hasUnsavedChanges() ? unsavedChangesTrueContextValue : unsavedChangesFalseContextValue);
         return createContextValue(values);
     }
 
+    get description(): string | undefined {
```
alexweininger, line 68: Can this be private?

src/tree/revisionManagement/RevisionDraftItem.ts
MicroFish91: `RevisionDraftItem.hasDescendant()` checks to see if the passed item is a child of the `RevisionDraftItem`.  Since the item has to have a `RevisionsItemModel` and must match the revision base draft, we can conclude it's a child of the draft item (actually I just realized there's an edge case where this doesn't make sense - if you pass `RevisionDraftItem` to itself... but that's just kind of weird to do anyway...).

One example where this method is useful - technically every revision in multiple revisions mode has a `ScaleItem` child.  We don't want all of them to show unsaved changes when we detect a diff, we only care about showing those changes when the child is a descendant of the draft item.  This method will be useful for future children as well.  If you prefer we could always make this a utility function rather than a static method.  I just kind of liked reading: `RevisionDraftItem.hasDescendant(this)` because it felt like a cool way to describe exactly what is being accomplished.

src/tree/ContainerAppItem.ts
MicroFish91: Yessir

src/constants.ts
MicroFish91: You are correct, I probably could comment this out or take it out of this PR.

MicroFish91: It definitely feels weird having revisionDraft and unsavedChanges, but there is an extremely small nuanced difference that ends up being useful for showing context commands

src/constants.ts
MicroFish91: I think for my purposes so far, it will only be used in `ContainerAppItem` so I'm happy to move it back.  At the time I was thinking way down the line and it may end up being used by other items, but that's probably in a future future PR

src/commands/editContainerApp.ts
MicroFish91: I'd have to take this command out of the command palette, then I think it would be unreachable 🤔

src/tree/revisionManagement/RevisionDraftItem.ts
MicroFish91: Actually, I just realized why this might be confusing.  Even though reading off the active revision draft file works, I think it makes more intuitive sense to do:

`return item.revision.name === this.revision.name;` or similar.

src/utils/isDeepEqual.ts
```diff
@@ -0,0 +1,30 @@
...
+
+        if ((areObjects && !isDeepEqual(<object>value1, <object>value2)) || !areObjects && value1 !== value2) {
+            return false;
+        }
+    }
+
+    return true;
+}
+
+function isObject(value: unknown): boolean {
```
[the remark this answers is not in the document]
MicroFish91: Oh cool!  Learned something new again today 🥳

src/tree/scaling/ScaleItem.ts
MicroFish91: Yeah, I totally get what you mean, I was thinking about doing something like this eventually as well.  I'm not sure if I prefer extending from a base here or just implementing an interface (leaning towards the latter).  If we ever decide we want to implement `discard changes` on the item level, that would also be a good candidate to include as well.

src/tree/revisionManagement/RevisionDraftItem.ts
MicroFish91: Actually, instead of enforcing inheritance for this, how do you feel about providing the children through a static method on the `RevisionItem`.  That feels simpler to me and we can still manage the children in one place.  It also makes sense to keep it on that item because the template changes are analogous to the concept of revisions.

src/commands/editContainerApp.ts
alexweininger: Couldn't you just only show Container Apps in single revision mode in the picker?

src/tree/revisionManagement/RevisionDraftItem.ts
MicroFish91: It would probably be more intuitive to read if it said `return item.revision.name === this.revision.name;`.  Unfortunately there's not context of `this` in the static method, so we have to fetch the revision draft file to get this info.

src/commands/editContainerApp.ts
MicroFish91: Hmm, how do you feel about adding a flag on `PickItemOptions` that you pass in when calling `pickContainerApp`?  (Reason being, I think there are still cases in multiple revisions mode where you might still want to be able to pick container app items, so we would probably want to differentiate that condition).

MicroFish91:
Added changes, ready for review again!

src/commands/editContainerApp.ts
nturinski: We know that most users don't use multiple revision mode. I don't think it's worth having this [huge block of code.](https://github.com/microsoft/vscode-azurecontainerapps/pull/405/files#diff-b845014478b2c4c69e8df5ed166a3c3b9b57a2c6f5ba5adc2b8defc7ec184570R44-R60) For the few people that do try to click on a multiple revision app, I think getting an error immediately isn't the worst experience.

Maybe it  would be nice if our picker actually threw an error "step" so that the wizard would allow you to go back.

package.json
```diff
@@ -76,11 +86,6 @@
                 "title": "%containerApps.deployImageApi%",
                 "category": "Azure Container Apps"
             },
-            {
```
nturinski, line 79: Why did you move this? I thought you might have moved it next to `createManagedEnvironment`  but it seems somewhat random?

package.nls.json
```diff
@@ -6,14 +6,16 @@
     "containerApps.enableOutputTimestamps": "Prepends each line displayed in the output channel with a timestamp.",
     "containerApps.browse": "Browse",
     "containerApps.createContainerApp": "Create Container App...",
-    "containerApps.deployImage": "Update Container App Image......",
+    "containerApps.editContainerApp": "Edit Container App (Advanced)",
```
nturinski, line 9:
```suggestion
    "containerApps.editContainerApp": "Edit Container App (Advanced)...",
```

As there are more steps for editing after clicking.

package.nls.json
```diff
@@ -6,14 +6,16 @@
...
+    "containerApps.deployImage": "Update Container App Image...",
```
nturinski, line 10: I'm wondering if we should call this: "Update Container..." or "Edit Container..." after our conversation offline. The image is just a part of the overall container, after all.

It's annoying and confusing, but Container =/= Container App.

src/tree/revisionManagement/RevisionDraftItem.ts
nturinski: Would it make more sense for the `ScaleItem` to have a `isDescendant` function then? And that could probably be some sort of shared function on an abstract base class or an interface or something. Then this could just call `ScaleItem.isDescendant(this)`.

I'm not entirely sure, but I can imagine that the behavior of what is a descendant might change as we add more nodes (like the containers node) depending on how deep it can drill down.

Also not entirely sure when this is being used, so if that wouldn't work, disregard.

package.nls.json
MicroFish91: Nope, not unless it's command palette initiated and you need to pick the container app

package.json
MicroFish91: I put the `createManagedEnvironment` command out from in between the container app commands and up into the managed environment commands section.  It's just misleading because `createContainerApp` is actually a managed environment command even though it has the word container app in it

src/tree/revisionManagement/RevisionDraftItem.ts
MicroFish91: That could work, but my fear is that enforcing inheritance here will make it more painful to maintain down the road.  I personally think the best options are a) either a static method on `RevisionDraftItem` or b) a separate util method if it seems unintuitive to couple the method to the draft item

MicroFish91: > I'm not entirely sure, but I can imagine that the behavior of what is a descendant might change as we add more nodes (like the containers node) depending on how deep it can drill down.

How do you feel about me refactoring to this method if we find that we're having to special case the node behaviors later down the line.  I think it's really unlikely that I'll end up having to do this, but I can't say with 100% certainty.  I I don't think refactoring to this method wouldn't be very expensive to do later if the need arises

package.nls.json
nturinski: Doesn't it open the json file and then you have to make edits for it to actually do anything though?

package.nls.json
MicroFish91: 😵‍💫

package.nls.json
MicroFish91: Oh does that count as two? In that case yes.  There's no wizard steps or anything though

nturinski: Yeah, I would say that because opening the file isn't actually editing it, and in order to edit it, you need to do more steps, that this is a `...` command.

Convoluted? Maybe. But so is this command 🤣

MicroFish91: Well, it is `advanced`™️

src/tree/revisionManagement/RevisionDraftItem.ts
nturinski: Yeah sure. We can address the problem (if it exists) when we actually have more children/complicated scenarios.

nturinski (approved):

PR #406, opened 2023-07-14
Add secrets UI to the tree view

MicroFish91:
Added:
- Secrets UI
- ~~Ability to toggle show and hide on a secret~~

image

More support for adding, delete, and editing in follow-up PRs


bwateratmsft (commented):
@danegsta do you have any concerns about this from a security perspective?

MicroFish91:
> @danegsta do you have any concerns about this from a security perspective?

Good question and thanks for bringing this up for a general discussion.  I'm always a little bit concerned about showing secrets, but I don't have any obvious vulnerabilities that I'm considering, so I'm not sure how to quantify the actual risk here and would love to hear other people's opinions on a best path forward.

From what I've seen so far, the portal does something similar, though it's in a browser environment so has a different set of risks than VS Code.

image

We do something similar for app settings as well across the extensions and this follows a similar format, though maybe not as high profile as `secrets`.  

I'd be open to just removing the ability to view the secrets directly though if we want to mitigate any potential risk altogether.  By the same token, if we go this route I would probably avoid allowing the user the ability to copy the secret value directly.   Otherwise, I think we'd still be able to perform the rest of the operations normally.  Would love to hear other opinions here before we make any further decisions.

MicroFish91:
Update:  Removed the ability to `show` the secret values

src/tree/configurations/secrets/SecretItem.ts
```diff
@@ -0,0 +1,28 @@
...
+import { ThemeIcon, TreeItem } from "vscode";
+import type { ContainerAppModel } from "../../ContainerAppItem";
+import type { ContainerAppsItem } from "../../ContainerAppsBranchDataProvider";
+import { SecretsItem } from "./SecretsItem";
+
+export class SecretItem implements ContainerAppsItem {
+    static readonly contextValue: string = 'secretItem';
+    static readonly contextValueRegExp: RegExp = new RegExp(SecretItem.contextValue);
+
+    constructor(readonly subscription: AzureSubscription, readonly containerApp: ContainerAppModel, readonly secret: Secret) { }
```
bwateratmsft, line 17: Can we get just the secret name instead of the `Secret` object?

MicroFish91: Yeah sure! (just for the record though, the default secret comes back with no `value` populated unless you call `listSecrets` directly)

nturinski (approved):

PR #410, opened 2023-07-18
Support for adding secrets

MicroFish91:
Demo:

### DIFF
--- a/src/commands/secret/ISecretContext.ts
+++ b/src/commands/secret/ISecretContext.ts
@@ -8,10 +8,8 @@ import type { ExecuteActivityContext } from "@microsoft/vscode-azext-utils";
 import type { IContainerAppContext } from "../IContainerAppContext";
 
 export interface ISecretContext extends IContainerAppContext, ExecuteActivityContext {
-    // New
-    secretName?: string;
-    secretValue?: string;
+    newSecretName?: string;
+    newSecretValue?: string;
 
-    // Existing
     secret?: Secret;
 }
--- a/src/commands/secret/ISecretContext.ts
+++ b/src/commands/secret/ISecretContext.ts
@@ -1,0 +1,17 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import type { Secret } from "@azure/arm-appcontainers";
+import type { ExecuteActivityContext } from "@microsoft/vscode-azext-utils";
+import type { IContainerAppContext } from "../IContainerAppContext";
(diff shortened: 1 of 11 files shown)


src/commands/secret/addSecret/SecretValueStep.ts
```diff
@@ -0,0 +1,34 @@
...
+        });
+        context.valuesToMask.push(context.secretValue);
+    }
+
+    public shouldPrompt(context: ISecretContext): boolean {
+        return !context.secretValue;
+    }
+
+    private validateInput(val: string | undefined): string | undefined {
+        val = val ? val.trim() : '';
```
bwateratmsft, line 26: Should the secret be trimmed, even for validation? A lot of services don't work with a secret that begins or ends with whitespace but IMO that's a bug.

src/utils/validateUtils.ts
```diff
@@ -40,4 +41,30 @@ export namespace validateUtils {
...
+    /**
+     * Validates that the given input string consists of lower case alphanumeric characters,
+     * starts and ends with an alphanumeric character, and does not contain any special symbols not explicitly specified
+     *
+     * @param value The original input string to validate
+     * @param symbols Any custom symbols that are also allowed in the input string. Defaults to '-'.
+     *
+     * @example
+     * "abcd-1234" // returns true
+     * "-abcd-1234" // returns false
```
bwateratmsft, line 54: Maybe some unit tests are in order for this?

src/utils/validateUtils.ts
```diff
@@ -40,4 +41,30 @@ export namespace validateUtils {
...
+     */
+    export function isLowerCaseAlphanumericWithSymbols(value: string, symbols: string = '-'): boolean {
```
MicroFish91, line 56: Was thinking we could expand on this with additional options in the future as well.  For example: (1) option to allow uppercase values (would have to rename the function), (2) option to restrict repeating symbols, etc.

Could also add these to shared utils if others find this useful in some capacity.  Open to any thoughts/recommendations :)

src/commands/secret/ISecretContext.ts
```diff
@@ -0,0 +1,17 @@
...
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import type { Secret } from "@azure/arm-appcontainers";
+import type { ExecuteActivityContext } from "@microsoft/vscode-azext-utils";
+import type { IContainerAppContext } from "../IContainerAppContext";
+
+export interface ISecretContext extends IContainerAppContext, ExecuteActivityContext {
+    // New
+    secretName?: string;
```
nturinski, line 12: I think it would be easier to just name the variable names `newSecretName` and `newSecretValue`.

nturinski (approved):

PR #414, opened 2023-07-27
Add `deployRevisionDraft` support for all revision modes

### LINKED ISSUE #311: Enable revision updates through editing container app envelope JSON

MicroFish91:
Closes #311 

[Demo](https://microsoft-my.sharepoint.com/:v:/p/matfisher/IQBFu96qLyiJRbgNBIHUEt8rAZwOtVW_H595XC9F_IUqwYg?e=VOK9c5) (Same as the one shown in previous revision draft PRs)

### DIFF
--- a/src/commands/revisionDraft/deployRevisionDraft/DeployRevisionDraftConfirmStep.ts
+++ b/src/commands/revisionDraft/deployRevisionDraft/DeployRevisionDraftConfirmStep.ts
@@ -3,23 +3,14 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { ContainerAppsAPIClient, KnownActiveRevisionsMode, Revision } from "@azure/arm-appcontainers";
-import { uiUtils } from "@microsoft/vscode-azext-azureutils";
-import { AzureWizardPromptStep, nonNullProp } from "@microsoft/vscode-azext-utils";
-import * as deepEqual from "deep-eql";
-import type { ContainerAppModel } from "../../../tree/ContainerAppItem";
-import { createContainerAppsAPIClient } from "../../../utils/azureClients";
+import { AzureWizardPromptStep } from "@microsoft/vscode-azext-utils";
 import { localize } from "../../../utils/localize";
 import type { IDeployRevisionDraftContext } from "./IDeployRevisionDraftContext";
 
 export class DeployRevisionDraftConfirmStep extends AzureWizardPromptStep<IDeployRevisionDraftContext> {
     public async prompt(context: IDeployRevisionDraftContext): Promise<void> {
-        const deployRevisionWarning: string = await this.hasUnsavedChanges(context) ?
-            localize('deployRevisionWarning', 'This will deploy a new revision to container app "{0}".', context.containerApp?.name) :
-            localize('noUnsavedChangesWarning', 'No unsaved changes detected.\n\nThis will deploy a new revision to container app "{0}".', context.containerApp?.name);
-
         await context.ui.showWarningMessage(
-            deployRevisionWarning,
+            localize('deployRevisionWarning', 'This will deploy a new revision to container app "{0}".', context.containerApp?.name),
             { modal: true },
             { title: localize('continue', 'Continue') }
         );
@@ -27,19 +18,5 @@ export class DeployRevisionDraftConfirmStep extends AzureWizardPromptStep<IDeplo
 
     public shouldPrompt(context: IDeployRevisionDraftContext): boolean {
         return !!context.template;
-    }
-
-    private async hasUnsavedChanges(context: IDeployRevisionDraftContext): Promise<boolean> {
-        const containerApp: ContainerAppModel = nonNullProp(context, 'containerApp');
-
-        if (context.containerApp?.revisionsMode === KnownActiveRevisionsMode.Single) {
-            return !!containerApp.template && !deepEqual(containerApp.template, context.template);
-        } else {
-            const client: ContainerAppsAPIClient = await createContainerAppsAPIClient(context);
-            const revisions: Revision[] = await uiUtils.listAllIterator(client.containerAppsRevisions.listRevisions(containerApp.resourceGroup, containerApp.name));
-            const baseRevision: Revision | undefined = revisions.find(revision => revision.name === context.baseRevisionName);
-
-            return !!baseRevision?.template && !deepEqual(baseRevision.template, context.template);
-        }
     }
 }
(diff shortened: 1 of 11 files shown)


bwateratmsft (approved):

MicroFish91:
Made a few additional changes since the original (demo).  What was added:
- Prevent deploying a new revision when no unsaved changes are detected by throwing an error (mostly a command palette-based concern).
- Don't show the `deploy` icon on the `RevisionDraftItem` when unsaved changes are not detected in multiple revisions mode
- Introduced a `RevisionDraftModel` interface for certain items to implement which currently only requires implementing a method called `hasUnsavedChanges`... but I anticipate in the future we may add additional draft methods like `discardChanges` (to handle dropping partial draft changes rather than forcing full draft discards)

src/commands/revisionDraft/deployRevisionDraft/DeployRevisionDraftConfirmStep.ts
```diff
@@ -0,0 +1,22 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { AzureWizardPromptStep } from "@microsoft/vscode-azext-utils";
+import { localize } from "../../../utils/localize";
+import type { IDeployRevisionDraftContext } from "./IDeployRevisionDraftContext";
+
+export class DeployRevisionDraftConfirmStep extends AzureWizardPromptStep<IDeployRevisionDraftContext> {
+    public async prompt(context: IDeployRevisionDraftContext): Promise<void> {
+        await context.ui.showWarningMessage(
+            localize('deployRevisionWarning', 'This will deploy a new revision to container app "{0}".', context.containerApp?.name),
```
nturinski, line 13: I think we should steer away from the word "revision" since we don't really expose that anywhere else. We should probably just have something to the effect that we're "Deploying to container app" or "Updating container app" or something.

src/commands/revisionDraft/deployRevisionDraft/DeployRevisionDraftConfirmStep.ts
```diff
@@ -0,0 +1,22 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { AzureWizardPromptStep } from "@microsoft/vscode-azext-utils";
+import { localize } from "../../../utils/localize";
+import type { IDeployRevisionDraftContext } from "./IDeployRevisionDraftContext";
+
+export class DeployRevisionDraftConfirmStep extends AzureWizardPromptStep<IDeployRevisionDraftContext> {
+    public async prompt(context: IDeployRevisionDraftContext): Promise<void> {
+        await context.ui.showWarningMessage(
+            localize('deployRevisionWarning', 'This will deploy any unsaved changes to container app "{0}".', context.containerApp?.name),
```
nturinski, line 13: I think we should probably say something more similar to deploy.

`Are you sure you want to deploy to changes "{0}"? This will overwrite the previous deployment`

src/commands/revisionDraft/deployRevisionDraft/DeployRevisionDraftConfirmStep.ts
```diff
@@ -0,0 +1,22 @@
...
+            { modal: true },
+            { title: localize('continue', 'Continue') }
```
nturinski, line 15: I would have this title be "Deploy"

src/commands/revisionDraft/deployRevisionDraft/DeployRevisionDraftStep.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export class DeployRevisionDraftStep extends AzureWizardExecuteStep<IDeployRevisionDraftContext> {
+    public priority: number = 260;
+
+    public async execute(context: IDeployRevisionDraftContext, progress: Progress<{ message?: string | undefined; increment?: number | undefined }>): Promise<void> {
+        const containerApp: ContainerAppModel = nonNullProp(context, 'containerApp');
+        const containerAppEnvelope = await getContainerEnvelopeWithSecrets(context, context.subscription, containerApp);
+        containerAppEnvelope.template = nonNullProp(context, 'template');
+
+        const creatingRevision: string = localize('creatingRevision', 'Creating revision...');
```
nturinski, line 24: Still says revision here.

src/commands/revisionDraft/deployRevisionDraft/DeployRevisionDraftStep.ts
```diff
@@ -0,0 +1,43 @@
...
+
+        const id: string = containerApp.revisionsMode === KnownActiveRevisionsMode.Single ? containerApp.id : `${containerApp.id}/${RevisionDraftItem.idSuffix}`;
+
+        await ext.state.runWithTemporaryDescription(id, creatingRevision, async () => {
+            await updateContainerApp(context, context.subscription, containerAppEnvelope);
+            const updatedContainerApp = await ContainerAppItem.Get(context, context.subscription, containerApp.resourceGroup, containerApp.name);
+
+            if (containerApp.revisionsMode === KnownActiveRevisionsMode.Multiple) {
+                // Display the name of the newly created revision when in multiple revisions mode
+                context.activityTitle = localize('deployRevision', 'Deploy revision "{0}" to container app "{1}"', updatedContainerApp.latestRevisionName, containerApp.name);
```
nturinski, line 35: Still says revision here.

PR #418, opened 2023-07-31
Add support for `deleteSecret`

MicroFish91:
image

image


src/utils/pickItem/pickSecret.ts
```diff
@@ -0,0 +1,44 @@
...
+    return new ContextValueQuickPickStep(ext.rgApiV2.resources.azureResourceTreeDataProvider, {
+        contextValueFilter: { include: SecretsItem.contextValueRegExp },
+        skipIfOne: true,
+    });
+}
+
+function getPickSecretStep(): AzureWizardPromptStep<QuickPickWizardContext> {
+    return new ContextValueQuickPickStep(ext.rgApiV2.resources.azureResourceTreeDataProvider, {
+        contextValueFilter: { include: SecretItem.contextValueRegExp },
+        skipIfOne: true,
```
bwateratmsft, line 25: IMO this step should not skip--it's the last step and shows what they're actually about to delete.

bwateratmsft (approved):

Issue #420, opened 2023-08-01
Hook up `Scaling` commands and tree items to utilize the new revision draft design


MicroFish91:
- [x] Scaling
- [x] Scaling Range (min/max replicas)
- [x] Add Scale Rule
- [x] Delete Scale Rule

PR #423, opened 2023-08-03
Enable updating revision drafts through commands & leverage with `Scale Rules`

### LINKED ISSUE #254: [Suggestion] It would be better to add a notification "Adding HTTP scaling rule..." when executing "Add Scale Rule..." command
**OS:** All
**Azure Container Apps Build Version:** 20230213.10
**Azure Resource Group Build Version:** 20230213.13

**Repro Steps:**
1. F1 -> Azure Container Apps: Add Scale Rule... -> Select a container app environment -> Select one container app -> Select a revision -> Enter a name for the new scale rule -> HTTP scaling -> Enter a number of concurrent requests.
2. Check whether there is a notification "Adding HTTP scaling rule..." when executing "Add Scale Rule..." command.

**Actual:**
There is no notification "Adding HTTP scaling rule..." when executing "Add Scale Rule..." command.

**Suggestion:**
It would be better to add a notification "Adding HTTP scaling rule..." when executing "Add Scale Rule..." command.

**More Info:**
1. There is only one notification "Successfully added HTTP scaling rule..." after creation.
image
2. Both "Adding..." and "Successfully added..." logs show in the output panel.
image

MicroFish91:
Addresses 2/3 of #420 
Closes #254 

**Summary**:
- Added a pathway for commands to leverage when updating the revision draft template
- Hooked up `Scale Rules` to update the revision draft template rather than deploy a new revision
- Refactor misc `Add scale rule...` improvements (this was my first big feature and going back I noticed a lot of things I wanted to change or fix 😮‍💨)
- Turn all scale items into classes, and add logic so that unsaved changes are detected and shown in the tree view

**Future**:
- Ended up removing `+ Create a secret` related logic from this PR, will add it in the follow-up one
- Will probably go back and refactor `Add scale rule...` a little bit more in the near future, mostly to clean up the validate input logic
- Still need to add the rest of the CRUD ops for `Scale Rules`

**Demo**:
- [Scale Rules](https://microsoft-my.sharepoint.com/:v:/p/matfisher/ETGXi2LTIX1NnI0lXlA1OpMBY2bze88sGqc7FUD-l9k-1Q?e=YgNjfD)

Note: It works basically the same for multiple revisions mode.  I was kind of lazy about recording that demo, but I at least tested the same ops in multiple revisions mode and it behaved the same way as in single revisions mode minus it being attached to a draft node.


src/commands/scaling/addScaleRule/queue/QueueNameStep.ts
```diff
@@ -3,24 +3,20 @@
...
         context.queueName = (await context.ui.showInputBox({
             prompt: localize('queueNamePrompt', 'Enter a name for the queue.'),
-            validateInput: (value: string | undefined): string | undefined => this.validateInput(value)
+            validateInput: this.validateInput
         })).trim();
     }
 
     public shouldPrompt(context: IAddScaleRuleContext): boolean {
-        return context.queueName === undefined;
+        return !context.queueName;
```
MicroFish91, line 19: Changed a lot of these to just a `!` because I feel like there are other falsey values besides `undefined` that I would like to prevent as valid input.

nturinski:
Just a sidenote: This would be a _great_ place where we have a follow up action for the activity log. Underneath Add HTTP Scale Rule, it would awesome to have the Deploy/Update command underneath it to make it more obvious that they are unsaved changes.

nturinski:
Also should change it from "creating revision" to "updating container app" or "deploying to container app"

MicroFish91:
> Also should change it from "creating revision" to "updating container app" or "deploying to container app"

Yeah good call, I'll update this in a separate PR

src/commands/scaling/addScaleRule/AddScaleRuleStep.ts
```diff
@@ -3,38 +3,53 @@
...
-        const adding = localize('addingScaleRule', 'Adding {0} rule "{1}" to "{2}"...', context.ruleType, context.ruleName, context.containerApp.name);
-        const added = localize('addedScaleRule', 'Successfully added {0} rule "{1}" to "{2}".', context.ruleType, context.ruleName, context.containerApp.name);
+    constructor(baseItem: RevisionsItemModel) {
+        super(baseItem);
+    }
+
+    public async execute(context: IAddScaleRuleContext, progress: Progress<{ message?: string | undefined; increment?: number | undefined }>): Promise<void> {
+        let adding: string | undefined;
+        let added: string | undefined;
+        if (context.containerApp.revisionsMode === KnownActiveRevisionsMode.Single) {
```
nturinski, line 27: nit: I think you could have a "resourceName" variable here that is either "container app" or "revision" and replace that in the adding and added strings.  It's just a little messy to see these long sentences written out twice (at least in GitHub)

src/commands/scaling/addScaleRule/AddScaleRuleStep.ts
```diff
@@ -3,38 +3,53 @@
...
+        this.revisionDraftTemplate.scale ||= {};
+        this.revisionDraftTemplate.scale.rules ||= [];
 
         const scaleRule: ScaleRule = this.buildRule(context);
-        this.integrateRule(context, template.scale.rules, scaleRule);
+        this.integrateRule(context, this.revisionDraftTemplate.scale.rules, scaleRule);
+        this.updateRevisionDraftWithTemplate();
+
+        // Artificial delay to make the activity log look like it's performing an action
+        await delay(1000);
```
nturinski, line 46: I feel like a second is a little bit long to block a user for something that's artificial.

My understanding is that it's so that you can see `Adding scale rule...` appear, but even if they don't see that progress message, they can still see that the "Added scale rule to container app (draft)` completed successfully, right?

I don't know if we need that progress message since it's instantaneous to begin with.

nturinski: If we do want to keep the delay though, I think we should put the delay in `updateRevisionDraftWithTemplate` so that all of the actions behave similarily.

src/commands/scaling/addScaleRule/addScaleRule.ts
```diff
@@ -3,75 +3,45 @@
...
-const addScaleRuleTitle = localize('addScaleRuleTitle', 'Add Scale Rule');
+import { ScaleRuleTypeListStep } from "./ScaleRuleTypeListStep";
 
 export async function addScaleRule(context: IActionContext, node?: ScaleRuleGroupItem): Promise<void> {
-    const { subscription, containerApp, revision } = node ?? await getContainerAppAndRevision(context);
+    const item: ScaleRuleGroupItem = node ?? await pickScaleRuleGroup(context, { autoSelectDraft: true });
+    const { subscription, containerApp, revision } = item;
+
+    // Branching path reasoning: https://github.com/microsoft/vscode-azurecontainerapps/blob/main/src/commands/revisionDraft/README.md
+    let scale: Scale | undefined;
```
nturinski, line 22: We're going to have to duplicate this logic a lot (any time we add/update the draft). I think we should go ahead and extract this out as a util function.

The function should just return `revision` or `template` depending on the revision type.

src/tree/scaling/ScaleRuleGroupItem.ts
```diff
@@ -3,43 +3,83 @@
...
+
+    // Used as the basis for the view; can reflect either the original or the draft changes
+    private readonly scaleRules: ScaleRule[];
+
+    constructor(
+        readonly subscription: AzureSubscription,
+        readonly containerApp: ContainerAppModel,
+        readonly revision: Revision
+    ) {
+        if (this.hasUnsavedChanges()) {
```
nturinski, line 32: It may be easier to discuss this offline, but I'll write a short blurb and we can discuss further.

But I think `RevisionsDraftModel` should have a function to set the properties based on if it has unsaved changes or not. That way, the `ScaleRuleGroupItem` (and any other item that needs this logic) doesn't have to have the branching logic in its constructor.

`RevisionDraftModel` should also have two abstract methods: `setProperties` and `setPropertiesDraft` (or something like that-- maybe `setTreeItem`?) that does the assignments.

So I think that would look something like this:

```
// For tree items that depend on the container app's revision draft template
export abstract class RevisionsDraftModel {
    hasUnsavedChanges: () => boolean | Promise<boolean>;
    abstract function setProperties(): void;
    abstract function setPropertiesDraft(): void;
    constructor(...) {
        this.hasUnsavedChanges() ? this.setPropertiesDraft() : this.setProperties();
    }
}

File: src/tree/scaling/ScaleRuleItem.ts [hunk 6]
```diff
@@ -3,52 +3,76 @@
...
+            return localize('azureQueue', 'Azure Queue');
+        } else if (this.scaleRule.custom) {
+            return localize('custom', 'Custom');
+        } else {
+            return localize('unknown', 'Unknown');
+        }
+    }
+
+    private get parentResource(): ContainerAppModel | Revision {
+        return this.containerApp.revisionsMode === KnownActiveRevisionsMode.Single ? this.containerApp : this.revision;
```
nturinski, line 50: Another spot we can use the util to choose between app or revision.

src/commands/revisionDraft/RevisionDraftFileSystem.ts
```diff
@@ -51,22 +50,15 @@ export class RevisionDraftFileSystem implements FileSystemProvider {
     }
 
     // Create
```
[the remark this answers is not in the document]
MicroFish91: Hmm, I feel like it helps me organize the methods into logical groups quickly.  Without the headers there to quickly parse, it's kind of hard to determine where to add new methods without taking a bit of time to re-organize the methods into groups mentally.  Unless you mean organizing them like that is unnecessary in general.

src/tree/scaling/ScaleRuleGroupItem.ts
MicroFish91: Ran into some issues here with implementing it this way.  `hasUnsavedChanges` usually returns a promise that needs to be awaited.  But you can't async/await in a constructor... potentially making this way much less straightforward to implement

src/commands/scaling/addScaleRule/ScaleRuleNameStep.ts
```diff
@@ -3,40 +3,58 @@
...
-        return localize('invalidChar', `A name must consist of lower case alphanumeric characters or '-', and must start and end with an alphanumeric character.`);
+        return undefined;
     }
 
-    const scaleRuleExists: boolean = !!scaleRules?.some((rule) => {
-        return rule?.name?.length && rule?.name === name;
-    });
-    if (scaleRuleExists) {
-        return localize('scaleRuleExists', 'The scale rule "{0}" already exists in container app "{1}". Please enter a unique name.', name, containerAppName);
+    private async validateNameAvailable(context: IAddScaleRuleContext, name: string): Promise<string | undefined> {
```
MicroFish91, line 37: Added validation that always does a network call to ensure we aren't working with stale data when we run our verification logic.

src/commands/revisionDraft/RevisionDraftFileSystem.ts
```diff
@@ -50,23 +49,15 @@ export class RevisionDraftFileSystem implements FileSystemProvider {
...
-        if (item instanceof ContainerAppItem) {
-            /**
-             * Sometimes there are micro-differences present between the latest revision template and the container app template.
-             * They end up being essentially equivalent, but not so equivalent as to always pass a deep copy test (which is how we detect unsaved changes).
-             *
-             * Since only container app template data is shown in single revision mode, and since revision data is not directly present on
-             * the container app tree item without further changes, it is easier to just use the container app template
-             * as the primary source of truth when in single revision mode.
-             */
+        if (item instanceof ContainerAppItem || item.containerApp.revisionsMode === KnownActiveRevisionsMode.Single) {
```
alexweininger, line 60: We like to avoid using `instanceof` if we can. Here it can be replaced with an `isContainerAppItem` function.

https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

src/tree/scaling/ScaleItem.ts
```diff
@@ -54,7 +45,17 @@ export class ScaleItem implements RevisionsItemModel, RevisionsDraftModel {
     }
 
     private get parentResource(): ContainerAppModel | Revision {
```
nturinski, line 47: Since this is private, is there any real reason to not just call `getParentResource` directly from the util?

src/commands/scaling/addScaleRule/IAddScaleRuleContext.ts
```diff
@@ -3,27 +3,28 @@
...
 import type { ExecuteActivityContext } from "@microsoft/vscode-azext-utils";
 import type { ContainerAppModel } from "../../../tree/ContainerAppItem";
 import type { IContainerAppContext } from "../../IContainerAppContext";
 
 export interface IAddScaleRuleContext extends IContainerAppContext, ExecuteActivityContext {
     // Make containerApp _required_
     containerApp: ContainerAppModel;
-
-    scaleRules: ScaleRule[];
+    parentResourceName: string;
```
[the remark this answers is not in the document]
MicroFish91: No, in general the parent resource name could be the container app name or the revision name.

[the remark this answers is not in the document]
MicroFish91: Potentially I could add `parentResource: ContainerAppItem | Revision` and just access the name property later when I need it.  Felt like overkill to tack the whole thing on at the time, but it would make it a lot less confusing for situations like this where parent is a bit ambiguous 🤔

src/tree/revisionManagement/RevisionDraftDescendantBase.ts
```diff
@@ -0,0 +1,32 @@
...
+import type { TreeElementBase } from "@microsoft/vscode-azext-utils";
+import type { AzureSubscription } from "@microsoft/vscode-azureresources-api";
+import type { TreeItem } from "vscode";
+import type { ContainerAppModel } from "../ContainerAppItem";
+import type { RevisionsDraftModel } from "./RevisionDraftItem";
+
+/**
+ * Can be implemented by any tree item that has the potential to show up as a RevisionDraftItem's descendant
+ */
+export abstract class RevisionDraftDescendantBase implements RevisionsDraftModel {
```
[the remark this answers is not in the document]
MicroFish91: Yeah good catch, I should have included that as well

src/tree/scaling/ScaleItem.ts
MicroFish91: Nah not really, mostly just calling it it through a getter looks a little cleaner, but functionally adds nothing else

src/tree/revisionManagement/RevisionDraftDescendantBase.ts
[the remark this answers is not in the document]
MicroFish91: Unfortunately `ContainerAppItem` uses `RevisionsDraftModel` without `RevisionsItemModel` which is why I left them uncoupled, but I agree that the base item should include both

src/commands/scaling/addScaleRule/AddScaleRuleStep.ts
```diff
@@ -21,52 +19,37 @@ export class AddScaleRuleStep<T extends IAddScaleRuleContext> extends RevisionDr
...
-        this.integrateRule(context, this.revisionDraftTemplate.scale.rules, scaleRule);
+        context.scaleRule = this.buildRule(context);
+        this.integrateRule(context, this.revisionDraftTemplate.scale.rules, context.scaleRule);
         this.updateRevisionDraftWithTemplate();
 
-        // Artificial delay to make the activity log look like it's performing an action
-        await delay(1000);
-
-        ext.outputChannel.appendLog(added);
+        const resourceName = context.containerApp.revisionsMode === KnownActiveRevisionsMode.Single ? context.containerApp.name : this.baseItem.revision.name;
```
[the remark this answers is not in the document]
MicroFish91: I _think_ I get what you mean, but let me know if I was off with the change I made

src/tree/scaling/ScaleItem.ts
MicroFish91: `this.parentResource` vs `getParentResource(this.containerApp, this.revision)` each time

src/commands/revisionDraft/RevisionDraftFileSystem.ts
```diff
@@ -146,8 +135,30 @@ export class RevisionDraftFileSystem implements FileSystemProvider {
...
+            this.createRevisionDraft(item);
+        }
+
+        const newContent: Uint8Array = Buffer.from(JSON.stringify(template, undefined, 4));
+        const file: RevisionDraftFile = nonNullValue(this.draftStore.get(uri.path));
+        if (file.contents === newContent) {
+            return;
+        }
+
+        file.contents = newContent;
```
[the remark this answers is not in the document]
MicroFish91: Good catch, I think that separate function is already written actually.  I think I can just call `this.writeFile()` with the new contents

src/tree/scaling/ScaleRuleGroupItem.ts
MicroFish91: After some testing, ran into some limitations with doing it this way, will discuss more offline

src/tree/scaling/ScaleItem.ts
nturinski: Yeah, that's a good point. I agree.

src/commands/scaling/addScaleRule/IAddScaleRuleContext.ts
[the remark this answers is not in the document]
nturinski: Maybe just a comment about that then? I personally felt like it was pretty confusing.

nturinski (approved):
(seal of approval)

PR #431, opened 2023-08-09
Update `deleteManagedEnvironment` activity messages and general housekeeping

### LINKED ISSUE #389: Make managed environment activity delete description shorter
image <br />

In general, I want to start making our activity descriptions shorter.  One idea is to remove the `Deleting Container Apps environment "mwf-3"` or at least shorten it to something simpler (e.g. `Deleting environment...` or similar) since this information is already shown in the activity title.

MicroFish91:
Closes #389 

From: 
image

To:
image

### DIFF
--- a/src/commands/deleteManagedEnvironment/DeleteManagedEnvironmentStep.ts
+++ b/src/commands/deleteManagedEnvironment/DeleteManagedEnvironmentStep.ts
@@ -5,22 +5,22 @@
 
 import { ContainerAppsAPIClient } from "@azure/arm-appcontainers";
 import { AzureWizardExecuteStep, parseError } from "@microsoft/vscode-azext-utils";
-import { Progress, window } from "vscode";
+import { Progress } from "vscode";
 import { ext } from "../../extensionVariables";
 import { createContainerAppsAPIClient } from "../../utils/azureClients";
 import { localize } from "../../utils/localize";
-import { IDeleteManagedEnvironmentWizardContext } from "./IDeleteManagedEnvironmentWizardContext";
+import { IDeleteManagedEnvironmentContext } from "./IDeleteManagedEnvironmentContext";
 
-export class DeleteManagedEnvironmentStep extends AzureWizardExecuteStep<IDeleteManagedEnvironmentWizardContext> {
+export class DeleteManagedEnvironmentStep extends AzureWizardExecuteStep<IDeleteManagedEnvironmentContext> {
     public priority: number = 110;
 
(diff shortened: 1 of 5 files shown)


src/commands/deleteManagedEnvironment/DeleteManagedEnvironmentStep.ts
```diff
@@ -3,24 +3,24 @@
...
-export class DeleteManagedEnvironmentStep extends AzureWizardExecuteStep<IDeleteManagedEnvironmentWizardContext> {
+export class DeleteManagedEnvironmentStep extends AzureWizardExecuteStep<IDeleteManagedEnvironmentContext> {
     public priority: number = 110;
 
-    public async execute(context: IDeleteManagedEnvironmentWizardContext, progress: Progress<{ message?: string | undefined; increment?: number | undefined }>): Promise<void> {
-        const deleting: string = localize('deletingManagedEnv', 'Deleting Container Apps environment "{0}"... this may take several minutes.', context.managedEnvironmentName);
+    public async execute(context: IDeleteManagedEnvironmentContext, progress: Progress<{ message?: string | undefined; increment?: number | undefined }>): Promise<void> {
         const client: ContainerAppsAPIClient = await createContainerAppsAPIClient([context, context.subscription]);
 
+        const deleting: string = localize('deletingManagedEnv', 'Deleting environment... this may take several minutes...');
```
nturinski, line 20: I prefer having the name in here since it takes so long. Also don't like the added ellipsis at the end.

MicroFish91: What, really? It's already in the title though

MicroFish91: I tried without ellipses, and it looks kind of weird.  What phrase are you looking at in total as the replacement?

MicroFish91: image

nturinski: Ahhh sorry, I forgot that's how it looked on the activity log.

Maybe we just chnage it to `This may take a few minutes...`?

MicroFish91: I'm also cool with that, but let me give you some additional context just in case.  

The reason I kept the `Deleting environment...` portion prefixed was because the description could first show the `Deleting container app {x}...` message if the managed environment has children - so keeping it does let the user know what the current step is actually doing and lets the user know that the process has progressed.  That said, it is a bit long with it, so I'm happy to remove it... I personally don't have very strong opinions between these two descriptions

nturinski (approved):

Issue #425, opened 2023-08-04
Implement the `Click2Run` command


MicroFish91:
Create a workspace command that allows a user to deploy their workspace project to ACA, while providing a minimal number of prompts/configurations for easy deployment.  We will leverage the ACR tasks method for deploying new images.  Project will require a `Dockerfile` to deploy.

**Todo**:
- [x] Determine which values can be defaulted/assumed (Team - 8/10)
- [x] Enable creating an Azure Container Registry if the user doesn't already have one available (@motm32)
- [ ] Implement the command/contract (@MicroFish91)

MicroFish91:
Update: First iteration will assume all resources need to be created from scratch.  A future iteration may include saving values as local project settings to save time for repeat deployments.  Also possible future iteration may include a `basic` vs `advanced` version of this command.

MicroFish91:
For creating a new ACR, this is relevant:
https://github.com/microsoft/vscode-azurecontainerapps/issues/78

We need to keep in mind that in some cases we are pushing images, and in other situations we are pulling images and design around those needs accordingly.

PR #435, opened 2023-08-15
Add Create Azure Container Registry command

motm32:
Part of #425.

### DIFF
--- a/package.nls.json
+++ b/package.nls.json
@@ -41,5 +41,5 @@
     "containerApps.openGitHubRepo": "Open Repo in GitHub",
     "containerApps.startStreamingLogs": "Start Streaming Logs...",
     "containerApps.stopStreamingLogs": "Stop Streaming Logs...",
-    "containerApps.createAcr": "Create Azure Container Registry"
+    "containerApps.createAcr": "Create Azure Container Registry..."
 }
--- a/package.nls.json
+++ b/package.nls.json
@@ -40,5 +40,6 @@
     "containerApps.disconnectRepo": "Disconnect from Repo",
     "containerApps.openGitHubRepo": "Open Repo in GitHub",
     "containerApps.startStreamingLogs": "Start Streaming Logs...",
-    "containerApps.stopStreamingLogs": "Stop Streaming Logs..."
+    "containerApps.stopStreamingLogs": "Stop Streaming Logs...",
+    "containerApps.createAcr": "Create Azure Container Registry"
 }
--- a/src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/ICreateAcrContext.ts
+++ b/src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/ICreateAcrContext.ts
@@ -1,0 +1,15 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { KnownSkuName, Registry } from "@azure/arm-containerregistry";
+import { IResourceGroupWizardContext } from "@microsoft/vscode-azext-azureutils";
+import { ExecuteActivityContext } from "@microsoft/vscode-azext-utils";
+
+export interface ICreateAcrContext extends IResourceGroupWizardContext, ExecuteActivityContext {
+    registryName?: string;
+    sku?: KnownSkuName;
+    registry?: Registry;
+    newResourceGroupName?: string;
+}
--- a/src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/ICreateAcrContext.ts
+++ b/src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/ICreateAcrContext.ts
@@ -8,8 +8,7 @@ import { IResourceGroupWizardContext } from "@microsoft/vscode-azext-azureutils"
 import { ExecuteActivityContext } from "@microsoft/vscode-azext-utils";
 
 export interface ICreateAcrContext extends IResourceGroupWizardContext, ExecuteActivityContext {
-    registryName?: string;
+    newRegistryName?: string;
     sku?: KnownSkuName;
     registry?: Registry;
-    newResourceGroupName?: string;
 }
--- a/src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/RegistryCreateStep.ts
+++ b/src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/RegistryCreateStep.ts
@@ -7,12 +7,12 @@ import { ContainerRegistryManagementClient } from "@azure/arm-containerregistry"
 import { LocationListStep } from "@microsoft/vscode-azext-azureutils";
 import { AzureWizardExecuteStep, nonNullProp } from "@microsoft/vscode-azext-utils";
 import { createContainerRegistryManagementClient } from "../../../../../../utils/azureClients";
-import { ICreateAcrContext } from "./ICreateAcrContext";
+import { CreateAcrContext } from "./CreateAcrContext";
 
-export class RegistryCreateStep extends AzureWizardExecuteStep<ICreateAcrContext> {
+export class RegistryCreateStep extends AzureWizardExecuteStep<CreateAcrContext> {
     public priority: number = 150;
 
-    public async execute(context: ICreateAcrContext): Promise<void> {
+    public async execute(context: CreateAcrContext): Promise<void> {
         const client: ContainerRegistryManagementClient = await createContainerRegistryManagementClient(context);
 
         context.registry = await client.registries.beginCreateAndWait(
@@ -20,13 +20,13 @@ export class RegistryCreateStep extends AzureWizardExecuteStep<ICreateAcrContext
             nonNullProp(context, 'newRegistryName'),
             {
                 location: (await LocationListStep.getLocation(context)).name,
-                sku: { name: nonNullProp(context, 'sku') },
+                sku: { name: nonNullProp(context, 'newRegistrySku') },
                 adminUserEnabled: true
             }
         );
     }
 
-    public shouldExecute(context: ICreateAcrContext): boolean {
-        return !!context.newRegistryName && !!context.sku;
+    public shouldExecute(context: CreateAcrContext): boolean {
+        return !!context.newRegistryName && !!context.newRegistrySku;
     }
 }
--- a/src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/RegistryCreateStep.ts
+++ b/src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/RegistryCreateStep.ts
@@ -1,0 +1,32 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { ContainerRegistryManagementClient } from "@azure/arm-containerregistry";
+import { LocationListStep } from "@microsoft/vscode-azext-azureutils";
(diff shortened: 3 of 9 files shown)


nturinski:
> Don't want to use Docker's existing create ACR command?

The main reason we don't want to use Docker for this is because the code-to-cloud scenario should be enabled in VS Code for the Web which we wouldn't be able to support if we use the Docker extension.

bwateratmsft:
That's a very good reason. :smile:

src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/ICreateAcrContext.ts
```diff
@@ -0,0 +1,15 @@
...
+import { KnownSkuName, Registry } from "@azure/arm-containerregistry";
+import { IResourceGroupWizardContext } from "@microsoft/vscode-azext-azureutils";
+import { ExecuteActivityContext } from "@microsoft/vscode-azext-utils";
+
+export interface ICreateAcrContext extends IResourceGroupWizardContext, ExecuteActivityContext {
+    registryName?: string;
+    sku?: KnownSkuName;
+    registry?: Registry;
+    newResourceGroupName?: string;
+}
```
MicroFish91, line 15: Nit:  I think `IResourceGroupWizardContext` already provides `newResourceGroupName`, so you can probably omit that.  I also slightly prefer `registryName`  to be named `newRegistryName` to match our typical naming convention.

src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/RegistryCreateStep.ts
```diff
@@ -0,0 +1,32 @@
...
+            {
+                location: (await LocationListStep.getLocation(context)).name,
+                sku: { name: nonNullProp(context, 'sku') },
+                adminUserEnabled: true
+            }
+        );
+    }
+
+    public shouldExecute(): boolean {
+        return true;
```
MicroFish91, line 30: Not sure, but maybe this should be `return !!context.newRegistryName && !!context.sku;`

src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/RegistryNameStep.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    private async validateInput(context: ICreateAcrContext, name: string | undefined): Promise<string | undefined> {
+        name = name ? name.trim() : '';
+
+        const min = 5;
+        const max = 50;
+        if (name.length < min || name.length > max) {
+            return localize('validationLengthError', 'The name must be between {0} and {1} characters.', min, max);
+        } else if (!/^[a-z][a-zA-Z0-9]*$/.test(name)) {
+            return localize('validateInputError', `Connection names can only consist of alphanumeric characters.`); //make it between 5 and 50 character
```
MicroFish91, line 32: Old todo comment?

src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/RegistryNameStep.ts
```diff
@@ -0,0 +1,43 @@
...
+        } else {
+            const client: ContainerRegistryManagementClient = await createContainerRegistryManagementClient(context);
```
MicroFish91, line 34: So there's a separate property on `context.ui.showInputBox` now that we added called `asyncValidationTask` that only runs at the end when the user presses enter.  This is nice to use for validation requests like this because it condenses what would otherwise be a network call on each keystroke into potentially just one or a few at the end.  Hopefully there are some good examples already in the repo, if not feel free to reach out!

src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/SkuListStep.ts
```diff
@@ -0,0 +1,28 @@
...
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { KnownSkuName } from "@azure/arm-containerregistry";
+import { AzureWizardPromptStep, IAzureQuickPickItem } from "@microsoft/vscode-azext-utils";
+import { ICreateAcrContext } from "./ICreateAcrContext";
+
+export class SkuListStep extends AzureWizardPromptStep<ICreateAcrContext> {
+    public async prompt(context: ICreateAcrContext): Promise<void> {
+        const placeHolder: string = "Select a SKU";
```
MicroFish91, line 12: Localize 🙂

src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/createAcr.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    const title: string = localize('createAcr', "Create container registry");
+
+    const promptSteps: AzureWizardPromptStep<ICreateAcrContext>[] = [
+        new RegistryNameStep(),
+        new SkuListStep(),
+    ];
+
+    const executeSteps: AzureWizardExecuteStep<ICreateAcrContext>[] = [
+        new ResourceGroupCreateStep(),
```
MicroFish91, line 34: Should we use a `ResourceGroupListStep` instead of the create step here?  I think it might be nice to also allow the user to choose their own resource group rather than force a new one to be created each time 🤔

src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/ICreateAcrContext.ts
```diff
@@ -0,0 +1,14 @@
...
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { KnownSkuName, Registry } from "@azure/arm-containerregistry";
+import { IResourceGroupWizardContext } from "@microsoft/vscode-azext-azureutils";
+import { ExecuteActivityContext } from "@microsoft/vscode-azext-utils";
+
+export interface ICreateAcrContext extends IResourceGroupWizardContext, ExecuteActivityContext {
+    newRegistryName?: string;
+    sku?: KnownSkuName;
```
MicroFish91, line 12: Nit: Add the new prefix to sku - `newRegistrySku`

package.nls.json
```diff
@@ -40,5 +40,6 @@
     "containerApps.disconnectRepo": "Disconnect from Repo",
     "containerApps.openGitHubRepo": "Open Repo in GitHub",
     "containerApps.startStreamingLogs": "Start Streaming Logs...",
-    "containerApps.stopStreamingLogs": "Stop Streaming Logs..."
+    "containerApps.stopStreamingLogs": "Stop Streaming Logs...",
+    "containerApps.createAcr": "Create Azure Container Registry"
```
alexweininger, line 44: Add the ellipsis since I'm assuming it involves user input.

```suggestion
    "containerApps.createAcr": "Create Azure Container Registry..."
```

src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/ICreateAcrContext.ts
alexweininger: Is there a specific reason we are being super verbose and adding new to the property names here?

src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/SkuListStep.ts
```diff
@@ -0,0 +1,29 @@
...
+import { KnownSkuName } from "@azure/arm-containerregistry";
+import { AzureWizardPromptStep, IAzureQuickPickItem } from "@microsoft/vscode-azext-utils";
+import { localize } from "../../../../../../utils/localize";
+import { ICreateAcrContext } from "./ICreateAcrContext";
+
+export class SkuListStep extends AzureWizardPromptStep<ICreateAcrContext> {
+    public async prompt(context: ICreateAcrContext): Promise<void> {
+        const placeHolder: string = localize("sku", "Select a SKU");
+        const picks: IAzureQuickPickItem<KnownSkuName>[] = [
+            { label: "Basic", data: KnownSkuName.Basic },
```
alexweininger, line 15: You could use the `KnownSkuName` enum for the label too.

src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/createAcr.ts
```diff
@@ -0,0 +1,54 @@
...
+export async function createAcr(context: IActionContext, node?: { subscription: AzureSubscription }): Promise<void> {
+    const subscription = node?.subscription ?? await subscriptionExperience(context, ext.rgApiV2.resources.azureResourceTreeDataProvider);
+
+    const wizardContext: ICreateAcrContext = {
+        ...context,
+        ...createSubscriptionContext(subscription),
+        ...(await createActivityContext())
+    };
+
+    const title: string = localize('createAcr', "Create container registry");
```
alexweininger, line 26:
```suggestion
    const title: string = localize('createAcr', "Create Azure Container Registry");
```

src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/createAcr.ts
MicroFish91: Yeah, I've been meaning to ask Nathan about that actually.  My guess is that it was determined to just always skip prompting since it is the top level container apps resource, and likely the other resources would also sit in that same resource group?

For this one I took a sneak peak at how Docker implemented it, and they are using a list step too I believe, so might be worth trying to keep it similar 🤔

image

alexweininger: Depends on if we think users will care about what RG they create the registry in. Are there benefits from putting it in the same RG as other resources?

src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/ICreateAcrContext.ts
MicroFish91: I just noticed we do that in a ton of places to differentiate between an existing resource vs. information gathering for creating a new one.  I figured it was kind of an unspoken convention, but happy to not do it that way as well

alexweininger: I was just curious if there was a place where not having the new specified was confusing. But I think it's fine. It can't hurt to add more context to the property name.

src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/createAcr.ts
motm32: As far as I know it doesn't make a difference which RG the registry is in. You can still use images from registry's outside of the resource group the CA is in but not sure about other use cases.

src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/ICreateAcrContext.ts
```diff
@@ -0,0 +1,14 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { KnownSkuName, Registry } from "@azure/arm-containerregistry";
+import { IResourceGroupWizardContext } from "@microsoft/vscode-azext-azureutils";
+import { ExecuteActivityContext } from "@microsoft/vscode-azext-utils";
+
+export interface ICreateAcrContext extends IResourceGroupWizardContext, ExecuteActivityContext {
```
nturinski, line 10: We all collectively decided to stop using the I prefix for interfaces, I think

src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/createAcr.ts
MicroFish91: I'm happy with however way you end up implementing, just thought to bring it up for discussion

MicroFish91 (approved):

PR #441, opened 2023-09-01
Improve `validationUtils` invalid input message

### LINKED ISSUE #437: [Suggestion] It would be better to remove "one of" from the error message when creating a secret with an invalid name
**OS:** Win 10
**Build Version:** 20230815.2

**Repro Steps:**
1. Create a secret with the name "test-" for a container app, an error occurs.
2. Check whether the error message shows as 'A name must consist of lower case alphanumeric characters or **the following symbol: "-"**, and must start and end with a lower case alphanumeric characters'.

**Actual:**
The error message shows as 'A name must consist of lower case alphanumeric characters or **one of the following symbols: "-"**, and must start and end with a lower case alphanumeric characters'.
image

**Suggestion:**
It would be better to remove "one of" from the error message when creating a secret with an invalid name.

MicroFish91:
Closes #437


MicroFish91:
Closing in favor of a better solution

PR #442, opened 2023-09-01
Enforce new execute step priority ranges

MicroFish91:
As part of the express deploy command work, I'm revisiting all the execute step priorities to ensure creations are completed in the correct order (a lot of these commands were being called programmatically by the user before, and so the execute step priority values hadn't been especially important).  

Let me know if you think a general priority markdown file like this would be useful for tracking.


nturinski (approved):
Damn I thought I approved this already. It had the seal!

PR #443, opened 2023-09-01
Add `canSymbolsRepeat` option to `validateUtils` and improve invalid input messages

### LINKED ISSUE #437: [Suggestion] It would be better to remove "one of" from the error message when creating a secret with an invalid name

MicroFish91:
Closes #437 

Added the non-repeat option because a lot of Azure resources have this naming restriction and I'd like to be able to leverage this option for upcoming work.

### DIFF
--- a/src/commands/secret/addSecret/SecretNameStep.ts
+++ b/src/commands/secret/addSecret/SecretNameStep.ts
@@ -29,9 +29,8 @@ export class SecretNameStep extends AzureWizardPromptStep<ISecretContext> {
             return validateUtils.getInvalidLengthMessage();
         }
 
-        const symbols: string = '-';
-        if (!validateUtils.isLowerCaseAlphanumericWithSymbols(value, symbols, true /** canSymbolsRepeat */)) {
-            return validateUtils.getInvalidLowerCaseAlphanumericWithSymbolsMessage(symbols, true /** canSymbolsRepeat */);
+        if (!validateUtils.isLowerCaseAlphanumericWithSymbols(value)) {
+            return validateUtils.getInvalidLowerCaseAlphanumericWithSymbolsMessage();
         }
 
         const secrets: Secret[] = context.containerApp?.configuration?.secrets ?? [];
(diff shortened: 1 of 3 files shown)


src/commands/secret/addSecret/SecretNameStep.ts
```diff
@@ -29,8 +29,9 @@ export class SecretNameStep extends AzureWizardPromptStep<ISecretContext> {
             return validateUtils.getInvalidLengthMessage();
         }
 
-        if (!validateUtils.isLowerCaseAlphanumericWithSymbols(value)) {
-            return validateUtils.getInvalidLowerCaseAlphanumericWithSymbolsMessage();
+        const symbols: string = '-';
```
nturinski, line 32: I might be crazy, but because of how the property is named, I keep accidentally thinking that this is turning on a rule that _allows_ symbols to repeat, not that it's checking against it.

nturinski: Wait maybe I'm not crazy? But in that case, I think we shouldn't default `canSymbolsRepeat` to false.

MicroFish91: > I might be crazy, but because of how the property is named, I keep accidentally thinking that this is turning on a rule that allows symbols to repeat, not that it's checking against it.

Not crazy, that's exactly what it does!

What are you thinking?  Default `canSymbolsRepeat` to true? Or rename the option to something else?

I chose to have this be the default behavior because it is probably the more common/standard behavior for Azure resource naming, and I like passing in the least amount of vars to accomplish the more default/common behavior.

MicroFish91: Based on our discussion, it looks like there is a lot of non-containerApp resources that do allow repeating symbols as default, so making the default behavior true instead of false.

nturinski (approved):

PR #446, opened 2023-09-06
Add ability to create new secrets while building `Azure Queue` scale rules

### LINKED ISSUE #186: It would be better to add a "+Create a secret" option in the step "Choose a secret reference" [Secret management]
**OS:** All
**Build Version:** 20220908.8

**Repro Steps:**
1. Expand one container app **without any secrets** -> Right click "Scale Rules" node and select "Add Scale Rules..." -> Enter a name for the new scale rule -> Select "Azure queue" -> Enter a name for the queue -> Enter a queue length -> Press Enter.
2. Check whether there is a "+Create a secret" option in the next step "Choose a secret reference".

**Actual:**
There is no "+Create a secret" option in the next step "Choose a secret reference", an error occurs when loading the secret.
image

**Suggestion:**
It would be better to add a "+Create a secret" option in the step "Choose a secret reference", otherwise the user need to add a secret in the portal.

**More Info:**
The new scale rule with "Azure queue" selecting can be created successfully if there is at least one secret in the container app.

MicroFish91:
Closes #186 

Same as the previous scale rule PR demo:
[Scale Rules](https://microsoft-my.sharepoint.com/:v:/p/matfisher/ETGXi2LTIX1NnI0lXlA1OpMBY2bze88sGqc7FUD-l9k-1Q?e=YgNjfD)


src/commands/secret/SecretListStep.ts
```diff
@@ -0,0 +1,49 @@
...
+import type { Secret } from '@azure/arm-appcontainers';
+import { AzureWizardPromptStep, IAzureQuickPickItem, IWizardOptions, nonNullProp } from '@microsoft/vscode-azext-utils';
+import { ContainerAppModel, getContainerEnvelopeWithSecrets } from '../../tree/ContainerAppItem';
+import { localize } from '../../utils/localize';
+import type { ISecretContext } from './ISecretContext';
+import { SecretCreateStep } from './addSecret/SecretCreateStep';
+import { SecretNameStep } from './addSecret/SecretNameStep';
+import { SecretValueStep } from './addSecret/SecretValueStep';
+
+export class SecretListStep extends AzureWizardPromptStep<ISecretContext> {
```
nturinski, line 15: image

nturinski (commented):
My bad about this PR https://github.com/microsoft/vscode-azurecontainerapps/pull/442. I recommend merging this first and then pulling down the changes to make this PR look cleaner.

nturinski (approved):

PR #447, opened 2023-09-08
Scaffold `deployWorkspaceProject` command and add open workspace logic

MicroFish91:
Related to: #425 

This PR begins the process of adding the new `deployWorkspaceProject` express deployment command.

### Additions / Changes
1. Scaffolded the new command to show up in the Azure workspace pane
2. Added logic to ensure a workspace is open with a Dockerfile on init; otherwise prompt user to open a project in workspace

### Demo
- Here's what the full thing will eventually look like: [demo](https://microsoft-my.sharepoint.com/:v:/p/matfisher/EXziurykRO9Bhmg_Yp3Nq0oBx02_-ZC2CNzM8EWf_gEF5Q?nav=eyJyZWZlcnJhbEluZm8iOnsicmVmZXJyYWxBcHAiOiJPbmVEcml2ZUZvckJ1c2luZXNzIiwicmVmZXJyYWxBcHBQbGF0Zm9ybSI6IldlYiIsInJlZmVycmFsTW9kZSI6InZpZXciLCJyZWZlcnJhbFZpZXciOiJNeUZpbGVzTGlua0RpcmVjdCJ9fQ&e=NVcKcB&xsdata=MDV8MDF8fDQ1ZGViOTg0MjIzZTQ1MGE3NDE2MDhkYmFhNmZjMDcxfDcyZjk4OGJmODZmMTQxYWY5MWFiMmQ3Y2QwMTFkYjQ3fDB8MHw2MzgyOTExNjkxNDk4MDk2ODZ8VW5rbm93bnxWR1ZoYlhOVFpXTjFjbWwwZVZObGNuWnBZMlY4ZXlKV0lqb2lNQzR3TGpBd01EQWlMQ0pRSWpvaVYybHVNeklpTENKQlRpSTZJazkwYUdWeUlpd2lWMVFpT2pFeGZRPT18MXxMMk5vWVhSekx6RTVPbTFsWlhScGJtZGZUa2RhYVZsVVNYZFpNbFYwVFhwa2FFNURNREJQVkVGM1RGUm9hVmxYV1hST2VtaHJUVVJhYlZwVWJHbE5SRlpxUUhSb2NtVmhaQzUyTWk5dFpYTnpZV2RsY3k4eE5qa3pOVEU1TWpBMU9UYzJ8ZmUzZDBjMGM4MTgxNDliOTc0MTYwOGRiYWE2ZmMwNzF8MmNhYWI0Y2RhOGFiNDA4YmIyMzYwMGYzN2FkZmM2MGU%3D&sdata=YzZZMFNnT3c2ZE5RTDlmbFBJbW5JYWZGR2JMMW51SUc3T1BHKzFEbTJwST0%3D&ovuser=72f988bf-86f1-41af-91ab-2d7cd011db47%2Cmatfisher%40microsoft.com)
- Here's the demo for this part of the PR: [demo](https://microsoft-my.sharepoint.com/:v:/p/matfisher/Ee2xiWjlQlxBmbzX1uWAnKkB61Y-QgJg3xvkX39vx3fDJw?nav=eyJyZWZlcnJhbEluZm8iOnsicmVmZXJyYWxBcHAiOiJPbmVEcml2ZUZvckJ1c2luZXNzIiwicmVmZXJyYWxBcHBQbGF0Zm9ybSI6IldlYiIsInJlZmVycmFsTW9kZSI6InZpZXciLCJyZWZlcnJhbFZpZXciOiJNeUZpbGVzTGlua0RpcmVjdCJ9fQ&e=3eqldC)


package.json
```diff
@@ -230,7 +235,31 @@
                 "category": "Azure Container Apps"
             }
         ],
+        "submenus": [
+            {
+                "id": "containerApps.submenus.workspaceActions",
+                "label": "Azure Container Apps",
+                "icon": "resources/azure-containerapps.svg"
```
nturinski, line 242: You need to define the icon paths like this: https://github.com/microsoft/vscode-azurefunctions/pull/3784/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R352-R355

Or else, the icon won't show up in light theme for whatever reason.

package.nls.json
```diff
@@ -9,6 +9,7 @@
     "containerApps.editContainerApp": "Edit Container App (Advanced)...",
     "containerApps.deployImage": "Update Container App Image...",
     "containerApps.deployImageApi": "Update Container App Image (API)...",
+    "containerApps.deployWorkspaceProject": "Deploy Project from Workspace...",
```
nturinski, line 12: Should we call this `Create Container App from Workspace...`?

My thought process is that the user may infer that "deploy" means that they need to have an ca already created (especially since it's next to a create command)

nturinski: And following that logic, maybe we should rename `Create Container App...` to something like `Create Container App from Remote/Registry...` or

src/commands/deployWorkspaceProject/getDefaultValues/getDefaultContextValues.ts
```diff
@@ -0,0 +1,30 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import type { ISubscriptionActionContext } from "@microsoft/vscode-azext-utils";
+import type { DeployWorkspaceProjectContext } from "../DeployWorkspaceProjectContext";
+import { getWorkspaceProjectPaths } from "./getWorkspaceProjectPaths";
+
+export async function getDefaultContextValues(context: ISubscriptionActionContext): Promise<Partial<DeployWorkspaceProjectContext>> {
```
nturinski, line 10: I'm just going to ignore things that are commented out for code-review for now.

src/commands/deployWorkspaceProject/getDefaultValues/getWorkspaceProjectPaths.ts
```diff
@@ -0,0 +1,36 @@
...
+        // Silently throw an exception to exit the command while VS Code reloads the new workspace
+        throw new UserCancelledError();
+    }
+
+    // Check if chosen workspace has a Dockerfile at its root
+    const dockerfileUris: Uri[] = await workspace.findFiles(DOCKERFILE_GLOB_PATTERN);
+    if (!dockerfileUris.length) {
+        throw new Error(localize('noDockerfileError', 'Unable to locate a Dockerfile in your project\'s root.'));
+    } else if (dockerfileUris.length > 1) {
+        throw new Error(localize('multipleDockerfileError', 'Multiple Dockerfiles found. Unable to determine the correct Dockerfile to use in your project\'s root.'));
```
nturinski, line 29: I feel like we could just show the user a quick pick of dockerfiles so that they can select which one that they want. Maybe after we merge this, but I think that's kind of a must since some projects have test dockerfiles, prod dockerfiles, etc.

Forcing them to move/delete the files seems like a crummy UX.

src/utils/workspaceUtils.ts
```diff
@@ -22,9 +23,23 @@ export async function selectWorkspaceFile(context: IActionContext, placeHolder:
...
     return input?.data || (await context.ui.showOpenDialog(options))[0].path;
 }
+
+export async function getRootWorkspaceFolder(placeHolder?: string): Promise<WorkspaceFolder | undefined> {
+    if (!workspace.workspaceFolders?.length) {
+        return undefined;
+    } else if (workspace.workspaceFolders?.length === 1) {
+        return workspace.workspaceFolders[0];
+    } else {
+        const folder = await window.showWorkspaceFolderPick({ placeHolder: placeHolder ?? localize('selectRootWorkspace', 'Select a folder for your workspace') });
```
nturinski, line 39: I'm not sure I get what this is asking me.

Select a folder where my project is basically?

src/commands/deployWorkspaceProject/getDefaultValues/getWorkspaceProjectPaths.ts
MicroFish91: That's a good call, I agree

src/utils/workspaceUtils.ts
MicroFish91: Maybe it should be worded differently, but this condition would be for multi-root workspaces.

package.nls.json
MicroFish91: Maybe worth discussing with the CA team to see if there is a better name to use here.  I still kind of prefer `deploy` because it's more accurate than create, because this can also re-deploy to an existing

src/utils/workspaceUtils.ts
nturinski: Ahhh, okay, that makes more sense.

package.nls.json
nturinski: Yeah, it's a small nuance but I think we should do something about `Create Container App...` then. I have a bad feeling people may try to do that _first_ and see how many steps it is and then just give up.

src/commands/deployWorkspaceProject/getDefaultValues/getWorkspaceProjectPaths.ts
MicroFish91: Added todo

nturinski (approved):
~~Do you mind making issues for some of my comments?~~ I particularly want one for the container app create command name-- don't want to forget to discuss that.

Actually don't worry about it. It's not like we're going to look at issues right before the meeting (: I'll just write it down somewhere for me to mention it next meeting.

PR #448, opened 2023-09-08
Add default naming logic to the `deployWorkspaceProject` command

MicroFish91:
Related to: #425 

## Additions / Changes

1. Refactored name validation in existing steps to be re-useable via static methods
2. Add default resource naming step w/ tests

### Demo
- Here's what the full thing will eventually look like: [demo](https://microsoft-my.sharepoint.com/:v:/p/matfisher/EXziurykRO9Bhmg_Yp3Nq0oBx02_-ZC2CNzM8EWf_gEF5Q?nav=eyJyZWZlcnJhbEluZm8iOnsicmVmZXJyYWxBcHAiOiJPbmVEcml2ZUZvckJ1c2luZXNzIiwicmVmZXJyYWxBcHBQbGF0Zm9ybSI6IldlYiIsInJlZmVycmFsTW9kZSI6InZpZXciLCJyZWZlcnJhbFZpZXciOiJNeUZpbGVzTGlua0RpcmVjdCJ9fQ&e=NVcKcB&xsdata=MDV8MDF8fDQ1ZGViOTg0MjIzZTQ1MGE3NDE2MDhkYmFhNmZjMDcxfDcyZjk4OGJmODZmMTQxYWY5MWFiMmQ3Y2QwMTFkYjQ3fDB8MHw2MzgyOTExNjkxNDk4MDk2ODZ8VW5rbm93bnxWR1ZoYlhOVFpXTjFjbWwwZVZObGNuWnBZMlY4ZXlKV0lqb2lNQzR3TGpBd01EQWlMQ0pRSWpvaVYybHVNeklpTENKQlRpSTZJazkwYUdWeUlpd2lWMVFpT2pFeGZRPT18MXxMMk5vWVhSekx6RTVPbTFsWlhScGJtZGZUa2RhYVZsVVNYZFpNbFYwVFhwa2FFNURNREJQVkVGM1RGUm9hVmxYV1hST2VtaHJUVVJhYlZwVWJHbE5SRlpxUUhSb2NtVmhaQzUyTWk5dFpYTnpZV2RsY3k4eE5qa3pOVEU1TWpBMU9UYzJ8ZmUzZDBjMGM4MTgxNDliOTc0MTYwOGRiYWE2ZmMwNzF8MmNhYWI0Y2RhOGFiNDA4YmIyMzYwMGYzN2FkZmM2MGU%3D&sdata=YzZZMFNnT3c2ZE5RTDlmbFBJbW5JYWZGR2JMMW51SUc3T1BHKzFEbTJwST0%3D&ovuser=72f988bf-86f1-41af-91ab-2d7cd011db47%2Cmatfisher%40microsoft.com)


src/commands/deployWorkspaceProject/getDefaultValues/DefaultResourcesNameStep.ts
```diff
@@ -0,0 +1,139 @@
...
+            (!context.containerApp && !context.newContainerAppName);
+    }
+
+    private validateInput(name: string | undefined): string | undefined {
+        name ??= '';
+
+        // No symbols are allowed for ACR - we will strip out any offending characters from the base name, but still need to ensure this version has an appropriate length
+        const nameWithoutSymbols: string = name.replace(/[^a-z0-9]+/g, '');
+        if (!validateUtils.isValidLength(nameWithoutSymbols, 5, 20)) {
+            return localize('invalidLength', 'The alphanumeric portion of the name should total to at least 5 characters while not exceeding 20 characters.');
```
MicroFish91, line 73: Feel free to suggest better wording for this**

nturinski: 'The alphanumeric portion of the name must be least 5 characters but no more than 20 characters.'

src/commands/deployWorkspaceProject/getDefaultValues/DefaultResourcesNameStep.ts
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    protected async validateNameAvailability(context: DeployWorkspaceProjectContext, name: string): Promise<string | undefined> {
+        return await window.withProgress({
+            location: ProgressLocation.Notification,
+            cancellable: false,
+            title: localize('verifyingAvailabilityTitle', 'Verifying resource name availability...')
+        }, async () => {
+            const resourceGroupAvailable: boolean = await ResourceGroupListStep.isNameAvailable(context, name);
+            const managedEnvironmentAvailable: boolean = await ManagedEnvironmentNameStep.isNameAvailable(context, name, name);
```
nturinski, line 91: I think that this could be optimized quite a bit. I'm a little concerned about performance since it's 4 async calls.

1. If the resourceGroupAvailable is true, then managedEnviornment and containerApp must be available as well.
  > Why? Because both of these resources only need to be unique per RG. If the RG name is available, we are able to infer that ME and CA will also be available since it's going to be a new RG.
2. If managedEnvironment is available, since story for containerApp. Similar to the logic above.
3. I assume the implementation right now is that we are not re-using an existing RG, right? If that's the case, then we can actually stop at `isResourceGroupNameAvailable` because, assuming we're creating a new one for them regardless, then points 1 and 2 are always true.
4. We will always have to check registry since it's agnostic to RG. I think it has to be globally unique.

MicroFish91: That makes sense, I agree with all of your points.  The only thing is for point 3, we do re-use resource groups in cases where we choose the most-used managed environment to house the container app.

nturinski: Ah, then in that case, I think 1 and 2 are still true, but disregard 3.

MicroFish91: Since we talked about removing the auto selection logic for managed environment, point 3 is now correct!

MicroFish91: So I know we went back and talked about removing the auto select on managed environments, so point 3 would technically be correct again since the prompt step would skip altogether if re-deploying and all resources are already there.  

However, I did decide to keep the logic generally similar (mainly because I'm lazy but also...) in case we ever decide we want to be able to handle partial resource availability.  It will be a little bit over-engineered but will be more re-usable/future-proof.  I didn't feel like going back and changing all the logic since it still technically works for this scenario.

nturinski (approved):

src/commands/deployWorkspaceProject/getDefaultValues/DefaultResourcesNameStep.ts
nturinski: I'm not really sure how you'd ever have a case where you have a new environment but a invalid container name, but that's fine. I'm sure you have thought about it longer than me

MicroFish91: Nah, that was an oversight, thank you for catching that 😂

PR #449, opened 2023-09-11
Add `Dockerfile` port detection to ingress configuration logic

MicroFish91:
Related to: #425 

Added a pre-configuration path for ingress when a Dockerfile is detected + tests


src/commands/ingress/tryConfigureIngressUsingDockerfile.ts
```diff
@@ -0,0 +1,100 @@
...
+    if (!context.dockerfileExposePorts) {
+        context.enableIngress = false;
+        context.enableExternal = false;
+    } else if (context.dockerfileExposePorts) {
+        context.enableIngress = true;
+        context.enableExternal = true;
+        context.targetPort = getDefaultPort(context);
+    }
+
+    // If a container app already exists, activity children will be added automatically in later execute steps
```
MicroFish91, line 32: Will add this back in once we add all the activity children support in upcoming PR

src/commands/ingress/tryConfigureIngressUsingDockerfile.ts
```diff
@@ -0,0 +1,100 @@
...
+    if (!await AzExtFsExtra.pathExists(dockerfilePath)) {
+        return undefined;
+    }
+
+    const content: string = await AzExtFsExtra.readFile(dockerfilePath);
+    const lines: string[] = content.split('\n');
+
+    const portRanges: PortRange[] = [];
+    for (const line of lines) {
+        if (/^EXPOSE/i.test(line.trim())) {
```
bwateratmsft, line 83: Dockerfiles aren't too tough to parse, but it may be easier to use an existing parser. The Dockerfile language service used by the Docker extension has one. https://www.npmjs.com/package/dockerfile-ast

MicroFish91: Hmm, I did look into this one before and I ended up feeling unsure if it removed enough complexity to be worth it since my use case is relatively small here.

(1) I still have to loop through each line similar to what I'm already doing and search for an `EXPOSE` keyword 

(2) The args picked up are still raw strings that I have to manually loop through to identify ranges `-` and slashes `/` so that I can convert them into useable port numbers.

I think the one main benefit is that it will reduce the complexity of the `RegExp` that I'm using, but the overall process doesn't seem to get that much simpler.  Do you feel it may still be beneficial to use that dep?

bwateratmsft: That's fair, I think it's fine to manually parse, just make sure you know all the [variations](https://docs.docker.com/engine/reference/builder/#expose) that an EXPOSE statement can have.

src/commands/ingress/tryConfigureIngressUsingDockerfile.ts
```diff
@@ -0,0 +1,102 @@
...
+    get end(): number {
+        return this._end;
+    }
+
+    includes(port: number): boolean {
+        return port >= this.start && port <= this.end;
+    }
+}
+
+export async function getDockerfileExposePorts(dockerfilePath: string): Promise<PortRange[] | undefined> {
```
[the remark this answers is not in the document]
nturinski: Also feels like that this should just be in its own file.

src/commands/ingress/tryConfigureIngressUsingDockerfile.ts
```diff
@@ -0,0 +1,105 @@
...
+import { getDefaultPort } from './editTargetPort/getDefaultPort';
+
+export async function tryConfigureIngressUsingDockerfile(context: IngressContext): Promise<void> {
+    if (!context.dockerfilePath) {
+        return;
+    }
+
+    context.dockerfileExposePorts = await getDockerfileExposePorts(context.dockerfilePath);
+
+    if (context.alwaysPromptIngress) {
```
nturinski, line 19: Small nit, but should probably be at the top of the file so we don't waste any time doing the async call above.

test/ingress/dockerfileSamples/sample7.Dockerfile
```diff
@@ -0,0 +1,7 @@
+## Modified example from Azure-Samples/acr-build-helloworld-node
+FROM node:15-alpine
```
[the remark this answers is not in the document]
MicroFish91: That's fair, I literally just copy pasted everything from the original azure-sample and just changed the expose section

test/ingress/MockIngressContext.ts
```diff
@@ -0,0 +1,19 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import type { PortRange } from "../../extension.bundle";
+
+export interface MockIngressContext {
```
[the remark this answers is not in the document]
MicroFish91: Yeah, that's a big part of the reason

test/ingress/tryConfigureIngressUsingDockerfile.test.ts
```diff
@@ -0,0 +1,72 @@
...
+            { enableIngress: undefined, enableExternal: undefined, dockerfileExposePorts: expectedSamplePorts[3], targetPort: undefined }, // alwaysPromptIngress
+            { enableIngress: true, enableExternal: true, dockerfileExposePorts: expectedSamplePorts[4], targetPort: 443 },
+            { enableIngress: true, enableExternal: true, dockerfileExposePorts: expectedSamplePorts[5], targetPort: 80 },
+            { enableIngress: false, enableExternal: false, dockerfileExposePorts: undefined, targetPort: undefined }, // no expose
+        ];
+
+        for (const [i, ds] of dockerfileSamples.entries()) {
+            const context: MockIngressContext = {
+                dockerfilePath: i === 1 ? undefined : ds.fsPath,
+                alwaysPromptIngress: i === 3
```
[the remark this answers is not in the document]
MicroFish91: I have it in the expected results array at the top!

src/commands/ingress/tryConfigureIngressUsingDockerfile.ts
MicroFish91: This position is intentional, because we want to get the expose ports still so we can provide port suggestions later down the line

test/ingress/tryConfigureIngressUsingDockerfile.test.ts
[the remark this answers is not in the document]
MicroFish91: Let me add the `=true` modifier

nturinski (approved):

PR #450, opened 2023-09-15
Updates related to `EnvironmentVariablesListStep` 

MicroFish91:
Continuation of: #425 

- Consolidate two steps into one step:
image

- Add/use static method for checking if workspace has an env file


nturinski (commented):
Somewhat related to this PR-- do you know if you hit "skip for now", does the container app keep its existing environment settings or does that get reset?

MicroFish91:
I think currently it would just have no environment variables.

nturinski:
Yeah, this came up in a service connector meeting. Deploying to a container app and resetting all the environment settings is a pretty bad experience because it'd break a lot of things that were previously working.  I don't know if we should have another option that says "use existing settings" or, if you just skip it, it preserves whatever settings were on the app.

src/commands/deployImage/imageSource/EnvironmentVariablesListStep.ts
```diff
@@ -4,32 +4,93 @@
...
+
+    public static async workspaceHasEnvFile(): Promise<boolean> {
+        const envFileUris: Uri[] = await workspace.findFiles(allEnvFilesGlobPattern);
+        return !!envFileUris.length;
+    }
+
+    // Todo: It might be nice to add a direct command to update just the environment variables rather than having to suggest to re-run the entire command again
+    private outputLogs(_context: ImageSourceBaseContext, setEnvironmentVariableOption: SetEnvironmentVariableOption): void {
+        if (setEnvironmentVariableOption !== SetEnvironmentVariableOption.ProvideFile) {
+            // context.activityChildren?.push(
```
nturinski, line 66: Why did you comment out of the activity log stuff? Not hooked up yet?

nturinski (approved):

src/commands/deployImage/imageSource/EnvironmentVariablesListStep.ts
MicroFish91: Yup, going to hook up all the activity children logic in one PR later down the line because it's a bit more involved than just these item pushes... figured it's easier to review together in one wombo combo

MicroFish91:
> Yeah, this came up in a service connector meeting. Deploying to a container app and resetting all the environment settings is a pretty bad experience because it'd break a lot of things that were previously working. I don't know if we should have another option that says "use existing settings" or, if you just skip it, it preserves whatever settings were on the app.

That makes sense, I'll open an issue

PR #452, opened 2023-09-18
Default context logic for `deployWorkspaceProject`

MicroFish91:
Related to #425.

This PR primarily involves work to wrap up the `getDefaultContextValues` portion of the `deployWorkspaceProject` command.  The supporting work involved the following changes:

- Add three new ACA/VS Code deploy settings (`.vscode/settings.json`).  We will be using these as re-deploy settings.
- Refactor `settingUtils` (made some additional changes since I was already making changes to the file)
- Refactor some of the `createAcr` steps so that they play well with the express deploy process (mostly updates to `AcrListStep`)

General behavior for the defaulting logic is as follows:
1) Check the workspace settings for existing resources to deploy to, if they exist, leverage them
2) If resources do not exist or settings don't exist, leave the context values undefined and later we will assign new resource names using `DefaultResourcesNameStep` (which was already added in the second of the PRs).

Note: This PR also has the side effect of impacting the `Update Container App Image...` deploy path because they both leverage `AcrListStep`.  It now includes a `+ Create` quick pick when required to build an image using an ACR.

image


src/utils/settingUtils.ts
```diff
@@ -33,16 +51,49 @@ export namespace settingUtils {
...
+    }
+
+    /**
+     * Iteratively retrieves one of the user's workspace settings - sequentially checking for a defined value starting from the `WorkspaceFolder` up to the provided target configuration limit.
+     * @param key The key of the setting to retrieve
+     * @param fsPath The optional path of the workspace configuration settings
+     * @param targetLimit The optional target configuration limit (inclusive). Uses the `Workspace` configuration target unless otherwise specified
+     * @param prefix The optional extension prefix. Uses ext.prefix `containerApps` unless otherwise specified
+     */
+    export function getWorkspaceSettingIteratively<T>(
```
MicroFish91, line 78: I thought it would be nice to be able to differentiate the behavior of picking a configuration target value `directly` vs. using VS Code's `get` command to `iteratively` search for a settings value from most to least specific.

MicroFish91:
Closing in favor of smaller PRs

PR #453, opened 2023-09-19
Update the `RevisionDraftItem` icon to be more easily visible

MicroFish91:
We don't use the purple version of the icon anywhere else anymore, so we can probably use it as the default draft icon now.  It's a lot easier to see and looks better IMO.

Before: 

image

After:

image


alexweininger (approved):
I like

PR #454, opened 2023-09-19
Enable new workspace re-deploy settings for `deployWorkspaceProject`

MicroFish91:
Related to #425 

Another partial PR for `deployWorkspaceProject`.

Added support for 3 new (re-)deploy settings for `deployWorkspaceProject`.  We use these three settings to check for existing resources to add to the context before executing the deployment.

The default resources affected by this PR are: container app, managed environment, resource group, ACR.

Logic goes:
1) Check if workspace settings exist
2) If yes, check that Azure actually has those resources available, and if so, grab them for re-use
3) If no to any of the above, set/keep as `undefined` and allow the `DefaultResourcesNameStep` (added in previous PR) to name the resources for later creation

The workflow in general has been demo'd to Anthony and Misty including the showing of a re-deploy scenario with these same settings.  Feel free to suggest any improvements as necessary though.

What the full thing will eventually look like: [demo](https://microsoft-my.sharepoint.com/:v:/p/matfisher/EXziurykRO9Bhmg_Yp3Nq0oBx02_-ZC2CNzM8EWf_gEF5Q?nav=eyJyZWZlcnJhbEluZm8iOnsicmVmZXJyYWxBcHAiOiJPbmVEcml2ZUZvckJ1c2luZXNzIiwicmVmZXJyYWxBcHBQbGF0Zm9ybSI6IldlYiIsInJlZmVycmFsTW9kZSI6InZpZXciLCJyZWZlcnJhbFZpZXciOiJNeUZpbGVzTGlua0RpcmVjdCJ9fQ&e=NVcKcB&xsdata=MDV8MDF8fDQ1ZGViOTg0MjIzZTQ1MGE3NDE2MDhkYmFhNmZjMDcxfDcyZjk4OGJmODZmMTQxYWY5MWFiMmQ3Y2QwMTFkYjQ3fDB8MHw2MzgyOTExNjkxNDk4MDk2ODZ8VW5rbm93bnxWR1ZoYlhOVFpXTjFjbWwwZVZObGNuWnBZMlY4ZXlKV0lqb2lNQzR3TGpBd01EQWlMQ0pRSWpvaVYybHVNeklpTENKQlRpSTZJazkwYUdWeUlpd2lWMVFpT2pFeGZRPT18MXxMMk5vWVhSekx6RTVPbTFsWlhScGJtZGZUa2RhYVZsVVNYZFpNbFYwVFhwa2FFNURNREJQVkVGM1RGUm9hVmxYV1hST2VtaHJUVVJhYlZwVWJHbE5SRlpxUUhSb2NtVmhaQzUyTWk5dFpYTnpZV2RsY3k4eE5qa3pOVEU1TWpBMU9UYzJ8ZmUzZDBjMGM4MTgxNDliOTc0MTYwOGRiYWE2ZmMwNzF8MmNhYWI0Y2RhOGFiNDA4YmIyMzYwMGYzN2FkZmM2MGU%3D&sdata=YzZZMFNnT3c2ZE5RTDlmbFBJbW5JYWZGR2JMMW51SUc3T1BHKzFEbTJwST0%3D&ovuser=72f988bf-86f1-41af-91ab-2d7cd011db47%2Cmatfisher%40microsoft.com)


src/commands/deployWorkspaceProject/DeployWorkspaceProjectSettings.ts
```diff
@@ -0,0 +1,37 @@
...
+        const containerRegistryName: string | undefined = settingUtils.getWorkspaceSetting(`${deployWorkspaceProjectPrefix}.containerRegistryName`, settingsPath);
+
+        if (containerAppName || containerAppResourceGroupName || containerRegistryName) {
+            return {
+                containerAppName,
+                containerAppResourceGroupName,
+                containerRegistryName
+            };
+        }
+    } catch { /** Do nothing */ }
```
alexweininger, line 34: Might be useful to add a comment here that explains what errors are being ignored here and why

src/commands/deployWorkspaceProject/getDefaultValues/getDefaultAcrResources.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    if (!settings || !settings.containerRegistryName) {
+        return noMatchingResource;
+    }
+
+    const registries: Registry[] = await AcrListStep.getRegistries(context);
+    const savedRegistry: Registry | undefined = registries.find(r => r.name === settings.containerRegistryName);
+
+    if (savedRegistry) {
+        ext.outputChannel.appendLog(localize('foundResourceMatch', 'Used saved workspace settings and found an existing container registry.'));
```
alexweininger, line 29: Big fan of the logs in this function. You could include the found resource names in the log if you wanted to be even more verbose. But just an idea.

src/commands/deployWorkspaceProject/DeployWorkspaceProjectSettings.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export async function getDeployWorkspaceProjectSettings(rootFolder: WorkspaceFolder): Promise<DeployWorkspaceProjectSettings | undefined> {
+    const settingsPath: string = settingUtils.getDefaultRootWorkspaceSettingsPath(rootFolder);
+
+    try {
+        const containerAppName: string | undefined = settingUtils.getWorkspaceSetting(`${deployWorkspaceProjectPrefix}.containerAppName`, settingsPath);
+        const containerAppResourceGroupName: string | undefined = settingUtils.getWorkspaceSetting(`${deployWorkspaceProjectPrefix}.containerAppResourceGroupName`, settingsPath);
+        const containerRegistryName: string | undefined = settingUtils.getWorkspaceSetting(`${deployWorkspaceProjectPrefix}.containerRegistryName`, settingsPath);
+
+        if (containerAppName || containerAppResourceGroupName || containerRegistryName) {
```
alexweininger, line 27: This isn't a blocker, but I think it's awkward to only return an object if one of the properties is defined. And otherwise we return undefined.

In the future, if you add more settings, then you need to add them to this if expression.

I think I'd prefer if you just always return an object, even if the object just contains undefined values.

MicroFish91: Yeah I agree on it feeling a little awkward; the settings section felt a little clunky to me but I wasn't sure how to get around it without just changing the log outputs we display.

I kept that `undefined` path so that I could easily use this statement:
image

Though I've gone back and forth on whether it's even worth expressing that level of output detail to the user or not.  If I keep the statement and return an empty object, I'd just be moving that verbose `if` check up one level.  Is there a better way to handle this/what do you think?

src/commands/deployWorkspaceProject/getDefaultValues/getDefaultAcrResources.ts
MicroFish91: Yeah, actually this suggestion will be added in the following PR!

src/commands/deployWorkspaceProject/DeployWorkspaceProjectSettings.ts
MicroFish91: Yeah I agree, I'll have to think a little more about it, but the reason I was returning `undefined` like that was because it made it a lot easier to just check `if(!settings)` later.  I think if I remove the statement in this file, I'd then have to add a similar verbose check up one level to implement that logic path.  I'm wondering if there's a more clever way we can omit the verbose check entirely that I haven't though of yet.

src/commands/deployWorkspaceProject/DeployWorkspaceProjectSettings.ts
alexweininger: Ok now that I see this again, you should probably log something inside that catch block too.

MicroFish91: This was a good suggestions, because it actually pushed me to test the get setting method a little bit more, and I realized that it doesn't actually throw an error even with bad settings path and property... so I removed the catch entirely.  I also made a small change to the setting utils so that it reads a missing value as undefined rather than as an empty string `''`.

alexweininger: Ok that's great

alexweininger (approved):

PR #455, opened 2023-09-19
Update `settingUtils`

MicroFish91:


alexweininger (approved):
These probably belong in utils, but we can move them later.

PR #458, opened 2023-09-21
Enable the remaining context defaulting logic for `deployWorkspaceProject`

### LINKED ISSUE #78: There is no command "+ Create new registry..." in the command list when asking to select an azure container registry
**OS:** All
**Build Version:** 20220405.6

**Repro Steps:**

1. Right click a container app environment -> Create Container App -> Type a name -> Select "Azure Container Registries".
2. Check whether there is a command "+ Create new registry..." in the command list.

**Expect:**
There is a command "+ Create new registry..." in the command list.

**Actual:**
There is no command "+ Create new registry..." in the command list.
Screen Shot 2022-04-06 at 18 36 21

MicroFish91:
Closes #78 
Related to #425 

Involves a small refactor of the existing ACR steps so that it plays nicely with express deploy (mostly some updates to `AcrListStep`).

Note: This PR also has the side effect of impacting the `Update Container App Image...` path because they both leverage `AcrListStep`.  It now includes a `+ Create` quick pick when building an image remotely using ACR.

image


src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/RegistryNameStep.ts
```diff
@@ -29,14 +29,14 @@ export class RegistryNameStep extends AzureWizardPromptStep<CreateAcrContext> {
...
-            return localize('validateInputError', `Connection names can only consist of alphanumeric characters.`);
+            return localize('validateInputError', `The name must consist of alphanumeric characters.`);
         }
 
         return undefined;
     }
 
     private async validateNameAvalability(context: CreateAcrContext, name: string) {
-        if (await RegistryNameStep.isNameAvailable(context, name)) {
+        if (!await RegistryNameStep.isNameAvailable(context, name)) {
```
MicroFish91, line 39: Fixing a small error here that I accidentally introduced when I refactored to use a static method.

src/commands/deployImage/imageSource/containerRegistry/acr/AcrListStep.ts
```diff
@@ -69,3 +112,22 @@ export class AcrListStep extends AzureWizardPromptStep<IContainerRegistryImageCo
...
+    promptSteps: AzureWizardPromptStep<IContainerRegistryImageContext>[],
+): Promise<void> {
+    // No need to pollute the base context with all the potential pre-create typings as they are not otherwise used
+    const resourceCreationContext = context as Partial<ICreateContainerAppContext> & Partial<IManagedEnvironmentContext> & CreateAcrContext;
+    if (resourceCreationContext.resourceGroup || resourceCreationContext.newResourceGroupName) {
+        return;
+    }
+
+    // Try to leverage an existing resource group (managed environment name is a fallback here because our create flow has the resource group name mirror the environment name)
+    const resourceGroupName: string | undefined = resourceCreationContext.containerApp?.resourceGroup || resourceCreationContext.managedEnvironment?.name;
```
MicroFish91, line 126: I've been going back and forth on whether using the managed environment name as a search fallback makes sense or not, let me know if you have thoughts

src/commands/deployImage/imageSource/containerRegistry/acr/AcrListStep.ts
```diff
@@ -69,3 +112,22 @@ export class AcrListStep extends AzureWizardPromptStep<IContainerRegistryImageCo
...
+    context: IContainerRegistryImageContext,
+    promptSteps: AzureWizardPromptStep<IContainerRegistryImageContext>[],
+): Promise<void> {
+    // No need to pollute the base context with all the potential pre-create typings as they are not otherwise used
+    const resourceCreationContext = context as Partial<ICreateContainerAppContext> & Partial<IManagedEnvironmentContext> & CreateAcrContext;
+    if (resourceCreationContext.resourceGroup || resourceCreationContext.newResourceGroupName) {
+        return;
+    }
+
+    // Try to leverage an existing resource group (managed environment name is a fallback here because our create flow has the resource group name mirror the environment name)
```
nturinski, line 125: Shouldn't you check what resource group the managed environment is in rather than using its name?

src/commands/deployImage/imageSource/containerRegistry/acr/RegistryEnableAdminUserStep.ts
```diff
@@ -30,7 +29,6 @@ export class RegistryEnableAdminUserStep extends AzureWizardPromptStep<IContaine
     }
 
     public shouldPrompt(context: IContainerRegistryImageContext): boolean {
-        // this is only a requirement for ACR
-        return context.registryDomain === acrDomain && !!context.registry && !context.registry.adminUserEnabled;
+        return !!context.registry && !context.registry.adminUserEnabled;
```
nturinski, line 32: Is checking for an acrDomain not a requirement anymore? Or do we know for a fact that this step is only called for acrs?

src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/RegistryNameStep.ts
```diff
@@ -29,14 +29,14 @@ export class RegistryNameStep extends AzureWizardPromptStep<CreateAcrContext> {
         if (name.length < minLength || name.length > maxLength) {
             return localize('validationLengthError', 'The name must be between {0} and {1} characters.', minLength, maxLength);
         } else if (!/^[a-z][a-zA-Z0-9]*$/.test(name)) {
-            return localize('validateInputError', `Connection names can only consist of alphanumeric characters.`);
+            return localize('validateInputError', `The name must consist of alphanumeric characters.`);
```
nturinski, line 32: If we're going to make the error more generic anyway, may as well just use your validation utils here.

src/commands/deployImage/imageSource/containerRegistry/acr/RegistryEnableAdminUserStep.ts
MicroFish91: My thought was that `context.registry` has a type definition tied to being an ACR already and so we probably don't need to check for the `acrDomain` if the registry is populated 🤔

src/commands/deployImage/imageSource/containerRegistry/acr/createAcr/RegistryNameStep.ts
MicroFish91: I agree, I just need to add the upper-case option because right now it only supports lower-case alphanumeric.  I figured I'd do that in a separate PR, maybe even after release.  I just wanted to change the message because I thought `Connection` might have been a typo

src/commands/deployImage/imageSource/containerRegistry/acr/AcrListStep.ts
MicroFish91: Yes, this would be the wiser way to do it.  My brain must not have been working properly when I wrote this part 😆

PR #461, opened 2023-09-26
Add support for deleting scale rules 

motm32:
Contributes to #420


src/commands/scaling/deleteScaleRule/ScaleRuleListStep.ts
```diff
@@ -0,0 +1,34 @@
...
+        context.scaleRule = (await context.ui.showQuickPick(this.getPicks(context), { placeHolder: localize('scaleRule', 'Select a Scale Rule') })).data;
+    }
+
+    public shouldPrompt(context: IDeleteScaleRuleContext): boolean {
+        return !context.scaleRule;
+    }
+
+    private async getPicks(context: IDeleteScaleRuleContext): Promise<IAzureQuickPickItem<ScaleRule>[] | undefined> {
+        const scaleRules = context.containerApp?.template?.scale?.rules;
+        if (scaleRules === undefined) {
```
MicroFish91, line 22: An alternative could be something like `if (!scaleRules?.length)`.  

This has the benefit of executing both when scaleRules is `undefined` and when `scaleRules` has a length of `0`.

src/commands/scaling/deleteScaleRule/DeleteScaleRuleStep.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    constructor(baseItem: RevisionsItemModel) {
+        super(baseItem);
+    }
+
+    public async execute(context: IDeleteScaleRuleContext): Promise<void> {
+        this.revisionDraftTemplate.scale ||= {};
+        this.revisionDraftTemplate.scale.rules ||= [];
+
+        const index = this.revisionDraftTemplate.scale.rules.indexOf(nonNullProp(context, 'scaleRule'));
```
MicroFish91, line 25: The idea here is spot on, but due to how objects are compared with Node (by reference/address), this will probably always return `-1`.  

To circumvent this, a likely better way to match would be to compare one of their properties instead.  Something like this could work:
`const index = this.revisionDraftTemplate.scale.rules.findIndex(r => r.name === nonNullValueAndProp(context.scaleRule, 'name'));`

I would also maybe add an if statement guard clause immediately after for the case when index is `-1` (no match) and handle it by throwing an error or something 🤔

src/commands/scaling/deleteScaleRule/ScaleRuleListStep.ts
```diff
@@ -0,0 +1,34 @@
...
+*--------------------------------------------------------------------------------------------*/
+
+import { ScaleRule } from "@azure/arm-appcontainers";
+import { AzureWizardPromptStep, IAzureQuickPickItem, nonNullProp, nonNullValue } from "@microsoft/vscode-azext-utils";
+import { localize } from "../../../utils/localize";
+import { IDeleteScaleRuleContext } from "./IDeleteScaleRuleContext";
+
+export class ScaleRuleListStep extends AzureWizardPromptStep<IDeleteScaleRuleContext> {
+    public async prompt(context: IDeleteScaleRuleContext): Promise<void> {
+        context.scaleRule = (await context.ui.showQuickPick(this.getPicks(context), { placeHolder: localize('scaleRule', 'Select a Scale Rule') })).data;
```
MicroFish91, line 13:
```suggestion
        context.scaleRule = (await context.ui.showQuickPick(this.getPicks(context), { placeHolder: localize('scaleRule', 'Select a scale rule') })).data;
```

MicroFish91:
Note: Since Megan is OOF I decided to make just add a commit with the above changes and will merge.  I also added a new scale rules folder and created a shared context for scale rule commands to use.

MicroFish91 (approved):

PR #463, opened 2023-09-26
Clean up `Ingress` tree refreshing and activity title behaviors

MicroFish91:
Since some of these steps are re-usable, it doesn't make sense to make activity title and refresh changes in the execute steps themselves.  I wish I had realized this earlier, but I pulled these behaviors out and attached them to the commands instead and will implement this way going forward.


alexweininger (approved):
Just make sure the capitalization of "container app" is consistent

MicroFish91:
Okay, looks good I think; have them all consistently lower case unless we use `Azure` as a prefix (which we don't in these cases)

PR #471, opened 2023-10-04
Adds logic to display `noMatchingResources` for multiple container registry steps

### LINKED ISSUE #444: There is no prompt "No matching resources found" when selecting an empty Azure Container Registry
**OS:** Win10
**Build Version:** 20230903.1

**Repro Steps:**
1. F1 -> Execute "Azure Container Apps: Create Azure Container Registry..." command to create an Azure Container Registry.
2. Right click one Azure Container Environment -> Select "Create Container App..." -> Enter a name for the new container app -> Select "Use image from registry" -> Select "Azure Container Registry" -> Select the above created Azure Container Registry.
3. Check whether there is a prompt "No matching resources found" or not.

**Expect:**
There is a prompt "No matching resources found".

**Actual:**
There is no prompt "No matching resources found".
image

MicroFish91:
Fixes #444 

Adds logic to display `noMatchingResources` to the following steps:
- `AcrListStep`
- `AcrRepositoriesListStep`
- `DockerHubContainerRepositoryListStep`

Example: 
image

### DIFF
--- a/src/commands/deployImage/imageSource/containerRegistry/acr/AcrRepositoriesListStep.ts
+++ b/src/commands/deployImage/imageSource/containerRegistry/acr/AcrRepositoriesListStep.ts
@@ -6,7 +6,7 @@
 import { uiUtils } from "@microsoft/vscode-azext-azureutils";
 import { nonNullValue } from "@microsoft/vscode-azext-utils";
 import type { QuickPickItem } from "vscode";
-import { acrDomain, currentlyDeployed, noMatchingResourceQp, quickStartImageName } from "../../../../../constants";
+import { acrDomain, currentlyDeployed, noMatchingResourcesQp, quickStartImageName } from "../../../../../constants";
 import { createContainerRegistryClient } from "../../../../../utils/azureClients";
 import { parseImageName } from "../../../../../utils/imageNameUtils";
 import type { IContainerRegistryImageContext } from "../IContainerRegistryImageContext";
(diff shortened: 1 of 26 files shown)


src/constants.ts
```diff
@@ -69,6 +69,14 @@ export type SupportedRegistries = 'azurecr.io' | 'docker.io';
 
 export const browseItem: IAzureQuickPickItem<undefined> = { label: `$(file-directory) ${localize('browse', 'Browse...')}`, description: '', data: undefined };
 export const loadMoreQp: IAzureQuickPickItem = { label: '$(sync) Load More', data: undefined, suppressPersistence: true };
+
+export const noMatchingResources = 'noMatchingResources';
+export const noMatchingResourcesQp: IAzureQuickPickItem<typeof noMatchingResources> = {
```
nturinski, line 74: Makes me feel like this behavior should be implemented in the `ui.showQuickPick` command instead. If it has 0 picks, we could just show this.

src/commands/deployImage/imageSource/containerRegistry/acr/AcrRepositoriesListStep.ts
```diff
@@ -37,11 +38,18 @@ export class AcrRepositoriesListStep extends RegistryRepositoriesListStepBase {
...
+            picks.push(noMatchingResourcesQp);
+        }
+
         // Preferring 'suppressPersistence: true' over 'priority: highest' to avoid the possibility of a double parenthesis appearing in the description
-        return repositoryNames.map((rn) => {
-            return !!suggestedRepository && rn === suggestedRepository ?
-                { label: rn, description: currentlyDeployed, suppressPersistence: true } :
-                { label: rn, suppressPersistence: srExists };
-        });
+        return picks.concat(
```
nturinski, line 47: Small thing, but you shouldn't need this concat here (since the picks array should be empty. I think you could probably just return an array with `[noMatchingResourcesQp]` on line 43 and leave the rest of the code unchanged.

nturinski (approved):

src/constants.ts
MicroFish91: Yeah I agree, we should probably revisit implementing this as part of the wrapper in tools later

PR #472, opened 2023-10-04
Change non-`Azure` resource titles to lower case

MicroFish91:


alexweininger (approved):
Looks good. I'm curious, are we following some other place for doing this? Like is it this way on the portal?

MicroFish91:
Not specifically doing it based off of the portal or anything, just thought this format looked better when displayed in the activity log so I started doing it this way a while back.  My understanding was that we were only required to capitalize these when we were using the `Azure` prefix for branding.  Let me know if you prefer it the other way though!

PR #476, opened 2023-10-08
Refactor / consolidate `deployImage` & `deployImageApi` commands

MicroFish91:
**Part I** 

In preparation for adding support for the new update image command (revision draft command), I am removing the old entry point for `deployImage` as it will be replaced.

I have consolidated the existing `deployImageApi` under the new `image` directory where the image-related commands will now sit (this is the same command that is shared with the docker extension).  Its general implementation/flow should remain unchanged.  

The container update commands and steps have been consolidated under a new utils folder called `updateContainerApp`.


src/commands/image/imageSource/containerRegistry/ContainerRegistryImageConfigureStep.ts
```diff
@@ -41,6 +41,10 @@ export class ContainerRegistryImageConfigureStep extends AzureWizardExecuteStep<
             }
         }
 
+        // Preserve existing secrets/registries even if new ones haven't been added
+        context.secrets ??= secrets;
+        context.registries ??= registries;
```
MicroFish91, line 46: This is really covering the else condition above (so third party public Docker registries), but it's also just a general good fallback to have here.

src/utils/updateContainerApp/updateContainerApp.ts
```diff
@@ -7,7 +7,7 @@ import { ContainerApp, ContainerAppsAPIClient } from "@azure/arm-appcontainers";
 import { getResourceGroupFromId } from "@microsoft/vscode-azext-azureutils";
 import { IActionContext, createSubscriptionContext, nonNullProp } from "@microsoft/vscode-azext-utils";
 import { AzureSubscription } from "@microsoft/vscode-azureresources-api";
-import { createContainerAppsAPIClient } from "./azureClients";
+import { createContainerAppsAPIClient } from "../azureClients";
 
 export async function updateContainerApp(context: IActionContext, subscription: AzureSubscription, containerApp: ContainerApp, updatedSetting?: Omit<ContainerApp, 'location'>): Promise<void> {
```
nturinski, line 12: I know that this isn't technically a command, but it still seems strange that this is in the utils folder with other files in command like pattern. I would prefer we just move it somewhere with the commands since everything that is using it is in commands anyway.

src/utils/updateContainerApp/ContainerAppUpdateStep.ts
```diff
@@ -4,21 +4,21 @@
...
-import { ExecuteActivityOutput, ExecuteActivityOutputStepBase } from "../../utils/activity/ExecuteActivityOutputStepBase";
-import { createActivityChildContext } from "../../utils/activity/activityUtils";
-import { localize } from "../../utils/localize";
-import { updateContainerApp } from "../../utils/updateContainerApp";
-import type { IDeployImageContext } from "./deployImage";
-import { getContainerNameForImage } from "./imageSource/containerRegistry/getContainerNameForImage";
+import { ExecuteActivityOutput, ExecuteActivityOutputStepBase } from "../activity/ExecuteActivityOutputStepBase";
+import { createActivityChildContext } from "../activity/activityUtils";
+import { localize } from "../localize";
+import { updateContainerApp } from "./updateContainerApp";
```
nturinski, line 16: Rather than importing this directly, shouldn't each wizard just include the execute step? It seems kind of weird to have the step, but then directly call it like this.

MicroFish91: Now that I'm looking at it more, `ContainerAppUpdateStep` probably belongs in the `imageSource` folder since it's pretty heavily tied to those context values.  The other two files in `updateContainerApp` I'm not sure of... kind of thinking maybe they should still sit in utils IDK yet

MicroFish91: Yeah, there's a bit of stuff to unpack here that I'll reach out to discuss with you.  Some topics to cover:
- Renaming `ContainerAppUpdateStep`
- Moving `ContainerAppUpdateStep` to `imageSouce/`
- Potentially keeping the renaming two files in the `utils` folder

MicroFish91: Summary of actions:
- Moved `ContainerAppUpdateStep` to `imageSource/`.
- Moved `updateContainerApp` and `ContainerAppOverwriteConfirmStep` to `commands/`

nturinski (approved):

PR #478, opened 2023-10-09
Add a conditional revision draft deploy pop-up

MicroFish91:
**Part 3**

Pretty much ready to review but getting final feedback from Anthony + ACA group.

This PR covers the pop-up that appears at the end of this [demo](https://microsoft-my.sharepoint.com/:v:/p/matfisher/EVIuBVnabXdOqQsjx-M90QMBjT9tLD_sEAeQ1YH9qheR3g?nav=eyJyZWZlcnJhbEluZm8iOnsicmVmZXJyYWxBcHAiOiJPbmVEcml2ZUZvckJ1c2luZXNzIiwicmVmZXJyYWxBcHBQbGF0Zm9ybSI6IldlYiIsInJlZmVycmFsTW9kZSI6InZpZXciLCJyZWZlcnJhbFZpZXciOiJNeUZpbGVzTGlua0RpcmVjdCJ9fQ&e=us1zFD)

Pending team feedback, this is what it currently looks like:
image


src/commands/image/updateImage/updateImage.ts
```diff
@@ -70,4 +70,6 @@ export async function updateImage(context: IActionContext, node?: ContainerAppIt
 
     await wizard.prompt();
     await wizard.execute();
+
+    void showRevisionDraftDeployPopup(context, containerApp);
```
nturinski, line 74: Wouldn't it be easier to just include this once in `updateContainerApp`? This should basically only be called wehenver we save the revision draft, right?

src/utils/revisionDraftUtils.ts
```diff
@@ -26,3 +34,43 @@ export function getParentResourceFromItem(item: ContainerAppItem | RevisionsItem
...
+export async function showRevisionDraftDeployPopup(context: IActionContext, containerApp: ContainerAppModel): Promise<void> {
+    if (!await settingUtils.getGlobalSetting('showDraftCommandDeployPopup')) {
+        return;
+    }
+
+    const yes: string = localize('yes', 'Yes');
+    const no: string = localize('no', 'No');
+    const dontShowAgain: string = localize('dontShowAgain', 'Don\'t show again');
+
+    const message: string = localize('message', 'Would you like to deploy these changes? Click "Yes" to proceed, or "No" to continue making changes.');
```
nturinski, line 50: I think we should definitely run through this with Anthony/Misty to finalize the wording.

src/utils/pickItem/pickContainerApp.ts
```diff
@@ -31,5 +43,23 @@ export async function pickContainerApp(context: IActionContext, options?: PickIt
     return await runQuickPickWizard(context, {
         promptSteps,
         title: options?.title,
+        showLoadingPrompt: options?.showLoadingPrompt
+    });
+}
+
+export async function pickContainerAppWithoutPrompt(
```
nturinski, line 50: Since you're passing in the containerApp name as the context, I imagine that this will always get the container app? I feel like it should just be called `getContainerApp`.

src/commands/image/updateImage/updateImage.ts
MicroFish91: That's a good idea.  I don't think we'd want it in `updateContainerApp` though because that's the one that does the deploying, but maybe the `RevisionDraftUpdateBaseStep` is a good candidate since it's always used for commands to write to the revision draft.

image

nturinski: Oh yeah, that's the one I mean. I thought `updateContainerApp` was the one that was just creating a draft, but yeah, you get what I mean.

MicroFish91: I also consolidated this logic in with the base step since it's really just used there now.

MicroFish91:
Should be good for a final review pass 👍

nturinski (approved):

nturinski:
We can get feedback on the strings after merging once we give them the vsix

PR #481, opened 2023-10-10
Add ability to choose an existing managed environment for `deployWorkspaceProject`

MicroFish91:
image

Some of this logic can probably be eventually re-purposed into a `ManagedEnvironmentListStep`

### DIFF
--- a/src/commands/deployWorkspaceProject/getDefaultValues/getDefaultContainerAppsResources.ts
+++ b/src/commands/deployWorkspaceProject/getDefaultValues/getDefaultContainerAppsResources.ts
@@ -5,13 +5,13 @@
 
 import { ContainerApp, ContainerAppsAPIClient, ManagedEnvironment } from "@azure/arm-appcontainers";
 import type { ResourceGroup } from "@azure/arm-resources";
-import { ResourceGroupListStep, uiUtils } from "@microsoft/vscode-azext-azureutils";
-import type { ISubscriptionActionContext } from "@microsoft/vscode-azext-utils";
+import { ResourceGroupListStep, getResourceGroupFromId, uiUtils } from "@microsoft/vscode-azext-azureutils";
+import { nonNullProp, type IAzureQuickPickItem, type ISubscriptionActionContext } from "@microsoft/vscode-azext-utils";
 import { ext } from "../../../extensionVariables";
 import { ContainerAppItem, ContainerAppModel } from "../../../tree/ContainerAppItem";
 import { createContainerAppsAPIClient } from "../../../utils/azureClients";
 import { localize } from "../../../utils/localize";
-import { DeployWorkspaceProjectSettings } from "../deployWorkspaceProjectSettings";
+import type { DeployWorkspaceProjectSettings } from "../deployWorkspaceProjectSettings";
 
 interface DefaultContainerAppsResources {
     resourceGroup?: ResourceGroup;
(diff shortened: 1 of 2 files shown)


src/commands/deployWorkspaceProject/getDefaultValues/getDefaultContainerAppsResources.ts
```diff
@@ -56,3 +67,38 @@ export async function getDefaultContainerAppsResources(context: ISubscriptionAct
         return noMatchingResources;
     }
 }
+
+async function promptForContainerAppsEnvironmentResources(context: ISubscriptionActionContext): Promise<DefaultContainerAppsResources> {
```
nturinski, line 71: Why not call it `promptForManagedEnvironmentResources`?

src/commands/deployWorkspaceProject/getDefaultValues/getDefaultContainerAppsResources.ts
```diff
@@ -56,3 +67,38 @@ export async function getDefaultContainerAppsResources(context: ISubscriptionAct
...
+    const client: ContainerAppsAPIClient = await createContainerAppsAPIClient(context)
+    const managedEnvironments: ManagedEnvironment[] = await uiUtils.listAllIterator(client.managedEnvironments.listBySubscription());
+
+    if (!managedEnvironments.length) {
+        return noMatchingResources;
```
nturinski, line 76: If there are no managed environments, shouldn't we still prompt with only `$(plus) Create new container apps environment` as the option?

src/commands/deployWorkspaceProject/getDefaultValues/getDefaultContainerAppsResources.ts
```diff
@@ -56,3 +67,38 @@ export async function getDefaultContainerAppsResources(context: ISubscriptionAct
...
+                description: '',
+                data: env
+            };
+        })
+    ];
+
+    const placeHolder: string = localize('selectManagedEnvironment', 'Select a container apps environment');
+    const managedEnvironment: ManagedEnvironment | undefined = (await context.ui.showQuickPick(picks, { placeHolder })).data;
+
+    if (!managedEnvironment) {
```
nturinski, line 97: Wouldn't `managedEnvironment` be undefined here if the user selected `create new container app environment`? I'm thoroughly confused on what is going on here.

Like you said in your PR, I feel like this logic should be in a `ManagedEnvironmentListStep` rather than having it be a part of the `getDefaultContainerAppsResources`.

My opinion is that this command should _only_ try to get the default values. If it can't, it exits out and we rely on the wizard prompts to get input from the user.

nturinski (commented):
I don't remember exactly how this implementation works, but I'm guessing that `getDefaultContainerAppsResources` is called somewhere and then we do some prompts or something based off of that?

Could you show me how it's used? There's been so many PRs, I'm not even sure if that part has been merged in yet.

But the way I envision it working is we call `getDefaultContainerAppsResources` and it will set the context values appropriately.  After that, we just call the wizards and the `ListSteps` will handle the logic of whether or not we should skip the steps.

src/commands/deployWorkspaceProject/getDefaultValues/getDefaultContainerAppsResources.ts
MicroFish91: Nah, just a reminder since it's probably been a while since the naming PR, but an `undefined` resource means that we were unable to acquire an existing resource to use.  Therefore, later, the undefined resource will indicate to the command that we need to create the missing resource, thus in this scenario, returning `undefined` is equivalent to the user selecting the create button.  Since there's only one option, we can just pick it for them and skip displaying the prompt.

src/commands/deployWorkspaceProject/getDefaultValues/getDefaultContainerAppsResources.ts
MicroFish91: Based on the comment above, and this comment, I think the confusion is due to thinking about this as if it were a list step being used as part of our Azure Wizard flow.  

> Wouldn't managedEnvironment be undefined here if the user selected create new container app environment?
In this scenario, `undefined`
Yes, exactly this, if the user selects create, by setting as `undefined`, we are indicating that no default resource exists and that one needs to be created later.

> Like you said in your PR, I feel like this logic should be in a ManagedEnvironmentListStep rather than having it be a part of the getDefaultContainerAppsResources.

I thought about making it all as a list step, but there's so much other stuff I would then have to build out and think about when all I really need is the `getPicks` portion of the list step essentially... so I decided to defer creating a list step until I actually need to use one.

MicroFish91:
> Could you show me how it's used? There's been so many PRs, I'm not even sure if that part has been merged in yet.

Yeah, come to think of it, I think some of the related sections were merged while you were out, so you may not have the full context with what you already reviewed.  Let me message you offline to discuss more.

src/commands/deployWorkspaceProject/getDefaultValues/getDefaultContainerAppsResources.ts
nturinski: I think I get how this is supposed to work now. I honestly really don't like this design pattern-- I find it's super confusing and not very congruent with how we would normally do it with wizard steps.

That being said, I understand that we've been kicking this can on the release for quite a while, so let's just get this in. We can always refactor into wizard steps later.

nturinski (approved):

src/commands/deployWorkspaceProject/getDefaultValues/getDefaultContainerAppsResources.ts
MicroFish91: Yeah, given the time restraints and the potential for side effects, I think it's just easier for now to go with the implementation like this.  After release we can add an engineering task to refactor this into a list step and throw it into our normal wizard flow.

PR #487, opened 2023-10-12
Fix command palette support for `restartRevision`

### LINKED ISSUE #486: Revision prompt doesn't show up when executing the "Restart Revision" command through the command palette
When a container app is in multi revision mode there is no prompt to choose a specific revision when trying to restart a revision from the command palette.

MicroFish91:
Fixes #486

### DIFF
--- a/src/commands/revision/changeRevisionActiveState.ts
+++ b/src/commands/revision/changeRevisionActiveState.ts
@@ -3,17 +3,26 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
-import { ContainerAppsAPIClient } from "@azure/arm-appcontainers";
-import { IActionContext, nonNullProp } from "@microsoft/vscode-azext-utils";
+import { ContainerAppsAPIClient, KnownActiveRevisionsMode } from "@azure/arm-appcontainers";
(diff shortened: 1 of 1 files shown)


src/commands/revision/changeRevisionActiveState.ts
```diff
@@ -3,17 +3,26 @@
...
 import type { ContainerAppItem } from "../../tree/ContainerAppItem";
 import { RevisionItem } from "../../tree/revisionManagement/RevisionItem";
 import { createContainerAppsClient } from "../../utils/azureClients";
 import { localize } from "../../utils/localize";
 import { pickContainerApp } from "../../utils/pickItem/pickContainerApp";
+import { pickRevision } from "../../utils/pickItem/pickRevision";
 
 export async function executeRevisionOperation(context: IActionContext, node: ContainerAppItem | RevisionItem | undefined, operation: RevisionOperation): Promise<void> {
-    const item = node ?? await pickContainerApp(context);
+    if (!node) {
```
alexweininger, line 17: What if the `node` is a `ContainerAppItem` but it's in multiple revision mode?

MicroFish91: I think we should be good with this scenario because the command is removed from the context menu when in multiple revisions mode.  When in multiple revisions mode the command should only be executable from either no node (command-palette) or a `RevisionItem`!

alexweininger (approved):

Issue #496, opened 2023-10-17
Display the scale rule list of the latest revision when deleting a revision but not a latest one


v-xinda:
**OS:** All
**Build Version:** 20231016.3
**Regression:** Not a Regression

**Repro Steps:**
1. Create a container app.
2. Make sure the revision mode is multiple.
3. Create scale rules.
4. F1 -> Click "Delete Scale Rule..." -> Select a container apps environment -> Select a container app -> Select the latest revision -> Select a scale rule -> Cancel.
5. F1 -> Click "Delete Scale Rule..." -> Select a container apps environment -> Select a container app -> Select a revision (Not the latest one).
6. Check the scale rule list.

**Expect:**
Display the scale rule list of the revision selected in step 5.

**Actual:**
Display the scale rule list of the latest revision.
<img width="1044" alt="Screenshot 2023-10-17 at 10 40 39" src="https://github.com/microsoft/vscode-azurecontainerapps/assets/57662131/55633986-240f-4eb3-89fc-cde8a94e87e0">

MicroFish91:
Closed by the above PR

PR #499, opened 2023-10-17
Leverage new `pickScaleRule` util when deleting scale rules

### LINKED ISSUE #495: An error occurs when executing "Delete Scale Rule..." command for the revision from command palette
**OS:** All
**Build Version:** 20231016.3

**Repro Steps:**
1. Create a container app.
2. Make sure the revision mode is multiple.
3. Create scale rules.
4. F1 -> Click "Delete Scale Rule..." -> Select a container apps environment -> Select a container app -> Select a revision contains scale rules (Not the latest one).
5. Check whether no error occurs.

**Expect:**
No error occurs.

**Actual:**
An error occurs.
Screenshot 2023-10-17 at 09 20 18

**More Info:**

1. This issue doesn't reproduce when executing "Delete Scale Rule..." command for the latest revision.
2. This issue doesn't reproduce when executing "Delete Scale Rule..." action from context menu.

MicroFish91:
Closes #495, #496


src/commands/scaling/scaleRule/ScaleRuleListStep.ts
```diff
@@ -18,6 +18,7 @@ export class ScaleRuleListStep extends AzureWizardPromptStep<ScaleRuleContext> {
     }
 
     private async getPicks(context: ScaleRuleContext): Promise<IAzureQuickPickItem<ScaleRule>[] | undefined> {
+        // Todo: Expand support for multiple revisions
```
motm32, line 21: Don't think this is being used anywhere else. May want to add a comment on why we are still keeping this file.

MicroFish91: Maybe we should just remove it then and add it back in when we need it again?

motm32: Sounds good to me!

motm32 (approved):

PR #502, opened 2023-10-18
Increase draft creation visibility in the tree view

MicroFish91:
Anthony brought this scenario up a while back.  When running certain draft commands without the tree expanded, it may be difficult to locate changes.  I did my best to expand basically as high as we can currently go with reveal.  I can't reveal the actual container app because the id is currently tied to the hidden resource and not the container apps shown under the managed environment group.

I also moved the `showCreatingChild` into the `RevisionDraftFileSystem` create method so that the tree item shows up properly regardless of the command entry-point.


nturinski:
It doesn't seem to work if you don't have the subscription expanded ahead of time. Was that a limitation or something that we could do?

nturinski (approved):

MicroFish91:
> It doesn't seem to work if you don't have the subscription expanded ahead of time. Was that a limitation or something that we could do?

It should work I think, just sometimes is a bit slow :P

PR #503, opened 2023-10-18
Shorten the container app creation description

MicroFish91:
I think it looks better shorter, what do you think?

image

vs.

image


motm32:
I agree I like the shorter description

nturinski (approved):
I like the specificity of "container app" but really don't have a strong opinion. If the user is going through the steps to create a container app, I'm sure that they know what is being created.

MicroFish91:
Yeah plus the activity log already shows that it's a container app being created.  I like the shorter length since the side-pane width is typically more limited.

PR #504, opened 2023-10-18
Update location setting behavior for `deployWorkspaceProject`

MicroFish91:
I was thinking it may make more sense to set the location to match the highest resource specificity.  

Previously I stopped at the managed environment level, but started thinking that maybe it makes more sense to override with the container app location if one is found.  Pretty much the only resource which would remain to be created in that scenario would be the container registry.. Thoughts?

### DIFF
--- a/src/commands/deployWorkspaceProject/deployWorkspaceProject.ts
+++ b/src/commands/deployWorkspaceProject/deployWorkspaceProject.ts
@@ -95,9 +95,7 @@ export async function deployWorkspaceProject(context: IActionContext, item?: Con
         );
 
         await LocationListStep.setLocation(wizardContext, wizardContext.managedEnvironment.location);
-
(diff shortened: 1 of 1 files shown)


nturinski:
Personally, I would have thought the opposite direction. If the user specifies a resource group, then we take the location from there because it should be like the "folder" that dictates all of those common settings.

MicroFish91:
Yeah, to be fair we never just start with a resource group though.  It's always either a managed environment or a container app, and from that resource we extract the resource group to use.  So the thought exercise is really should we pull the resource group from the managed environment or the container app.

MicroFish91:
Ahh that's a good point, especially since we're typically extracting the resource group tied to these existing resources.  I think I agree with you and I'll refactor it to do that instead.

nturinski (approved):

PR #511, opened 2023-10-20
Remove symbols from `loginServer` when building `passwordSecretRef`

### LINKED ISSUE #488: Fail to execute "Deploy Image to Azure Container Apps..." action
**OS:** All
**Build Version:** 20231012.1
**Regression:** Not a regression

**Repro Steps:**

1. Connect "Generic Registry V2" registry with URL "http://localhost:5100/".
2. Push one image to the above registry.
3. Expand the above registry -> Right click the image -> Select "Deploy Image to Azure Container Apps..." action -> Select a container app environment -> Select a container app.
4. Check whether succeeds to execute "Deploy Image to Azure Container Apps..." action.

**Actual:**
Fail to execute "Deploy Image to Azure Container Apps..." action.
image

**Suggestion:**
It would be better to remove "Deploy Image to Azure Container Apps..." action for the Generic Registry V2 registry.

**More Info:**
Succeeds to execute "Deploy Image to Azure App Service..." action for the Generic Registry V2 registry..

MicroFish91:
Fixes #488

### DIFF
--- a/src/commands/image/imageSource/containerRegistry/getRegistryCredentialsAndSecrets.ts
+++ b/src/commands/image/imageSource/containerRegistry/getRegistryCredentialsAndSecrets.ts
@@ -40,7 +40,7 @@ export async function getAcrCredentialsAndSecrets(context: IContainerRegistryIma
 export function getThirdPartyCredentialsAndSecrets(context: IContainerRegistryImageContext, containerAppSettings?: RegistryCredentialsAndSecrets): RegistryCredentialsAndSecrets {
     // If 'docker.io', convert to 'index.docker.io', else use registryName as loginServer
     const loginServer: string = (context.registryDomain === dockerHubDomain) ? dockerHubRegistry : nonNullProp(context, 'registryName').toLowerCase();
-    const passwordSecretRef: string = `${loginServer.replace(/[\.]+/g, '')}-${context.username}`;
+    const passwordSecretRef: string = `${loginServer.replace(/[^a-z0-9-]+/g, '')}-${context.username}`;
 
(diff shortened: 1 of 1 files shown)


src/commands/image/imageSource/containerRegistry/getRegistryCredentialsAndSecrets.ts
```diff
@@ -40,7 +40,7 @@ export async function getAcrCredentialsAndSecrets(context: IContainerRegistryIma
 export function getThirdPartyCredentialsAndSecrets(context: IContainerRegistryImageContext, containerAppSettings?: RegistryCredentialsAndSecrets): RegistryCredentialsAndSecrets {
     // If 'docker.io', convert to 'index.docker.io', else use registryName as loginServer
     const loginServer: string = (context.registryDomain === dockerHubDomain) ? dockerHubRegistry : nonNullProp(context, 'registryName').toLowerCase();
-    const passwordSecretRef: string = `${loginServer.replace(/[\.]+/g, '')}-${context.username}`;
+    const passwordSecretRef: string = `${loginServer.replace(/[^a-z0-9-]+/g, '')}-${context.username}`;
```
motm32, line 43: I think based on the error message this should also include a "." character.

MicroFish91: This is actually intentional, otherwise this will happen:

image

I'm not really sure why we're seeing a different error than the previously posted one but we can avoid both errors pretty easily by just omitting "."

nturinski (approved):

PR #514, opened 2023-10-22
Add telemetry to `deployWorkspaceProject`

MicroFish91:


src/telemetry/SetTelemetryProps.ts
```diff
@@ -0,0 +1,10 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+export type SetTelemetryProps<Props extends object> = {
```
MicroFish91, line 6: Revisiting this, realized I'm overwriting a few telemetry properties that I hadn't intended, will update soon..

src/commands/deployWorkspaceProject/getDefaultValues/getDefaultContainerAppsResources/getResourceHelpers.ts
```diff
@@ -16,6 +16,7 @@ export async function getResourcesFromContainerAppHelper(context: ISubscriptionA
 
     const managedEnvironments: ManagedEnvironment[] = await uiUtils.listAllIterator(client.managedEnvironments.listBySubscription());
     const managedEnvironment = managedEnvironments.find(env => env.id === containerApp.managedEnvironmentId);
+    context.telemetry.properties.managedEnvironmentCount = String(managedEnvironments.length);
```
nturinski, line 19: Since this is only going to be called for `deployWorkspaceProject`, I'd recommend putting this somewhere else since I imagine that we want it for every command involving a container app.

You may be able to make a helper command to get all the azure resource count telemetry that you want. In the past, it was easy enough to just grab a parent's children and see how many there were, but since that's not possible anymore here are a couple of ways I can think of:

1. Save a global context that updates every time the listAll call is made. Would probably want to make a wrapper so that it's consistently tracked/updated.
2. Just make a list call of what you want count of wherever else we want the telemetry.

Honestly though, I'm not sure _how_ valuable it is to know managed environment count, especially since way 1 is kind of a pain and way 2 slows down the command. 

I see it's value mostly in 3 places:
1. `deployWorkspaceProject` (here)
2. `createManagedEnviornment`
3. `createContainerApp`

src/commands/deployWorkspaceProject/getDefaultValues/getDefaultContextValues.ts
```diff
@@ -23,9 +23,13 @@ export async function getDefaultContextValues(context: ISubscriptionActionContex
     const { rootFolder, dockerfilePath } = await getWorkspaceProjectPaths(context);
     const settings: DeployWorkspaceProjectSettings = await getDeployWorkspaceProjectSettings(rootFolder);
 
+    setDeployWorkspaceProjectSettingsTelemetry(context, settings);
+
     if (triggerSettingsOverride(settings, item)) {
         // Tree item / settings conflict
+        context.telemetry.properties.triggeredSettingsOverride = 'true';
```
nturinski, line 30: I think it could be easier for us to query this property if it was formatted like:

`properties.settingsOverride = triggered|accepted`. We know if it was accepted, it was triggered. That way we can just query the one property (unless there's some reason we'd only query one property and not the other)

src/commands/deployWorkspaceProject/getDefaultValues/getWorkspaceProjectPaths.ts
```diff
@@ -14,6 +14,8 @@ export async function getWorkspaceProjectPaths(context: IActionContext): Promise
     const rootFolder: WorkspaceFolder | undefined = await getRootWorkspaceFolder(prompt);
```
nturinski, line 15: Mentioned it somewhere else, but I'd mark `context.telemetry.properties.hasWorkspaceProjectOpen = 'true';` here so that it's easier to compare against. I think it just helps my brain more when I see it in the dashboard.

src/commands/deployWorkspaceProject/DeployWorkspaceProjectConfirmStep.ts
```diff
@@ -51,6 +51,7 @@ export class DeployWorkspaceProjectConfirmStep extends AzureWizardPromptStep<Dep
         );
 
         ext.outputChannel.appendLog(outputMessage);
+        context.telemetry.properties.confirmedResourceCreation = 'true';
```
nturinski, line 54: I don't think we explicitly need this because we can just check cancelled events and see if the step was `DeployWorkspaceProjectConfirmStep`

src/commands/deployWorkspaceProject/DeployWorkspaceProjectSaveSettingsStep.ts
```diff
@@ -37,7 +37,9 @@ export class DeployWorkspaceProjectSaveSettingsStep extends ExecuteActivityOutpu
         return !!context.shouldSaveDeploySettings;
     }
 
-    protected createSuccessOutput(): ExecuteActivityOutput {
+    protected createSuccessOutput(context: DeployWorkspaceProjectContext): ExecuteActivityOutput {
+        context.telemetry.properties.didSaveSettings = 'true';
```
nturinski, line 41: I think I would prefer this to be in `ExecuteActivityOutputStepBase` as a generic property. Maybe something like
`properties.createOutput = 'success'|'fail'`. I think that this information could be useful for all of these type of steps.

src/commands/deployWorkspaceProject/ShouldSaveDeploySettingsPromptStep.ts
```diff
@@ -32,6 +32,7 @@ export class ShouldSaveDeploySettingsPromptStep extends AzureWizardPromptStep<De
         );
 
         context.shouldSaveDeploySettings = userResponse === saveItem;
+        context.telemetry.properties.shouldSaveSettings = context.shouldSaveDeploySettings ? 'true' : 'false';
```
nturinski, line 35: nit: I think it's easier to look for this property later if it matches the setting name. Also, in the future, we may have another "saveSettings" property that has nothing to do with deploy so it would make the property name a little bit confusing.
```suggestion
        context.telemetry.properties.shouldSaveDeploySettings = context.shouldSaveDeploySettings ? 'true' : 'false';
```

src/commands/deployWorkspaceProject/deployWorkspaceProject.ts
```diff
@@ -188,8 +205,9 @@ export async function deployWorkspaceProject(context: IActionContext, item?: Con
     await wizard.execute();
 
     displayNotification(wizardContext);
-    ext.branchDataProvider.refresh();
+    wizardContext.telemetry.properties.revisionMode = wizardContext.containerApp?.revisionsMode;
```
nturinski, line 208: We can keep this here, but I reccommend that we also track this whenever we list container apps. I think it'll be helpful to see how many users actually use multiple revision mode to further drive some of our design decisions.

src/commands/deployWorkspaceProject/deployWorkspaceProject.ts
```diff
@@ -202,11 +220,19 @@ function displayNotification(context: DeployWorkspaceProjectContext): void {
 
     const containerApp: ContainerAppModel = nonNullProp(context, 'containerApp');
     void window.showInformationMessage(message, ...buttonMessages).then(async (result: string | undefined) => {
-        if (result === viewOutput) {
-            ext.outputChannel.show();
-        } else if (result === browse) {
-            await browseContainerApp(containerApp);
-        }
+        await callWithTelemetryAndErrorHandling('deployWorkspaceProject.displayNotification',
```
nturinski, line 223: 👍

src/commands/image/imageSource/EnvironmentVariablesListStep.ts
```diff
@@ -29,6 +29,8 @@ export class EnvironmentVariablesListStep extends AzureWizardPromptStep<ImageSou
 
     public async configureBeforePrompt(context: ImageSourceBaseContext): Promise<void> {
         if (context.environmentVariables?.length === 0) {
+            context.telemetry.properties.environmentVariableFileCount = '0';
+            context.telemetry.properties.setEnvironmentVariableOption = SetEnvironmentVariableOption.NoDotEnv;
             this.outputLogs(context, SetEnvironmentVariableOption.NoDotEnv);
         }
     }
```
nturinski, line 36: Conversely, it may be valuable to set the count if it's more than 0.

src/utils/workspaceUtils.ts
```diff
@@ -33,6 +33,17 @@ export async function selectWorkspaceFile(context: IActionContext, placeHolder:
     if (workspace.workspaceFolders?.length === 1) {
         // if there's a fileExtension, then only populate the quickPick menu with that, otherwise show the current folders in the workspace
         const files = globPattern ? await workspace.findFiles(globPattern) : await workspace.findFiles('**/*');
+
```
nturinski, line 36: I'd log the counts both at 0 here.

src/commands/deployWorkspaceProject/getDefaultValues/getDefaultContainerAppsResources/getResourceHelpers.ts
MicroFish91: It's not super important, I only logged it because we were already making a query and thought it could be interesting.  I don't really see a deep need to log it for any specific purposes though

src/commands/ContainerAppOverwriteConfirmStep.ts
```diff
@@ -14,6 +14,8 @@ export class ContainerAppOverwriteConfirmStep<T extends IContainerAppContext> ex
     public hideStepCount: boolean = true;
 
     public async prompt(context: T): Promise<void> {
+        context.telemetry.properties.hasUnsupportedFeatures = 'true';
```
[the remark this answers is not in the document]
MicroFish91: That's easy enough, like you said it's just a bit uglier, but I can probably just go in and log non-default states.  The initialization could work, but I think it might end up being slightly more complicated, so I'll prefer the other way for now unless it ends up looking bad

src/commands/deployWorkspaceProject/DeployWorkspaceProjectSaveSettingsStep.ts
MicroFish91: We'll have to discuss this one offline, I'm not entirely sure how this property would get mapped back to the step it is associated with

src/commands/image/imageSource/EnvironmentVariablesListStep.ts
MicroFish91: Agree, I ended up logging all the files based on glob patterns searched under `selectWorkspaceFile`:

image

Hopefully that's an okay way to handle this

src/utils/workspaceUtils.ts
MicroFish91: It should already be logged before it hits this point I believe

MicroFish91: Oh nevermind I'm dumb, I see what you mean now hahaha

End of day brain is mush brain

src/telemetry/WorkspaceFileTelemetryProps.ts
```diff
@@ -0,0 +1,9 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.txt in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+export interface WorkspaceFileTelemetryProps {
+    dockerfileCount?: string;
```
MicroFish91, line 7: This is new

nturinski: Really need to review this one...

nturinski (approved):

PR #517, opened 2023-10-23
Add telemetry to `deployImageApi`

MicroFish91:
Added telemetry to see what domains we're deploying from and whether or not they are private or public registries


src/commands/image/deployImageApi/deployImage.ts
```diff
@@ -26,6 +26,9 @@ export async function deployImage(context: IActionContext & Partial<IContainerRe
         containerApp
     };
 
+    wizardContext.telemetry.properties.revisionMode = containerApp.revisionsMode;
+    wizardContext.telemetry.properties.registryDomain = context.registryDomain;
```
nturinski, line 30: I think we may want to change this to automatically be "other" in the case `context.registryDomain` is undefined.

src/telemetry/telemetryProps.ts
```diff
@@ -4,8 +4,17 @@
 *--------------------------------------------------------------------------------------------*/
```
nturinski, line 4: Are you planning to change `telemetryProps` to be more specific ie naming it `DeployImageApiTelemetryProps` or something?

nturinski (approved):

src/telemetry/telemetryProps.ts
MicroFish91: I was going to rename the file to something like `commandTelemetryProps` and then continue sticking all the command props in this file.  The shared props can be contained in separate files

MicroFish91: Unless you mean in the command itself, which I'm actually doing this:

image

MicroFish91: Unless you're referring to how it's being called, because I'm actually renaming all of them to `TelemetryProps` when imported because it's easier to read:

image

nturinski: No, I was referring to the name of the file.

PR #519, opened 2023-10-24
Add telemetry for `updateImage` and consolidate `ImageSource` base telemetry props

MicroFish91:
Intellisense is basically showing up for telemetry properties in all the steps now and I think it's pretty sweet :D

Changes:
- Added `UpdateImageTelemetryProps`
- Added / consolidated `ImageSourceTelemetryProps` that many commands can now extend from
- Added matching telemetry properties in the steps
- Lots of automatic file renaming due to the updated context names/patterns


nturinski (approved):
This is pretty dope.

PR #521, opened 2023-10-25
Add `OverwriteConfirmStepBase` and reduce number of potential confirm step prompts

MicroFish91:
Some of the commands had the potential to show two confirm steps - one specific to the command deployment and one related to whether or not the the container app might lose some unsupported features.

To reduce the number of potential confirm prompts that can show up, I created a base overwrite confirm step that gives the same capabilities for checking as our old `ContainerAppOverwriteConfirmStep`.  By extending from this base, we now have greater flexibility in merging overwrite prompt steps together so we don't need to show as many.

Note: The bottom message on each prompt will only show up if we detect unsupported features, else it will display the prompt without it or not at all.

**New:**

_deployImageApi_:
image

_revisionDraftDeploy_:
image

_deployWorkspaceProject_:
image

image


src/commands/image/updateImage/UpdateRegistryAndSecretsStep.ts
```diff
@@ -29,6 +29,8 @@ export class UpdateRegistryAndSecretsStep extends AzureWizardExecuteStep<UpdateI
             return;
         }
 
+        context.telemetry.properties.skippedRegistryCredentialUpdate = 'false';
```
MicroFish91, line 32: Forgot to add this in the last PR merge

src/commands/deployWorkspaceProject/DeployWorkspaceProjectConfirmStep.ts
```diff
@@ -44,6 +45,11 @@ export class DeployWorkspaceProjectConfirmStep extends AzureWizardPromptStep<Dep
             outputMessage = localize('overwriteConfirmed', 'User confirmed overwrite of container app "{0}" and the latest image of registry "{1}".', context.containerApp?.name, context.registry?.name);
         }
 
+        if (this.hasUnsupportedFeatures(context)) {
+            confirmMessage += '\n\n' + this.unsupportedFeaturesWarning;
+            outputMessage += ' ' + localize('unsupportedOverwriteConfirmed', 'User also confirmed that proceeding means that any unsupported container app features in VS Code will be lost.');
```
nturinski, line 50: May be just me, but this feels a little extra.

nturinski (approved):

src/commands/deployWorkspaceProject/DeployWorkspaceProjectConfirmStep.ts
MicroFish91: I'll remove it :P

PR #524, opened 2023-10-26
Add `CreateContainerApp` telemetry

motm32:
- Added `CreateContainerAppTelemetryProps` 
- Renamed the `ICreateContainerAppContext` -> `CreateContainerAppContext`
- Added more ingress telemetry points
- Added image source telemetry points

### DIFF
--- a/src/commands/image/imageSource/ImageSourceListStep.ts
+++ b/src/commands/image/imageSource/ImageSourceListStep.ts
@@ -50,14 +50,17 @@ export class ImageSourceListStep extends AzureWizardPromptStep<ImageSourceContex
         switch (context.imageSource) {
             case ImageSource.QuickStartImage:
                 setQuickStartImage(context);
+                context.telemetry.properties.imageSource = ImageSource.QuickStartImage;
                 break;
             case ImageSource.ContainerRegistry:
                 promptSteps.push(new ContainerRegistryListStep());
                 executeSteps.push(new ContainerRegistryImageConfigureStep());
+                context.telemetry.properties.imageSource = ImageSource.ContainerRegistry;
                 break;
             case ImageSource.RemoteAcrBuild:
                 promptSteps.push(new BuildFromProjectListStep());
                 executeSteps.push(new ContainerRegistryImageConfigureStep());
+                context.telemetry.properties.imageSource = ImageSource.RemoteAcrBuild;
                 break;
             default:
         }
--- a/src/telemetry/commandTelemetryProps.ts
+++ b/src/telemetry/commandTelemetryProps.ts
@@ -5,7 +5,7 @@
 
 import type { KnownActiveRevisionsMode } from "@azure/arm-appcontainers";
 import type { AzdTelemetryProps } from "./AzdTelemetryProps";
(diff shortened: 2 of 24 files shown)


src/telemetry/commandTelemetryProps.ts
```diff
@@ -57,3 +53,16 @@ export interface UpdateImageTelemetryProps extends AzdTelemetryProps, ImageSourc
     revisionMode?: KnownActiveRevisionsMode;
     skippedRegistryCredentialUpdate?: 'true';
 }
+
+export interface IngressTelemetryProps {
```
MicroFish91, line 57: I think it would be good to migrate this to its own file `IngressTelemetryProps.ts`

src/telemetry/commandTelemetryProps.ts
```diff
@@ -57,3 +53,16 @@ export interface UpdateImageTelemetryProps extends AzdTelemetryProps, ImageSourc
...
+    dockerfileExposePortRangeCount?: string;  // IngressPromptStep
+    enableIngress?: 'true' | 'false'; //IngressPromptStep
+    suggestedTargetPort?: string; //getDefaultPort
+    targetPort?: string; //TargetPortInputStep
+}
+
+export interface CreateContainerAppTelemetryProps extends ImageSourceTelemetryProps, EnvironmentVariableTelemetryProps, AzdTelemetryProps, IngressTelemetryProps {
+    // Ingress
+    enableIngress?: 'true' | 'false'; //IngressPromptStep
```
MicroFish91, line 66: Couple things, I just added a PR removing the extra `EnvironmentVariableTelemetryProps` file, so you can just extend `ImageSourceTelemetryProps` now and it will cover the environment variables!

Also, now that you added `IngressTelemetryProps` I think you can probably remove these single ingress properties.

If we end up just extending and not needing to add any more custom properties, it might look cleaner to do something like:

`export type CreateContainerAppTelemetryProps = ImageSourceTelemetryProps & AzdTelemetryProps...`

src/commands/image/imageSource/ImageSourceListStep.ts
```diff
@@ -50,14 +50,17 @@ export class ImageSourceListStep extends AzureWizardPromptStep<ImageSourceContex
...
                 break;
             case ImageSource.ContainerRegistry:
                 promptSteps.push(new ContainerRegistryListStep());
                 executeSteps.push(new ContainerRegistryImageConfigureStep());
+                context.telemetry.properties.imageSource = ImageSource.ContainerRegistry;
                 break;
             case ImageSource.RemoteAcrBuild:
                 promptSteps.push(new BuildFromProjectListStep());
                 executeSteps.push(new ContainerRegistryImageConfigureStep());
+                context.telemetry.properties.imageSource = ImageSource.RemoteAcrBuild;
```
MicroFish91, line 63: Good catch on this 🫡, I clearly forgot to add the `imageSource` telemetry even though I marked it in the interface initially

MicroFish91 (approved):
Woo last big PR before v0.6.0!!! 🚀🫡

PR #529, opened 2023-10-30
Always assume dockerfile used is src folder

nturinski:
When looking for the Dockerfile, VS Code's `findFiles` checks every opened folder in the workspace. That means it'll find one even if it's a sub folder, ie the src folder, such as in this [sample](https://github.com/Azure-Samples/containerapps-albumapi-javascript).

We should assume that wherever the dockerfile lives, that's the root of the source folder and that's what should be deployed, not the root of the currently opened workspace.

### DIFF
--- a/src/commands/image/imageSource/buildImageInAzure/ImageNameStep.ts
+++ b/src/commands/image/imageSource/buildImageInAzure/ImageNameStep.ts
@@ -10,7 +10,7 @@ import { BuildImageInAzureImageSourceContext } from "./BuildImageInAzureContext"
 
 export class ImageNameStep extends AzureWizardPromptStep<BuildImageInAzureImageSourceContext> {
     public async prompt(context: BuildImageInAzureImageSourceContext): Promise<void> {
-        const suggestedImageName = await getSuggestedName(context, context.dockerfilePath);
+        const suggestedImageName = await getSuggestedName(context, context.rootFolder.uri.fsPath);
 
         context.imageName = await context.ui.showInputBox({
             prompt: localize('imageNamePrompt', 'Enter a name for the image'),
--- a/src/commands/image/imageSource/buildImageInAzure/RunStep.ts
+++ b/src/commands/image/imageSource/buildImageInAzure/RunStep.ts
@@ -19,15 +19,13 @@ export class RunStep extends ExecuteActivityOutputStepBase<BuildImageInAzureImag
     protected async executeCore(context: BuildImageInAzureImageSourceContext, progress: Progress<{ message?: string | undefined; increment?: number | undefined }>): Promise<void> {
         // Need to keep the additional try wrapper here to execute finally, then we can catch any error that percolates up and display its output
         try {
-            const rootUri = context.rootFolder.uri;
-
             const runRequest: AcrDockerBuildRequest = {
                 type: 'DockerBuildRequest',
                 imageNames: [context.imageName],
                 isPushEnabled: true,
                 sourceLocation: context.uploadedSourceLocation,
                 platform: { os: context.os },
-                dockerFilePath: path.relative(rootUri.path, context.dockerfilePath)
+                dockerFilePath: path.join('.', 'Dockerfile') /* Assume the dockerfile is always in the root of the source */
             };
 
             const building: string = localize('buildingImage', 'Building image...');
--- a/src/commands/image/imageSource/buildImageInAzure/RunStep.ts
+++ b/src/commands/image/imageSource/buildImageInAzure/RunStep.ts
@@ -25,7 +25,7 @@ export class RunStep extends ExecuteActivityOutputStepBase<BuildImageInAzureImag
                 isPushEnabled: true,
                 sourceLocation: context.uploadedSourceLocation,
                 platform: { os: context.os },
-                dockerFilePath: path.join('.', 'Dockerfile') /* Assume the dockerfile is always in the root of the source */
+                dockerFilePath: path.basename(context.dockerfilePath) /* Assume the dockerfile is always in the root of the source */
             };
 
             const building: string = localize('buildingImage', 'Building image...');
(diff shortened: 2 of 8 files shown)


src/commands/image/imageSource/buildImageInAzure/UploadSourceCodeStep.ts
```diff
@@ -13,6 +13,7 @@ import { createActivityChildContext } from '../../../../utils/activity/activityU
 import { createContainerRegistryManagementClient } from '../../../../utils/azureClients';
 import { localize } from '../../../../utils/localize';
 import { BuildImageInAzureImageSourceContext } from './BuildImageInAzureContext';
+import path = require('path');
```
MicroFish91, line 16: Thanks VS Code 😄

src/commands/image/imageSource/buildImageInAzure/RunStep.ts
```diff
@@ -19,15 +19,13 @@ export class RunStep extends ExecuteActivityOutputStepBase<BuildImageInAzureImag
...
-            const rootUri = context.rootFolder.uri;
-
             const runRequest: AcrDockerBuildRequest = {
                 type: 'DockerBuildRequest',
                 imageNames: [context.imageName],
                 isPushEnabled: true,
                 sourceLocation: context.uploadedSourceLocation,
                 platform: { os: context.os },
-                dockerFilePath: path.relative(rootUri.path, context.dockerfilePath)
+                dockerFilePath: path.join('.', 'Dockerfile') /* Assume the dockerfile is always in the root of the source */
```
MicroFish91, line 28: Do we want to pick up the Dockerfile name using `context.dockerfilePath`?

MicroFish91 (approved):
I like it 🚀

MicroFish91:
I forgot, but since this impacts the `Build Image in Azure` workflow, we'll probably need to define `context.sourceFilePath` somewhere in that prompt flow as well.  Maybe in `DockerFileItemStep`?

alexweininger (approved):

nturinski:
Originally I was going to have two separate messages: one where the source folder is the root and not. The former would have been the old messages, and the latter, the new message. I decided that it looks fine and is kind of valuable information for users to have "." as the directory and easier to manage all the messages since that would require to maintain 5 more strings.

With a source folder:
image

image

image

With the source as the root:
image

image

image

src/commands/image/imageSource/buildImageInAzure/UploadSourceCodeStep.ts
```diff
@@ -13,22 +14,23 @@ import { createActivityChildContext } from '../../../../utils/activity/activityU
...
 
     protected async executeCore(context: BuildImageInAzureImageSourceContext, progress: Progress<{ message?: string | undefined; increment?: number | undefined }>): Promise<void> {
         context.registryName = nonNullValue(context.registry?.name);
         context.resourceGroupName = getResourceGroupFromId(nonNullValue(context.registry?.id));
         context.client = await createContainerRegistryManagementClient(context);
+        /* relative path of src folder from rootFolder and what gets deployed*/
+        this._sourceFilePath = path.dirname(path.relative(context.rootFolder.uri.path, context.dockerfilePath));
 
-        const uploading: string = localize('uploadingSourceCode', 'Uploading source code...');
+        const uploading: string = localize('uploadingSourceCode', 'Uploading source code from "{0}" directory...', this._sourceFilePath);
```
MicroFish91, line 31: My vote is to keep the progress report message the old way, it's a little easier on the eyes in my opinion and we can see the longer versions in the output and activity log later.

src/commands/image/imageSource/buildImageInAzure/UploadSourceCodeStep.ts
```diff
@@ -13,22 +14,23 @@ import { createActivityChildContext } from '../../../../utils/activity/activityU
...
 export class UploadSourceCodeStep extends ExecuteActivityOutputStepBase<BuildImageInAzureImageSourceContext> {
     public priority: number = 430;
+    private _sourceFilePath: string;
 
     protected async executeCore(context: BuildImageInAzureImageSourceContext, progress: Progress<{ message?: string | undefined; increment?: number | undefined }>): Promise<void> {
         context.registryName = nonNullValue(context.registry?.name);
         context.resourceGroupName = getResourceGroupFromId(nonNullValue(context.registry?.id));
         context.client = await createContainerRegistryManagementClient(context);
+        /* relative path of src folder from rootFolder and what gets deployed*/
+        this._sourceFilePath = path.dirname(path.relative(context.rootFolder.uri.path, context.dockerfilePath));
```
MicroFish91, line 29: 😎

src/commands/image/imageSource/buildImageInAzure/ImageNameStep.ts
```diff
@@ -10,7 +10,7 @@ import { BuildImageInAzureImageSourceContext } from "./BuildImageInAzureContext"
 
 export class ImageNameStep extends AzureWizardPromptStep<BuildImageInAzureImageSourceContext> {
     public async prompt(context: BuildImageInAzureImageSourceContext): Promise<void> {
-        const suggestedImageName = await getSuggestedName(context, context.dockerfilePath);
+        const suggestedImageName = await getSuggestedName(context, context.rootFolder.uri.fsPath);
```
nturinski, line 13: I changed this because the dockerfile can be in the "src" folder, which causes the suggestion name to be "src". Figured that this is the behavior that we wanted.

src/commands/image/imageSource/buildImageInAzure/UploadSourceCodeStep.ts
nturinski: 🤓

src/commands/image/imageSource/buildImageInAzure/UploadSourceCodeStep.ts
nturinski: kk that's fine with me

PR #545, opened 2023-10-31
Automatically generate a unique name for registries

nturinski:
Because registry names have to be globally unique, it's almost guaranteed to have failed the workspace name check. This ensures that the registry name will almost always be unique.

### DIFF
--- a/src/commands/image/imageSource/containerRegistry/acr/createAcr/RegistryNameStep.ts
+++ b/src/commands/image/imageSource/containerRegistry/acr/createAcr/RegistryNameStep.ts
@@ -4,9 +4,10 @@
 *--------------------------------------------------------------------------------------------*/
 
 import { ContainerRegistryManagementClient, RegistryNameStatus } from "@azure/arm-containerregistry";
-import { AzureWizardPromptStep, ISubscriptionActionContext } from "@microsoft/vscode-azext-utils";
+import { AzureWizardPromptStep, ISubscriptionActionContext, randomUtils } from "@microsoft/vscode-azext-utils";
 import { createContainerRegistryManagementClient } from "../../../../../../utils/azureClients";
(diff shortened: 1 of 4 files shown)


src/commands/image/imageSource/containerRegistry/acr/createAcr/RegistryNameStep.ts
```diff
@@ -51,4 +52,29 @@ export class RegistryNameStep extends AzureWizardPromptStep<CreateAcrContext> {
             return { nameAvailable: true };
         }
     }
+
+    public static async tryGenerateRelatedName(context: DeployWorkspaceProjectContext, name: string): Promise<string | undefined> {
+        let registryAvailable: boolean = false;
+        let generatedName: string = '';
+
+        const timeoutSeconds: number = 60;
```
MicroFish91, line 60: Although rare that we would ever hit the full length of this timeout, I'm wondering if the timeout is too long to be hanging on an input name check.  Not really sure to be honest but just thought I'd bring it up, let me know what you think

MicroFish91 (approved):
Everything looks good! :D 

I'll run a few extra tests tonight just to double check though, will follow-up when finished with a comment

src/commands/image/imageSource/containerRegistry/acr/createAcr/RegistryNameStep.ts
nturinski: Yeah, I could see that. Maybe just try for 10 seconds?

MicroFish91: Yeah I personally think I'd be happy with anything between 10-30s.

MicroFish91:
I tested this with quite a few scenarios and all of them seem to work great, so probably good to merge :)

PR #546, opened 2023-10-31
Add telemetry for source code depth from root

nturinski:


src/commands/image/imageSource/buildImageInAzure/UploadSourceCodeStep.ts
```diff
@@ -27,6 +27,7 @@ export class UploadSourceCodeStep extends ExecuteActivityOutputStepBase<BuildIma
         context.client = await createContainerRegistryManagementClient(context);
         /* relative path of src folder from rootFolder and what gets deployed */
         this._sourceFilePath = path.dirname(path.relative(context.rootFolder.uri.path, context.dockerfilePath));
+        context.telemetry.properties.sourceDepth = this._sourceFilePath === '.' ? '0' : String(this._sourceFilePath.split(path.sep).length);
```
MicroFish91, line 30: Not sure, should this be length or length - 1?

MicroFish91 (approved):

src/commands/image/imageSource/buildImageInAzure/UploadSourceCodeStep.ts
nturinski: It should be length.

src/sub1/sub2 would be split into [src, sub1, sub2].length which equals 3.

PR #548, opened 2023-11-01
Add `deployWorkspaceProject` to README

MicroFish91:


MicroFish91:
I didn't change the tutorial url with the assumption that when Anthony does the update, it will just point to the updated tutorial.

PR #558, opened 2023-11-03
Add `Microsoft.ContainerRegistry` provider to the verify providers list

### LINKED ISSUE #557: Verify `Microsoft.ContainerRegistry` provider
We should verify this provider for all the different steps.

image

MicroFish91:
Fixes #557

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -11,6 +11,7 @@ export const managedEnvironmentsId = 'managedEnvironments';
 export const containerAppsId = 'containerApps';
 export const appProvider: string = 'Microsoft.App';
 export const webProvider: string = 'Microsoft.Web';
+export const registryProvider: string = 'Microsoft.ContainerRegistry';
 export const operationalInsightsProvider: string = 'Microsoft.OperationalInsights';
 
 export const containerAppsWebProvider: string = `${webProvider}/${containerAppsId}`;
--- a/src/utils/getVerifyProvidersStep.ts
+++ b/src/utils/getVerifyProvidersStep.ts
@@ -1,0 +1,17 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.md in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { VerifyProvidersStep } from "@microsoft/vscode-azext-azureutils";
+import type { ISubscriptionActionContext } from "@microsoft/vscode-azext-utils";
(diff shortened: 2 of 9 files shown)


nturinski (commented):
I'd like to approach this differently so there's less room for error.

My two ideas are:

1. Have an array that contains all potential providers that are required by ACA and just always use that array whenever we do `VerifyProvidersStep`
2. Create a util function that returns `VerifyProvidersStep` that contains all of the providers that we need. It can be smart about it by checking the context and see if there are `newXXXXName` properties on it. but honestly, I think we don't lose much by just always verifying that the providers are registered.

src/constants.ts
```diff
@@ -11,6 +11,7 @@ export const managedEnvironmentsId = 'managedEnvironments';
 export const containerAppsId = 'containerApps';
 export const appProvider: string = 'Microsoft.App';
 export const webProvider: string = 'Microsoft.Web';
+export const registryProvider: string = 'Microsoft.ContainerRegistry';
```
nturinski, line 14: I think we should be good to move all these consts to `getVerifyProvidersStep`

src/utils/getVerifyProvidersStep.ts
```diff
@@ -0,0 +1,20 @@
...
+ *--------------------------------------------------------------------------------------------*/
+
+import { VerifyProvidersStep } from "@microsoft/vscode-azext-azureutils";
+import type { ISubscriptionActionContext } from "@microsoft/vscode-azext-utils";
+import { appProvider, operationalInsightsProvider, registryProvider, webProvider } from "../constants";
+
+/**
+ * Use to obtain a `VerifyProvidersStep` that registers all known container app providers to the user's subscription
+ */
+export function getVerifyProvidersStep<T extends ISubscriptionActionContext>(): VerifyProvidersStep<T> {
```
nturinski, line 13: I don't think you need to technically provide a context here based off the old way we used to create them.

src/constants.ts
MicroFish91: We can move them there, but they won't be localized to that file so will still have to export a few of them

src/utils/getVerifyProvidersStep.ts
MicroFish91: Yeah that makes sense, I'll just have to return using type inference instead

src/constants.ts
nturinski: Oh are they? Never mind then-- I'm surprised we are using them anywhere else.

nturinski (approved):

src/constants.ts
MicroFish91: two are, two aren't :P

PR #565, opened 2023-11-07
Make suggested images names unique by appending a timestamped tag

MicroFish91:
Suggested image name example:

image


src/utils/pickItem/pickContainerApp.ts
```diff
@@ -14,7 +14,7 @@ import { getPickEnvironmentSteps } from "./pickEnvironment";
 export function getPickContainerAppStep(containerAppName?: string | RegExp): AzureWizardPromptStep<AzureResourceQuickPickWizardContext> {
     let containerAppFilter: RegExp | undefined;
     if (containerAppName) {
-        containerAppFilter = containerAppName instanceof RegExp ? containerAppName : new RegExp(containerAppName);
+        containerAppFilter = containerAppName instanceof RegExp ? containerAppName : new RegExp(`^${containerAppName}$`);
```
MicroFish91, line 17: Snuck this change in, it's a small one to ensure we do an exact name match.

src/commands/image/updateImage/UpdateImageDraftStep.ts
```diff
@@ -29,8 +29,7 @@ export class UpdateImageDraftStep<T extends UpdateImageContext> extends Revision
         this.revisionDraftTemplate.containers.push({
             env: context.environmentVariables,
             image: context.image,
-            // We need the revision draft to always show up as having unsaved changes, we can ensure this by adding a unique ID at end of the container name
-            name: getContainerNameForImage(nonNullProp(context, 'image')) + `-${randomUtils.getRandomHexString(5)}`,
+            name: getContainerNameForImage(nonNullProp(context, 'image')),
```
MicroFish91, line 32: Based on Anthony's comments, we probably don't want to be updating container apps with identical image names.  I put in the different name originally to still allow us to deploy the same image name, but now there's probably no need for this behavior.

src/commands/image/imageSource/buildImageInAzure/ImageNameStep.ts
```diff
@@ -35,19 +41,35 @@ export class ImageNameStep extends AzureWizardPromptStep<BuildImageInAzureImageS
 
         return undefined;
     }
+
+    static generateSuggestedImageName(repositoryName: string = 'hello-world'): string {
```
nturinski, line 45: I'd say remove the default name here. I'm pretty sure everywhere you use it has a string passed into it. And even if not, I think I'd prefer we provide a default name that makes sense in context rather than "hello-world"

MicroFish91: This was really a just-in-case suggested name fallback because `getSuggestedRepositoryName` can return `undefined`.  My hunch is that by the time it gets to the end, it probably is always a string, but I wasn't 100% confident and I didn't write that section, so just to be safe I added the default.

MicroFish91: Actually I have a better idea. Coming soon to a theater near you ™️

src/commands/image/imageSource/buildImageInAzure/ImageNameStep.ts
```diff
@@ -35,19 +41,35 @@ export class ImageNameStep extends AzureWizardPromptStep<BuildImageInAzureImageS
...
+        const tag: string = getTimestampTag();
+        return repositoryName.slice(0, maxImageNameLength - (tag.length + 1)) + ':' + tag;
```
nturinski, line 47: Why are you referring to the names as a repositoryName? Isn't it an image name?

MicroFish91: What's passed is a repository name, what comes out is an image name

MicroFish91: Okay yeah I'm pretty sure that the part coming in is the repository name:

image

and what comes out is the actual image name

src/commands/image/imageSource/buildImageInAzure/ImageNameStep.ts
```diff
@@ -35,19 +41,35 @@ export class ImageNameStep extends AzureWizardPromptStep<BuildImageInAzureImageS
...
+    }
 }
 
-async function getSuggestedName(context: BuildImageInAzureImageSourceContext, dockerFilePath: string): Promise<string | undefined> {
-    let suggestedImageName: string | undefined;
-    suggestedImageName = Utils.dirname(URI.parse(dockerFilePath)).path.split('/').pop();
-    if (suggestedImageName === '') {
+async function getSuggestedRepositoryName(context: BuildImageInAzureImageSourceContext, dockerFilePath: string): Promise<string | undefined> {
```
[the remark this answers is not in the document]
MicroFish91: Yeah, I _think_ it's probably safe to remove it?

src/commands/image/imageSource/buildImageInAzure/ImageNameStep.ts
nturinski: nit: I'd prefer we call it `getTimestampedImageName` or something like that.

src/commands/image/imageSource/buildImageInAzure/ImageNameStep.ts
nturinski: Ah yeah. Okay, fair enough.

src/commands/image/imageSource/buildImageInAzure/ImageNameStep.ts
```diff
@@ -4,16 +4,21 @@
...
 const maxImageNameLength: number = 46;
 
 export class ImageNameStep extends AzureWizardPromptStep<BuildImageInAzureImageSourceContext> {
     public async prompt(context: BuildImageInAzureImageSourceContext): Promise<void> {
-        const suggestedImageName = await getSuggestedName(context, context.rootFolder.name);
+        const suggestedImageName = ImageNameStep.getTimestampedImageName(
+            context.containerApp?.name ||
+            // Step is also technically reachable from the `createContainerApp` entry point
+            (context as CreateContainerAppContext).newContainerAppName ||
+            'hello-world'
```
nturinski, line 20: Mark one of these as `nonNullValue` lol. I don't want this random 'hello-world' in here.

nturinski (approved):
You will never get to say hello-world while I'm around.

PR #568, opened 2023-11-09
Update with `eslint-config` and enforce import types

MicroFish91:
How does everyone feel about just adding this as an eslint rule so we don't have to think about whether something should be imported as a type or a value for optimization?

This should make it easier to enforce import types where we can essentially just run `eslint --fix`


alexweininger:
Oh @bwateratmsft is going to love this. Yeah we should add this to our shared eslint config.

MicroFish91:
Will update this next week using the shared eslint config*

MicroFish91:
Updated!

nturinski (approved):

PR #576, opened 2023-11-27
Add an entry-point to view ACR build image error logs in the activity log

### LINKED ISSUE #570: Find a better place to display ACR build image logs
Not sure where we should try to put these, maybe as activity children?  Need to investigate/discuss with the team.

MicroFish91:
Requires a couple updates to the Azure Tools shared package first, will provide links soon.

Review Notes:
- The bulk of the implementation you'll want to inspect is shown in `BuildImageStep.ts`
- Less important, but also worth inspecting - `openAcrBuildLogs.ts`
- Rest is just reconfiguring imports and scaffolding the new command

Closes #570

Adds ability to display additional nested children, see new logs item support shown below:

image

Clicking the logs item will open up the logs like so:

image


MicroFish91:
Shared package upgraded and incorporated, ready for review! :)

src/commands/image/imageSource/buildImageInAzure/BuildImageStep.ts
```diff
@@ -63,11 +72,26 @@ export class BuildImageStep extends ExecuteActivityOutputStepBase<BuildImageInAz
     }
 
     protected createFailOutput(context: BuildImageInAzureImageSourceContext): ExecuteActivityOutput {
+        let loadMoreChildrenImpl: (() => Promise<AzExtTreeItem[]>) | undefined;
+        if (this.acrBuildError) {
+            loadMoreChildrenImpl = () => {
+                const buildImageLogsItem = new GenericTreeItem(undefined, {
+                    contextValue: createActivityChildContext(['logsLinkItem']),
+                    label: localize('buildImageLogs', 'Click to view build image logs'),
+                    iconPath: new ThemeIcon('note', new ThemeColor('terminal.ansiWhite')),
```
nturinski, line 81: nit: What about using link-external? Kind of makes sense to me since it's opening a new file. Or maybe file-symlink-file?

Icons here: https://microsoft.github.io/vscode-codicons/dist/codicon.html

src/commands/image/imageSource/buildImageInAzure/BuildImageStep.ts
```diff
@@ -63,11 +72,26 @@ export class BuildImageStep extends ExecuteActivityOutputStepBase<BuildImageInAz
     }
 
     protected createFailOutput(context: BuildImageInAzureImageSourceContext): ExecuteActivityOutput {
+        let loadMoreChildrenImpl: (() => Promise<AzExtTreeItem[]>) | undefined;
```
nturinski, line 75: nit: You could define `loadMoreChildrenImpl` here as `Promise.resolve([]))`. That way on line 94, you wouldn't have to do that check and you could type define `loadMoreChildrenImpl` as just `(() => Promise<AzExtTreeItem[]>)`

nturinski (approved):
Just small nits, rest looks good.

src/commands/image/imageSource/buildImageInAzure/BuildImageStep.ts
MicroFish91: - Link-external: 
image

- File-symlink-file:
image

PR #577, opened 2023-12-05
Add retries for `Build image in Azure`

### LINKED ISSUE #562: Add retries when executing `Build Image in Azure`
There may be a timing issue where a file may have been uploaded to blob storage, but it's not technically ready yet for ACR to download from the link.  Adding retries to the run request will likely fix these timing issues.

motm32:
Fixes #562.

I arbitrarily chose 3 retries since that should cover the timing issue and building the image takes a bit of time. Let me know if anyone disagrees with that choice. 

I also tried a couple of times to test the retries but I could only get it to work on the first try. Guess our code is just too good 🤪

### DIFF
--- a/src/commands/image/imageSource/buildImageInAzure/RunStep.ts
+++ b/src/commands/image/imageSource/buildImageInAzure/RunStep.ts
@@ -38,7 +38,6 @@ export class RunStep extends ExecuteActivityOutputStepBase<BuildImageInAzureImag
                     progress.report({ message: message });
                     ext.outputChannel.appendLog(message);
 
-                    runRequest.sourceLocation = 'a'
                     context.run = await context.client.registries.beginScheduleRunAndWait(context.resourceGroupName, context.registryName, runRequest);
 
                 },
--- a/src/commands/image/imageSource/buildImageInAzure/RunStep.ts
+++ b/src/commands/image/imageSource/buildImageInAzure/RunStep.ts
@@ -5,9 +5,11 @@
 
 import { type DockerBuildRequest as AcrDockerBuildRequest } from "@azure/arm-containerregistry";
 import { AzExtFsExtra, GenericTreeItem } from "@microsoft/vscode-azext-utils";
+import * as retry from 'p-retry';
 import * as path from 'path';
 import { type Progress } from "vscode";
 import { activityFailContext, activityFailIcon } from "../../../../constants";
+import { ext } from "../../../../extensionVariables";
 import { ExecuteActivityOutputStepBase, type ExecuteActivityOutput } from "../../../../utils/activity/ExecuteActivityOutputStepBase";
 import { createActivityChildContext } from "../../../../utils/activity/activityUtils";
 import { localize } from "../../../../utils/localize";
@@ -28,10 +30,18 @@ export class RunStep extends ExecuteActivityOutputStepBase<BuildImageInAzureImag
                 dockerFilePath: path.basename(context.dockerfilePath) /* Assume the dockerfile is always in the root of the source */
             };
 
-            const building: string = localize('buildingImage', 'Building image...');
-            progress.report({ message: building });
-
-            context.run = await context.client.registries.beginScheduleRunAndWait(context.resourceGroupName, context.registryName, runRequest);
+            const retries = 3;
+            await retry(
+                async (currentAttempt: number) => {
+                    context.run = await context.client.registries.beginScheduleRunAndWait(context.resourceGroupName, context.registryName, runRequest);
+                    const message: string = currentAttempt === 1 ?
+                        localize('buildingImage', 'Building image...') :
+                        localize('buildingImageAttempt', 'Building image (Attempt {0}/{1})...', currentAttempt, retries + 1);
+                    ext.outputChannel.appendLog(message);
+                    progress.report({ message: message });
+                },
+                { retries, minTimeout: 2 * 1000 }
+            );
         } finally {
             if (await AzExtFsExtra.pathExists(context.tarFilePath)) {
                 await AzExtFsExtra.deleteResource(context.tarFilePath);
--- a/src/commands/image/imageSource/buildImageInAzure/RunStep.ts
+++ b/src/commands/image/imageSource/buildImageInAzure/RunStep.ts
@@ -32,13 +32,16 @@ export class RunStep extends ExecuteActivityOutputStepBase<BuildImageInAzureImag
 
             const retries = 3;
             await retry(
-                async (currentAttempt: number) => {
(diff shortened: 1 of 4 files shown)


src/commands/image/imageSource/buildImageInAzure/RunStep.ts
```diff
@@ -28,10 +30,18 @@ export class RunStep extends ExecuteActivityOutputStepBase<BuildImageInAzureImag
...
-            progress.report({ message: building });
-
-            context.run = await context.client.registries.beginScheduleRunAndWait(context.resourceGroupName, context.registryName, runRequest);
+            const retries = 3;
+            await retry(
+                async (currentAttempt: number) => {
+                    context.run = await context.client.registries.beginScheduleRunAndWait(context.resourceGroupName, context.registryName, runRequest);
+                    const message: string = currentAttempt === 1 ?
+                        localize('buildingImage', 'Building image...') :
+                        localize('buildingImageAttempt', 'Building image (Attempt {0}/{1})...', currentAttempt, retries + 1);
```
nturinski, line 39: Why's it retries + 1?

motm32: Techinically the first try is not a retry, so if there are 3 retries allowed that is technically the fourth try. If you think we should only do 3 tries I can change the retries to 2.

nturinski: Oh yeah, makes sense. I feel like 3 is a good number to me.

MicroFish91:
> Is there a way for us to parse the error that's coming from the logs? I remember that there was a specific error about the url being incorrect that Anthony was running into.
> 
> I'm think that we may want to check for that error specifically because otherwise the retry logic might be annoying if it's failing for totally different reasonsl

Yeah that's a good idea.  I know we talked briefly about seeing if we could parse errors from the logs for other reasons as well.  Would it be worth it for us to reach out to Julius' team to get a better understanding about how the log messages are formatted under different error states?

MicroFish91:
> Have we done a test run on how long it takes to retry 3 times? while retrying, do we show any UI indications?

I believe Megan is showing some UI indications in the activity log.  

I've tried throwing a fake error for testing purposes to simulate the retries - under this condition the entire process happens fairly quickly (~10-20s); however, we probably won't get a realistic understanding of the true time until we are able to find someone who can actually repro this error consistently without throwing fake ones.  Maybe we can reach out to Anthony later to test since he ran into this issue before.

> Is there a way for us to parse the error that's coming from the logs? I remember that there was a specific error about the url being incorrect that Anthony was running into.

I was looking into this issue with Megan, and I think we might not even need to parse the logs after all.  The error seems to throw specifically when the url is correct or not ready yet.  If, say, the Dockerfile is incorrect and will eventually result in a fail, that will not necessarily throw in that step because the run merely starts there, and until we poll it again during `BuildImageStep`, we won't know whether or not the build failed.  So basically, if we see an error throw, it's probably safe to assume the URL was the issue without any additional parsing whatsoever.

src/commands/image/imageSource/buildImageInAzure/RunStep.ts
```diff
@@ -28,10 +30,18 @@ export class RunStep extends ExecuteActivityOutputStepBase<BuildImageInAzureImag
...
             };
 
-            const building: string = localize('buildingImage', 'Building image...');
-            progress.report({ message: building });
-
-            context.run = await context.client.registries.beginScheduleRunAndWait(context.resourceGroupName, context.registryName, runRequest);
+            const retries = 3;
+            await retry(
+                async (currentAttempt: number) => {
+                    context.run = await context.client.registries.beginScheduleRunAndWait(context.resourceGroupName, context.registryName, runRequest);
```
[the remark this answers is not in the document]
MicroFish91: I think repositioning the messages up will likely allow you to avoid adding the `onFailedAttempt` callback after all.  That along with the findings in the comments below :)

motm32:
Yes Matt is correct the Activity log does show an indication of each attempt like this: 
image

The Azure Container Apps logs also contain this information: 
image
I also checked and the retries also only look about 10 seconds to complete for me.

src/commands/image/imageSource/buildImageInAzure/RunStep.ts
```diff
@@ -27,10 +29,21 @@ export class RunStep extends ExecuteActivityOutputStepBase<BuildImageInAzureImag
...
-            const building: string = localize('buildingImage', 'Building image...');
-            progress.report({ message: building });
+            const retries = 3;
+            await retry(
+                async (currentAttempt: number): Promise<void> => {
+                    const message: string = currentAttempt === 1 ?
+                        localize('buildingImage', 'Building image...') :
+                        localize('buildingImageAttempt', 'Building image (Attempt {0}/{1})...', currentAttempt, retries + 1);
+                    progress.report({ message: message });
+                    ext.outputChannel.appendLog(message);
```
MicroFish91, line 39: How do you feel if we only append log once at the end with a summary of the actions taken rather than sending it the same report updates?  

I think the other log messages in the series follow that format:
image

That way the activity log progress can assume responsibility for the continuous "live" updates and the log can output a summary of what happened as well as any other important information that the user may need to resolve issues or gain more info.

nturinski: I'm okay with the live updates in the activity log, but I think appending the log might be important too. If the user is only looking at the output channel, it may appear that the operation froze while it's retrying.

MicroFish91: Totally cool with showing attempts in the output log.  I think my main issue is against writing out `Building image...` immediately followed by the `Finished building...`.  At a minimum, I would prefer omitting `Building image...` on the first pass so this doesn't happen --  could also maybe play with the `attempt` wording to make sure it matches the style of the rest of the log:

image

src/commands/image/imageSource/buildImageInAzure/RunStep.ts
```diff
@@ -27,10 +29,21 @@ export class RunStep extends ExecuteActivityOutputStepBase<BuildImageInAzureImag
...
 
-            context.run = await context.client.registries.beginScheduleRunAndWait(context.resourceGroupName, context.registryName, runRequest);
+                    runRequest.sourceLocation = 'a'
```
nturinski, line 41: What's this all about?

motm32: That is an oopsie by me. I was using this to make the run fail.

nturinski (approved):

MicroFish91 (approved):

PR #585, opened 2024-01-08
Add `deployWorkspaceProjectInternal`

MicroFish91:
Add internal implementation which will be shared by the API command and the normal version of the command.

Helps to decouple this implementation before setting up the API in this PR:
https://github.com/microsoft/vscode-azurecontainerapps/pull/578


nturinski (approved):
Nice. This is going to be so much cleaner with the API entry-point PR.

PR #588, opened 2024-01-12
Add JSON validation support for editing container app envelopes

### LINKED ISSUE #569: Add `Intellisense` support for editing container app envelopes

MicroFish91:
Closes #569


nturinski (approved):
Nice, love that everything has descriptions now:
image

The name kind of looks like ass now though: 
image

I think it should have a user-friendly name, but we use the above convention as the id or something?

Also a side tangent, but do you know if it'd be possible to give the user auto-complete with intellisense for the values? It's pretty tough to know that I have to enter 1Gi for `memory`, for instance.

Edit: Actually, I see that some of them do provide that (offering an empty array, for example). I'm not super familiar with the schema, but is there a way to offer some default values without it being an enum? Like, (1Gi, 2Gi for example, but it's not an enum of all of the `memory` capacities.)

MicroFish91:
> Edit: Actually, I see that some of them do provide that (offering an empty array, for example). I'm not super familiar with the schema, but is there a way to offer some default values without it being an enum? Like, (1Gi, 2Gi for example, but it's not an enum of all of the memory capacities.)

Hmm, I bet there is, do you think it's worth it for us to make the schema more specific than the one exposed by the ARM Schema repo?

nturinski:
> Yeah I agree, the name looks pretty bad especially due to how long it is. I'm not very clever with naming 😂, do you happen to have any good naming ideas for this, it needs to be a naming convention that maintains uniqueness across all subscription/rg/ca-name combos for lookup. It'd be cool to do this in a way that's independent from the existing resource names and IDs, but I'm not exactly sure how to do that for this.

I believe that the label doesn't have to be unique. The file collision is based on the full path, which should just be the id that we already show.

Container app named naturins-test5
image

Two container apps named the same thing:
image

If you hover, you can see the path is different:
image

MicroFish91:
Semi-fixed by adding the standard path separators back in.

PR #589, opened 2024-01-12
Convert errors to use `GenericParentTreeItem` and add more specific context values

MicroFish91:
Update activity log error items to use `GenericParentTreeItem` so that any additional error messages can be automatically appended as a child of the step that threw it.

Example:
image

### DIFF
--- a/src/commands/image/imageSource/buildImageInAzure/UploadSourceCodeStep.ts
+++ b/src/commands/image/imageSource/buildImageInAzure/UploadSourceCodeStep.ts
@@ -22,12 +22,13 @@ export class UploadSourceCodeStep extends ExecuteActivityOutputStepBase<BuildIma
     private _sourceFilePath: string;
 
     protected async executeCore(context: BuildImageInAzureImageSourceContext, progress: Progress<{ message?: string | undefined; increment?: number | undefined }>): Promise<void> {
-        context.registryName = nonNullValue(context.registry?.name);
-        context.resourceGroupName = getResourceGroupFromId(nonNullValue(context.registry?.id));
-        context.client = await createContainerRegistryManagementClient(context);
         /* relative path of src folder from rootFolder and what gets deployed */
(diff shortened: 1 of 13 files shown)


src/commands/image/imageSource/buildImageInAzure/UploadSourceCodeStep.ts
```diff
@@ -22,13 +22,14 @@ export class UploadSourceCodeStep extends ExecuteActivityOutputStepBase<BuildIma
...
 
     protected async executeCore(context: BuildImageInAzureImageSourceContext, progress: Progress<{ message?: string | undefined; increment?: number | undefined }>): Promise<void> {
-        context.registryName = nonNullValue(context.registry?.name);
-        context.resourceGroupName = getResourceGroupFromId(nonNullValue(context.registry?.id));
-        context.client = await createContainerRegistryManagementClient(context);
         /* relative path of src folder from rootFolder and what gets deployed */
         this._sourceFilePath = path.dirname(path.relative(context.rootFolder.uri.path, context.dockerfilePath));
         context.telemetry.properties.sourceDepth = this._sourceFilePath === '.' ? '0' : String(this._sourceFilePath.split(path.sep).length);
 
+        context.registryName = nonNullValue(context.registry?.name);
```
nturinski, line 29: Any particular reason why you moved this? Not a big deal, but wondering if it was an incidental change.

nturinski (approved):

src/commands/image/imageSource/buildImageInAzure/UploadSourceCodeStep.ts
MicroFish91: Moved it because if an error throws before that line is set, the error message at the end will display `undefined`.  I don't know if awaiting the client can ever throw so it may be a non-issue, but I think it's just a little safer to move that to execute first, just in case.

PR #592, opened 2024-02-02
Expand retries to the entire ACR image run/build process

MicroFish91:
WIP...

In an effort to solve the elusive `failed to download` ACR build issue...

So far my changes involve:
- Consolidating the existing `RunStep` and `BuildImageStep` into one step so we can wrap with retries (it probably makes sense to consolidate regardless though)
- Retry the run whenever we detect `Failed to download` in the logs as we suspect the url may not always be ready by the time the run request has been issued

Unfortunately, there was no obvious way to query or download from the blob directly due to auth restrictions on native ACR blob.


src/commands/image/imageSource/buildImageInAzure/BuildImageStep.ts
```diff
@@ -3,60 +3,121 @@
...
-                    await openAcrBuildLogs(context, this.acrBuildError);
-                }
-            });
+            // Schedule the ACR build image run
+            try {
+                acrRunId = (await context.client.registries.beginScheduleRunAndWait(context.resourceGroupName, context.registryName, runRequest)).runId;
+                this.acrRunCount++;
+            } catch (e) {
+                // No known benefits to enforcing retries at this stage-- only check for retries if we can interpret specific actions from the build logs
+                await this.removeTarFile(context);
```
nturinski, line 56: Don't we want to remove this on success as well? I think we should put it in a `finally`.

MicroFish91: Yeah, I was kind of hoping to find a way to not have to introduce the mega try/finally wrapper 😂

PR #594, opened 2024-02-06
Update the tar build logic

MicroFish91:


src/commands/image/imageSource/buildImageInAzure/UploadSourceCodeStep.ts
```diff
@@ -36,7 +35,7 @@ export class UploadSourceCodeStep extends ExecuteActivityOutputStepBase<BuildIma
         let items = await AzExtFsExtra.readDirectory(source);
         items = items.filter(i => !vcsIgnoreList.includes(i.name));
 
-        tar.c({ cwd: source }, items.map(i => i.name)).pipe(fse.createWriteStream(context.tarFilePath));
+        await tar.c({ cwd: source, file: context.tarFilePath }, items.map(i => path.relative(source, i.fsPath)));
```
nturinski, line 38: I wonder if we'll need a `writeStream` for deploying to Oryx++.

I do think in general, it'd probably be better to just be able to use a stream rather than compressing locally and then uploading. Not sure if ACR actually accepts a stream or not.

But I'm not trying to give you a heart attack; if this is working, let's just leave as is... for now. 😈

MicroFish91: > I do think in general, it'd probably be better to just be able to use a stream rather than compressing locally and then uploading. Not sure if ACR actually accepts a stream or not.

Not sure if they do, I believe the CLI also builds a tar and uploads as well.  But yeah once we migrate to Oryx++ hopefully we can just do as you mentioned :)

> But I'm not trying to give you a heart attack; if this is working, let's just leave as is... for now. 😈

All I can say is that now I've got a tarrible headache 😭

nturinski (approved):

PR #601, opened 2024-02-09
Automatically remove the `--platform` flag from the Dockerfile on deployment

### LINKED ISSUE #598: Remove build platform flag from Dockerfile prior to deploying
ACR builds are unable to handle the build platform flag commonly found in Dockerfiles. Since it's needed for some local containers (like if you're running on a Mac, for instance) we can't just delete the flag.

We should be able to copy the Dockerfile contents, remove the flag, and add that to the compressed tar file rather than using the Dockerfile selected by the user.

MicroFish91:
Closes #598 

ACR does not support the `--platform` flag and will automatically error out if it detects this flag.  They have an open issue for it, but we have no idea when it will be addressed and enough people are running into it that we have been asked to help address this on our end.

Updated activity log:
image

Big thanks and credit out to Anthony Chu for helping us come up with the bulk of the core logic for custom Dockerfile replacement! :)

### DIFF
--- a/src/commands/image/imageSource/buildImageInAzure/UploadSourceCodeStep.ts
+++ b/src/commands/image/imageSource/buildImageInAzure/UploadSourceCodeStep.ts
@@ -110,12 +110,12 @@ export class UploadSourceCodeStep<T extends BuildImageInAzureImageSourceContext>
         let loadMoreChildrenImpl: (() => Promise<AzExtTreeItem[]>) | undefined;
         if (this._customDockerfileDirPath) {
             loadMoreChildrenImpl = () => {
-                const removePlatformSuccessItem = new GenericTreeItem(undefined, {
-                    contextValue: createActivityChildContext(['removePlatformSuccessItem']),
+                const removePlatformFlagItem = new GenericTreeItem(undefined, {
+                    contextValue: createActivityChildContext(['removePlatformFlagItem']),
                     label: localize('removePlatformFlag', 'Remove unsupported ACR "--platform" flag'),
                     iconPath: new ThemeIcon('dash', new ThemeColor('terminal.ansiWhite')),
                 });
-                return Promise.resolve([removePlatformSuccessItem]);
+                return Promise.resolve([removePlatformFlagItem]);
             };
         }
 
--- a/src/commands/image/imageSource/buildImageInAzure/UploadSourceCodeStep.ts
+++ b/src/commands/image/imageSource/buildImageInAzure/UploadSourceCodeStep.ts
@@ -60,7 +60,7 @@ export class UploadSourceCodeStep<T extends BuildImageInAzureImageSourceContext>
                 await AzExtFsExtra.deleteResource(this._customDockerfileDirPath, { recursive: true });
             } catch {
                 // Swallow error, don't halt the deploy process just because we couldn't delete the temp files, provide a warning instead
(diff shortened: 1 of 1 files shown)


alexweininger:
> They have an open issue for it, but we have no idea when it will be addressed and enough people are running into it that we should probably try to help address this on our end.

Can you link said issue?

MicroFish91:
Here is the open issue on ACR... looks like it might be going stale lol

https://github.com/Azure/acr/issues/697

src/commands/image/imageSource/buildImageInAzure/UploadSourceCodeStep.ts
```diff
@@ -35,7 +40,31 @@ export class UploadSourceCodeStep extends ExecuteActivityOutputStepBase<BuildIma
...
+
+            // Append/Overwrite the original Dockerfile with the custom one that was made
+            await tar.r({ cwd: this._customDockerfileDirPath, file: tempTarFilePath }, [path.relative(source, context.dockerfilePath)]);
+
+            // Create the final compressed version
+            // Note: Noticed some hanging issues when using the async version to add existing tar archives;
+            // however, the issues seem to disappear when utilizing the sync version
+            tar.c({ cwd: tmpdir(), gzip: true, sync: true, file: context.tarFilePath }, [`@${path.basename(tempTarFilePath)}`]);
+
+            try {
```
nturinski, line 57: Is this try/catch supposed to only be in the if clause? I thought we need to delete the tarFile that we create regardless of the custom logic.

src/commands/image/imageSource/buildImageInAzure/UploadSourceCodeStep.ts
```diff
@@ -35,7 +40,31 @@ export class UploadSourceCodeStep extends ExecuteActivityOutputStepBase<BuildIma
         let items = await AzExtFsExtra.readDirectory(source);
         items = items.filter(i => !vcsIgnoreList.includes(i.name));
 
-        await tar.c({ cwd: source, gzip: true, file: context.tarFilePath }, items.map(i => path.relative(source, i.fsPath)));
+        await this.buildCustomDockerfileIfNecessary(context);
```
nturinski, line 43: Maybe this should be in a the try/catch too? I don't know; if this fails, maybe it's not worth trying the build with their dockerfile anyway?

src/commands/image/imageSource/buildImageInAzure/UploadSourceCodeStep.ts
MicroFish91: I believe the tarFile deleting comes in the `runStep` immediately after.

MicroFish91: In general I have some thoughts about/want some feedback regarding the try/catch structure though.  I'll reach out to discuss more offline.

src/commands/image/imageSource/buildImageInAzure/UploadSourceCodeStep.ts
MicroFish91: There's a few points here that an error could be thrown that I haven't wrapped with try/catch to preserve readability and because I don't necessarily have an elegant solution for error handling in mind.  Let's discuss this offline as well

nturinski (approved):

nturinski (approved):

PR #602, opened 2024-02-13
Fix default name validation logic for `deployWorkspaceProject`

### LINKED ISSUE #596: Deploying from a folder with an invalid name fails
1. Open a folder named `20230206-my-app`.
2. Create a simple app with dockerfile.
3. Deploy project from workspace.
4. It should fail because it tries to create a container app with a name that starts with a number.

Should try to ensure the name is a valid ACA name.

MicroFish91:
Don't allow names to start with a number.

Closes #596

### DIFF
--- a/src/commands/deployWorkspaceProject/getDefaultValues/DefaultResourcesNameStep.ts
+++ b/src/commands/deployWorkspaceProject/getDefaultValues/DefaultResourcesNameStep.ts
@@ -66,18 +66,15 @@ export class DefaultResourcesNameStep extends AzureWizardPromptStep<DeployWorksp
             (!context.containerApp && !context.newContainerAppName);
     }
 
-    private validateInput(name: string | undefined): string | undefined {
-        name ??= '';
-
+    private validateInput(name: string = ''): string | undefined {
         // No symbols are allowed for ACR - we will strip out any offending characters from the base name, but still need to ensure this version has an appropriate length
         const nameWithoutSymbols: string = name.replace(/[^a-z0-9]+/g, '');
         if (!validateUtils.isValidLength(nameWithoutSymbols, 5, 20)) {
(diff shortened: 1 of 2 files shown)


src/commands/deployWorkspaceProject/getDefaultValues/DefaultResourcesNameStep.ts
```diff
@@ -66,18 +66,15 @@ export class DefaultResourcesNameStep extends AzureWizardPromptStep<DeployWorksp
...
         // No symbols are allowed for ACR - we will strip out any offending characters from the base name, but still need to ensure this version has an appropriate length
         const nameWithoutSymbols: string = name.replace(/[^a-z0-9]+/g, '');
         if (!validateUtils.isValidLength(nameWithoutSymbols, 5, 20)) {
             return localize('invalidLength', 'The alphanumeric portion of the name must be least 5 characters but no more than 20 characters.');
         }
 
-        const symbols: string = '-';
-        if (!validateUtils.isLowerCaseAlphanumericWithSymbols(name, symbols, false /** canSymbolsRepeat */)) {
-            return validateUtils.getInvalidLowerCaseAlphanumericWithSymbolsMessage(symbols);
+        if (!/^[a-z][a-z0-9]*(-[a-z0-9]+)*$/.test(name)) {
```
nturinski, line 76: suggestion: You could make the name step's `validateInput` method static and just use it here.

nturinski (approved):

src/commands/deployWorkspaceProject/getDefaultValues/DefaultResourcesNameStep.ts
MicroFish91: Unfortunately I can't just re-use the method because it has different length requirements.  However, I can make the RegExp itself static, let me know if that sounds like a good alternative!

PR #604, opened 2024-02-15
Update `.nvmrc` to 18.15 and update VS Code version to match

MicroFish91:


nturinski:
Do we need the `build.yml` in this repo too?

MicroFish91:
I thought we didn't need it anymore after migrating to using the shared pipelines, see:
https://github.com/microsoft/vscode-azurecontainerapps/pull/394

nturinski (approved):

Issue #607, opened 2024-02-29
Monorepo: Expand support for deploying multiple app configurations within the same workspace project


MicroFish91:
**I. Overview**

At the current time we only support saving & deploying from one workspace app configuration.  To support saving and deploying from multiple app configurations, we need to migrate to a new schema and add additional capabilities as shown below.

**II. Flow**
![image](https://github.com/microsoft/vscode-azurecontainerapps/assets/40250218/d5e60ff3-a7b3-4ad6-b03f-8ec14f389860)

**III. Schema**

Old:
![image](https://github.com/microsoft/vscode-azurecontainerapps/assets/40250218/9fa0b585-cdd9-4641-8cae-0c2b75518525)

New:
<img src="https://github.com/microsoft/vscode-azurecontainerapps/assets/40250218/e823cfaa-a87d-4d1d-80f6-8026515172fc" width="400px">

**IV. Work Items**

Deploy Project from Workspace (internal):
- [x] [Add support for taking an existing `envPath` and converting it to environment variables on the CA envelope](https://github.com/microsoft/vscode-azurecontainerapps/issues/640)
- [x] [Update the settings save logic to abide by the new schema](https://github.com/microsoft/vscode-azurecontainerapps/issues/608)
- [x] [Update default resource naming logic](https://github.com/microsoft/vscode-azurecontainerapps/issues/611)
- [x] [Decouple settings validation and parsing logic from `deployWorkspaceProjectInternal`](https://github.com/microsoft/vscode-azurecontainerapps/issues/610)

Monorepo:
- [x] [Add hover descriptions for the new json schemas](https://github.com/microsoft/vscode-azurecontainerapps/issues/645)
- [x] [Add ability to detect and convert old schema to new schema](https://github.com/microsoft/vscode-azurecontainerapps/issues/612) **(In Review)**
- [x] [Monorepo core logic](https://github.com/microsoft/vscode-azurecontainerapps/issues/613) 
See linked issue for additional task breakdown.
- [x] [Telemetry](https://github.com/microsoft/vscode-azurecontainerapps/issues/636)

Finishing touches (added 3/26):
- [x] Better default naming convention for each resource (e.g. `-rg`, `-me`, `-ca`)
- [x] Try to default container registry when a tree item is provided
- [x] Also try to default a container registry within the internal logic if one still hasn't been provided
- [x] Don't show a dropdown for loading workspace configuration in activity log if no children provided
[...]

PR #614, opened 2024-03-05
Housekeeping for `deployWorkspaceProject`

MicroFish91:
- Organized files into new folders:
  - `settings/`
  - `internal/`

- Wrapped existing settings logic under new V1 names (logic hasn't been changed at all).  Some of this logic will eventually be revisited/removed once we migrate to V2.

Eventual schema changes outlined [here](https://github.com/microsoft/vscode-azurecontainerapps/issues/607).


nturinski (approved):
Are you planning to make a `dwpSettingUtilsV2` file or just make new files (sort of like how `dwpSettingsUtilsV1` used to be)?

I'm asking because if we get rid of `dwpSettingsUtilsV1` at some point, it'll look really weird 🤣

MicroFish91:
> I'm asking because if we get rid of dwpSettingsUtilsV1 at some point, it'll look really weird 🤣
😂😂 I think we'll have to keep some concept of V1 and V2 so we'll probably be able to avoid that scenario

Issue #610, opened 2024-02-29
Decouple settings validation and parsing logic from `deployWorkspaceProjectInternal`


MicroFish91:
Work item related to the monorepo work: https://github.com/microsoft/vscode-azurecontainerapps/issues/607

Right now, settings validation is tightly coupled to `deployWorkspaceProjectInternal`.  We should remove as much of this coupling as we can and have it be the responsibility of the caller to read, verify, and provide any starter fields as inputs to the internal context.  

In our case, the caller would be the actual `deployWorkspaceProject` command.  If we ever support multiple deployments of apps at the same time, this will be an especially crucial decoupling that will need to take place, so best to do this as early as possible.

MicroFish91:
I think this is basically done now.

PR #615, opened 2024-03-05
Decouple workspace and tree item settings from `deployWorkspaceProjectInternal`

MicroFish91:
Closes https://github.com/microsoft/vscode-azurecontainerapps/issues/610

- Decouple workspace and tree item settings from `deployWorkspaceProjectInternal`
- Removed options from internal logic since we decoupled this (i.e. they no longer apply): `ignoreExistingDeploySettings`, `suppressProgress`.
- Increment api version


src/commands/deployWorkspaceProject/deployWorkspaceProject.ts
```diff
@@ -15,28 +15,39 @@ import { localize } from "../../utils/localize";
 import { type DeployWorkspaceProjectResults } from "../api/vscode-azurecontainerapps.api";
 import { browseContainerApp } from "../browseContainerApp";
 import { type DeployWorkspaceProjectContext } from "./DeployWorkspaceProjectContext";
+import { getDefaultContextValues } from "./getDefaultValues/getDefaultContextValues";
 import { getDeployWorkspaceProjectResults } from "./getDeployWorkspaceProjectResults";
 import { deployWorkspaceProjectInternal, type DeployWorkspaceProjectInternalContext } from "./internal/deployWorkspaceProjectInternal";
 
 export async function deployWorkspaceProject(context: IActionContext & Partial<DeployWorkspaceProjectContext>, item?: ContainerAppItem | ManagedEnvironmentItem): Promise<DeployWorkspaceProjectResults> {
+    const subscription: AzureSubscription = await subscriptionExperience(context, ext.rgApiV2.resources.azureResourceTreeDataProvider);
```
nturinski, line 23: Any reason for moving this? 

If we have an item, we can get the subscription information from that, so I feel like it should probably be below that check still. Granted, it doesn't look like we were doing that before, but we probably should have been.

src/commands/deployWorkspaceProject/getDefaultValues/getDefaultContextValues.ts
```diff
@@ -32,18 +32,16 @@ export async function getDefaultContextValues(
...
-            context.telemetry.properties.settingsOverride = 'triggered';
-            await displaySettingsOverrideWarning(context, item as ContainerAppItem | ManagedEnvironmentItem);
-            context.telemetry.properties.settingsOverride = 'accepted';
-        } else {
-            // No settings conflict
-            context.telemetry.properties.settingsOverride = 'none';
-            dwpSettingUtilsV1.displayDeployWorkspaceProjectSettingsOutput(settings);
-        }
+    // Logic to display local workspace settings related outputs
+    if (triggerSettingsOverride(settings, item)) {
```
nturinski, line 36: Do we still need this logic? I'm assuming that we won't ever overwrite workspace deployment settings anymore.

MicroFish91: Not really, but I'm keeping it here because it's necessary to make V1 work which is what is still hooked up until we finish V2 and swap it over :)

src/commands/deployWorkspaceProject/deployWorkspaceProject.ts
MicroFish91: Oh that's a good point, I'll move it back below.  And we can consider grabbing the subscription if an item is passed in for the V2 iteration

nturinski (approved):

PR #616, opened 2024-03-05
Scaffold outline of logic for `deployWorkspaceProjectV2`

MicroFish91:
Scaffold the `deployWorkspaceProjectV2` command which represents the new monorepo code path that is in development.  The V2 command represents an islanded segment of code that is not accessible to any users, so we don't need to worry about breaking any existing features.  Future monorepo related changes should be made towards this V2 command until we are feature complete.

I mocked up some todo functions with pre-defined interfaces to help us visualize all the tasks and how they fit together. 
The V2 suffix for this command will eventually be removed and will take the place of the existing `deployWorkspaceProject` once completed.


src/commands/deployWorkspaceProject/deploymentConfiguration/getTreeItemDeploymentConfiguration.ts
```diff
@@ -0,0 +1,12 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { type IContainerAppContext } from "../../IContainerAppContext";
+import { type DeploymentConfiguration } from "./DeploymentConfiguration";
+
+// Todo: Monorepo core logic (tree item path) https://github.com/microsoft/vscode-azurecontainerapps/issues/613
+export async function getTreeItemDeploymentConfiguration(_: IContainerAppContext): Promise<DeploymentConfiguration> {
```
MicroFish91, line 10: Whoever implements the logic here can actually worry about what inputs to pass here (like the tree item I'm assuming), I'm mostly just trying to define a flow and the general shape of output ahead of time.

src/commands/deployWorkspaceProject/deployWorkspaceProjectV2.ts
```diff
@@ -0,0 +1,93 @@
...
+        suppressConfirmation: false,
+        suppressContainerAppCreation: false,
+        suppressWizardTitle: false
+    });
+
+    displayNotification(deployWorkspaceProjectResultContext);
+    return await getDeployWorkspaceProjectResults(deployWorkspaceProjectResultContext);
+}
+
+function displayNotification(context: DeployWorkspaceProjectContext): void {
```
MicroFish91, line 65: This is all just carryover stuff from the existing `deployWorkspaceProject`

nturinski (approved):

Issue #613, opened 2024-02-29
Monorepo core logic


MicroFish91:
Work item related to the monorepo work: https://github.com/microsoft/vscode-azurecontainerapps/issues/607

See steps 3-4 [here](https://github.com/microsoft/vscode-azurecontainerapps/issues/607#issue-2160097230).

#### Workspace settings path (no tree item):
  Prompt:
  - [x] (1) Root folder step
  - [x] (2) Prompt for settings config

  Execute:
  - [x] (3) Shallow validation (~ensure file paths still exist) **(In Review)**
  - [x] (4) Deep validation (~ensure Azure resources still exist)
    - This will involve converting the existing `getDefaultContextValues` branch of logic into steps
    - This task may be able to be split down further, but it's hard to know until we start trying to implement/convert
  - [x] (5) ACR defaulting logic
    - Check existing monorepo app configurations, and try to default to an existing ACR rather than creating a new one when `context.registry` hasn't yet been populated

#### Tree item path:
  - [x] (6) Load tree items into context 
    - This work item had to be added since we needed to decouple the settings and tree loading from the deployment's internal logic

MicroFish91:
Now that I'm seeing the logic here, I think some of this could be split up into additional issues/PRs.  Based on the timing of tasks, we can revisit how to partition this work again at a later date.

MicroFish91:
Done!

PR #619, opened 2024-03-08
Add `DeploymentConfigurationListStep`

MicroFish91:
Step 2 in https://github.com/microsoft/vscode-azurecontainerapps/issues/613

Example:

image


src/commands/deployWorkspaceProject/deploymentConfiguration/workspace/DeploymentConfigurationListStep.ts
```diff
@@ -0,0 +1,59 @@
...
+            ]
+        };
+    }
+
+    private getPicks(deploymentConfigurations: DeploymentConfigurationSettings[]): IAzureQuickPickItem<DeploymentConfigurationSettings | undefined>[] {
+        const picks: IAzureQuickPickItem<DeploymentConfigurationSettings | undefined>[] = deploymentConfigurations.map(deploymentConfiguration => {
+            return {
+                label: deploymentConfiguration.label ?? localize('unnamedApp', 'Unnamed app'),
+                // Show the container app name as the description by default, unless the label has the same name
+                description: deploymentConfiguration.label === deploymentConfiguration.containerApp ? undefined : deploymentConfiguration.containerApp,
```
nturinski, line 47: Todo: Let's brainstorm some more ideas of what the description can contain. I still think that there are more useful things that we can display to the user.

src/commands/deployWorkspaceProject/deploymentConfiguration/workspace/DeploymentConfigurationListStep.ts
```diff
@@ -0,0 +1,59 @@
...
+                // Todo: Shallow (local fs) validation step(s)
+                // Todo: Deep (Azure resource) validation step(s)
+            ]
+        };
+    }
+
+    private getPicks(deploymentConfigurations: DeploymentConfigurationSettings[]): IAzureQuickPickItem<DeploymentConfigurationSettings | undefined>[] {
+        const picks: IAzureQuickPickItem<DeploymentConfigurationSettings | undefined>[] = deploymentConfigurations.map(deploymentConfiguration => {
+            return {
+                label: deploymentConfiguration.label ?? localize('unnamedApp', 'Unnamed app'),
```
nturinski, line 45: If there's no label, I think we should prompt for a label as part of the wizard.

Another idea; when we convert V1 settings to V2, the label should be called something like "Converted from "$APP_NAME" settings". In this case, we would also prompt them for the label.

nturinski (approved):
Just a few comments, but looks good to me

PR #623, opened 2024-03-14
Add validation for container app resources in `deployWorkspaceProjectV2`

MicroFish91:
Partially addresses step 4 of https://github.com/microsoft/vscode-azurecontainerapps/issues/613

Deep validation for container app resources (resource group / container app).  Same instructions to run, just swap `deployWorkspaceProject` to `deployWorkspaceProjectV2`.

image

image

image

It would be nice if there was an option we could pass that would automatically expand the activity node for us, otherwise the next activity can basically start and the user won't be able to see the nested tree items in the most recent activity run (which I think contextually would be very useful leading into the immediate execution of the next activity).  Maybe that's a todo item I need to add for the activity log updates

Edit: Updated the log item message slightly so it will look more congruent with the soon coming ACR verify step:

image


MicroFish91:
@nturinski Double check UI here.

nturinski (approved):

PR #626, opened 2024-03-17
Add ability to parse a `DeploymentConfiguration` from tree items in `deployWorkspaceProjectV2`

MicroFish91:
Addresses Step 6 of https://github.com/microsoft/vscode-azurecontainerapps/issues/613


src/commands/deployWorkspaceProject/deploymentConfiguration/treeItem/getTreeItemDeploymentConfiguration.ts
```diff
@@ -3,10 +3,53 @@
...
+import { localize } from "../../../../utils/localize";
 import { type IContainerAppContext } from "../../../IContainerAppContext";
+import { type DefaultContainerAppsResources } from "../../getDefaultValues/getDefaultContainerAppsResources/getDefaultContainerAppsResources";
 import { type DeploymentConfiguration } from "../DeploymentConfiguration";
 
-// Todo: Monorepo core logic (tree item path) https://github.com/microsoft/vscode-azurecontainerapps/issues/613
-export async function getTreeItemDeploymentConfiguration(_: IContainerAppContext): Promise<DeploymentConfiguration> {
-    return {};
+export async function getTreeItemDeploymentConfiguration(context: IContainerAppContext, item: ContainerAppItem | ManagedEnvironmentItem): Promise<DeploymentConfiguration> {
+    if (ContainerAppItem.isContainerAppItem(item)) {
```
MicroFish91, line 20: Do you think I should add a todo here for trying to obtain a default ACR, maybe checking to leverage any within the existing resource group?

src/commands/deployWorkspaceProject/deploymentConfiguration/treeItem/getTreeItemDeploymentConfiguration.ts
```diff
@@ -3,10 +3,53 @@
...
+    } else if (ManagedEnvironmentItem.isManagedEnvironmentItem(item)) {
+        return await getResourcesFromManagedEnvironmentItem(context, item.managedEnvironment);
+    } else {
+        const incompatibleMessage: string = localize('incompatibleTreeItem', 'An incompatible tree item was provided to Azure Container Apps for project deployment.');
+        ext.outputChannel.appendLog(localize('incompatibleMessageLog', 'Error: {0}', incompatibleMessage));
+        throw new Error(incompatibleMessage);
+    }
+}
+
+async function getResourcesFromContainerAppItem(context: ISubscriptionActionContext, containerApp: ContainerAppModel): Promise<DefaultContainerAppsResources> {
```
MicroFish91, line 31: Also, most of the logic in this page already existed and so has been reviewed before; I just repurposed it / moved it over from V1.

src/commands/deployWorkspaceProject/deploymentConfiguration/getTreeItemDeploymentConfiguration.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { ext } from "../../../extensionVariables";
+import { ContainerAppItem } from "../../../tree/ContainerAppItem";
+import { ManagedEnvironmentItem } from "../../../tree/ManagedEnvironmentItem";
+import { localize } from "../../../utils/localize";
+import { type DeploymentConfiguration } from "./DeploymentConfiguration";
+
+export async function getTreeItemDeploymentConfiguration(item: ContainerAppItem | ManagedEnvironmentItem): Promise<DeploymentConfiguration> {
+    if (ContainerAppItem.isContainerAppItem(item)) {
+        return { containerApp: item.containerApp };
```
nturinski, line 14: I think you should grab the managedEnvironment and any other configurations that we can grab from just the container item.

MicroFish91: I originally had this logic here, but I was thinking that this responsibility should really be handled in the internal logic.  I could still parse it here as well, definitely no harm in doing that, but since I refactored it to handle that automatically later, I'm wondering if we'd be okay just skipping the assignment here unless we have a particular need to parse it ahead of time.

Benefit, it's much more readable and the code is simpler here and will still work the same.

MicroFish91: FWIW the change happened a few PRs later so you probably didn't see the changes added until later 😅

But just for ref: https://github.com/microsoft/vscode-azurecontainerapps/pull/629

nturinski (approved):

Issue #451, opened 2023-09-15
Allow option to carry-over existing environment variables


MicroFish91:
We should talk about/explore a logic path that allows the user to carry over existing environment variables in the latest revision when deploying a new update.  Currently, "skip for now" would reset the next revision's environment variables even if some had previously existed.

![image](https://github.com/microsoft/vscode-azurecontainerapps/assets/40250218/4a0d9d21-09c5-4ed7-b885-bfe363149132)

See [this](https://github.com/microsoft/vscode-azurecontainerapps/pull/450#issuecomment-1721676972) for more context.

PR #627, opened 2024-03-18
Use existing env variables when updating container app if "skipped"

nturinski:
Fixes https://github.com/microsoft/vscode-azurecontainerapps/issues/451

Note that this only applies for `updateContainerApp` since `deployWorkspace` leverages `getDefaultContextValue` which sets `context.environmentVariables` to `[ ]` if it doesn't detect an `.env` file.

I didn't change this behavior because I figured that we will be altering `deployWorkspace`'s behavior; specifically we'll be looking at an `envPath` setting.

### DIFF
--- a/src/commands/image/imageSource/EnvironmentVariablesListStep.ts
+++ b/src/commands/image/imageSource/EnvironmentVariablesListStep.ts
@@ -44,11 +44,15 @@ export class EnvironmentVariablesListStep extends AzureWizardPromptStep<Environm
 
     private async selectEnvironmentSettings(context: EnvironmentVariablesContext): Promise<DotenvParseOutput | undefined> {
         const placeHolder: string = localize('setEnvVar', 'Select a {0} file to set the environment variables for the container instance', '.env');
+        // since we only allow one container, we can assume that we want the first container's env settings
+        const existingData: DotenvParseOutput | undefined = context.containerApp?.template?.containers?.[0].env as DotenvParseOutput | undefined;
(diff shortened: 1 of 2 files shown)


src/commands/image/imageSource/EnvironmentVariablesListStep.ts
```diff
@@ -44,11 +44,15 @@ export class EnvironmentVariablesListStep extends AzureWizardPromptStep<Environm
 
     private async selectEnvironmentSettings(context: EnvironmentVariablesContext): Promise<DotenvParseOutput | undefined> {
         const placeHolder: string = localize('setEnvVar', 'Select a {0} file to set the environment variables for the container instance', '.env');
+        // since we only allow one container, we can assume that we want the first container's env settings
+        const existingData: DotenvParseOutput | undefined = context.containerApp?.template?.containers?.[0].env as DotenvParseOutput | undefined;
```
MicroFish91, line 48: Is it safe to typecast to this? Wondering if the environment variables will always be converted properly when the value reaches this block?  (My hunch is that we'd end up with weird variables attached, like the array `length` property, but haven't tried testing it myself yet)

image

MicroFish91 (approved):

src/commands/image/imageSource/EnvironmentVariablesListStep.ts
MicroFish91: Didn't realize prototype properties don't get iterated over; looks good

PR #629, opened 2024-03-19
Infer and add any missing starting resources to context for `deployWorkspaceProjectInternal`

MicroFish91:
Related https://github.com/microsoft/vscode-azurecontainerapps/issues/610

Logic in a nutshell is basically...

If a container app is provided, automatically extract the resource group and managed environment if they aren't already populated and set the context with any missing values.

If a managed environment is provided, automatically extract the resource group and set if not already populated.


src/commands/deployWorkspaceProject/internal/startingConfiguration/containerAppsResourceHelpers.ts
```diff
@@ -0,0 +1,43 @@
...
+import { type ContainerAppModel } from "../../../../tree/ContainerAppItem";
+import { createContainerAppsAPIClient } from "../../../../utils/azureClients";
+
+interface ContainerAppsResources {
+    resourceGroup?: ResourceGroup;
+    managedEnvironment?: ManagedEnvironment;
+    containerApp?: ContainerAppModel;
+}
+
+export async function getResourcesFromContainerAppHelper(context: ISubscriptionActionContext, containerApp: ContainerAppModel): Promise<ContainerAppsResources> {
```
MicroFish91, line 19: Brought over this existing logic from `getDefaultContextValues`...

src/commands/deployWorkspaceProject/internal/startingConfiguration/getStartingConfiguration.ts
```diff
@@ -41,3 +42,16 @@ export async function getStartingConfiguration(context: DeployWorkspaceProjectIn
         environmentVariables: await EnvironmentVariablesListStep.workspaceHasEnvFile() ? undefined : [], // Todo: revisit this
     };
 }
+
+async function tryAddMissingAzureResourcesToContext(context: DeployWorkspaceProjectInternalContext): Promise<void> {
+    if (!context.containerApp && !context.managedEnvironment) {
+        return;
+    } else if (context.containerApp) {
+        const resources = await getResourcesFromContainerAppHelper(context, context.containerApp);
```
nturinski, line 50: Can we get the ACR from the container as well? I figured the image name would have it.

nturinski (approved):

src/commands/deployWorkspaceProject/internal/startingConfiguration/getStartingConfiguration.ts
MicroFish91: Yeah, potentially.  I think that logic is complicated enough plus we might have to investigate if that's a good UX so I'll leave that investigation for a different PR/discussion

PR #631, opened 2024-03-19
Add `DwpManagedEnvironmentListStep` to `getStartingConfiguration`

MicroFish91:
Related https://github.com/microsoft/vscode-azurecontainerapps/issues/610


nturinski (commented):
Probably need to talk to Anthony about the behavior when there is only 1 managed environment as well. Autoselect it maybe?

If we're displaying all MEs in subscription, I think we should display the resource group as a description or maybe as a group sort of like this:

image

MicroFish91:
Okay, I will add the group sorting and leave a todo for discussion with Anthony.

MicroFish91:
Discussion Notes:
We will make a uniform changes across the entire extension at a later date showing the newer resource grouping scheme rather.  Keeping this mostly as-is.  Added the resource group in the description for the time being if it has a different name from the environment name.

src/commands/deployWorkspaceProject/internal/startingConfiguration/DwpManagedEnvironmentListStep.ts
```diff
@@ -0,0 +1,74 @@
...
+                description: '',
+                data: undefined
+            },
+            ...managedEnvironments.map(env => {
+                const environmentName: string = nonNullProp(env, 'name');
+                const resourceGroupName: string = parseAzureResourceId(nonNullProp(env, 'id')).resourceGroup;
+
+                return {
+                    label: environmentName,
+                    description: environmentName === resourceGroupName ? undefined : resourceGroupName,
```
nturinski, line 62: Hmm, I think I'd never show the resource group as a description if you did `listByResourceGroup` since all of the managed environments are going to be in the same group anyway. It also may be confusing for users why some have descriptions and others don't.

I think I'd either not show it at all, or sort the managed environments first by resource group name and then by managed environment name and have the description shown for all of the environments.

nturinski (approved):

PR #632, opened 2024-03-19
Getting Started - Add a Workspace Project (Walkthrough)

MicroFish91:
First step in the getting started experience - looks like:

image

If workspace isn't empty, this displays:

image

I noticed the latest version of VS Code Insiders seems to not always load the markdown on the right, not sure what is causing this bug but let me know if you run into this too.  I may need to report it to VS Code.


MicroFish91:
@nturinski Double check this one please

motm32:
I have also run into the markdown on the right not showing up a couple of times so likely need to file an issue on VSCode if there isn't one already.

MicroFish91:
> I have also run into the markdown on the right not showing up a couple of times so likely need to file an issue on VSCode if there isn't one already.

I just updated to a new version of VS Code insiders today and this _may_ be fixed now.  Can you update and check if this resolves it for you as well please? :)

motm32:
Just checked and the markdown showed up for me!

motm32 (approved):

PR #633, opened 2024-03-19
Hook up v2 of `deployWorkspaceProject`

MicroFish91:
Since we're getting to the point of messing with the `deployWorkspaceProjectInternal` logic, there are certain parts of the old `deployWorkspaceProject` that are no longer going to interact the same.  I think we're at the point where we should just remove it and hook up the new V2 work now.

This involved completely deleting the old `getDefaultValues` directory/logic since it's not really applicable anymore.


src/commands/deployWorkspaceProject/deployWorkspaceProject.ts
```diff
@@ -25,25 +28,27 @@ export async function deployWorkspaceProject(context: IActionContext & Partial<D
...
-    });
-
-    const deployWorkspaceProjectInternalContext: DeployWorkspaceProjectInternalContext = Object.assign(context, {
-        ...defaultContextValues,
+    const containerAppContext: IContainerAppContext = Object.assign(context, {
         ...subscriptionContext,
         subscription
     });
 
+    let deploymentConfiguration: DeploymentConfiguration;
```
MicroFish91, line 38: All I did here was rename `deployWorkspaceProjectV2` to `deployWorkspaceProject`.  Since the old file shares this same name, it did a diff on this file, but it's really just a copy of the old v2 logic.

nturinski (approved):

Issue #611, opened 2024-02-29
Update default resource naming logic


MicroFish91:
Work item related to the monorepo work: https://github.com/microsoft/vscode-azurecontainerapps/issues/607

What to do:

- Remove the auto-naming of resources based on the workspace name.  All apps in the monorepo will share the same root workspace, so this no longer makes as much sense.  We can still try to suggest and pre-fill a name for the user on prompt, but the best method for this is still up for debate.

- Ensure naming logic works for different combinations of existing resource names.  I think this mostly already works, but I wanted to put in a note here to double check and improve as necessary.  As a reminder, when we go to validate existing resources, we could find any of the listed resources (resource group, container app, container registry, etc.) no longer existing.  We will have to create and name new resources for any that are missing.
  - Being able to smartly detect and use resource validate inputs would be nice, but may be a bit of a stretch goal for this first pass.  
  
    See this for more info: https://github.com/microsoft/vscode-azurecontainerapps/issues/603

Issue #603, opened 2024-02-14
Refactor `DefaultResourcesNameStep` to utilize each resource step's `validateInput`


MicroFish91:
We are currently checking specifically against the most restrictive resource naming conventions rather than checking the validate inputs from all the resource steps that may be utilized individually.  We should probably set it up so that we actually run the validate inputs from each resource step since they own this logic and we should only have to maintain this in one area.  

The underlying task:  We'll need to ensure a cohesive validation error message return across all the resources called.  This may involve ensuring all messages follow the same [general] format and that the most restrictive steps are called first in the list.

PR #634, opened 2024-03-19
Update `DefaultResourcesNameStep`

MicroFish91:
Closes https://github.com/microsoft/vscode-azurecontainerapps/issues/611
Closes https://github.com/microsoft/vscode-azurecontainerapps/issues/603

- Removes auto-naming of resources based on the workspace name and always prompts for a name
- Refactor to leverage each step's `validateInput` (now static methods)
- Remove tests that are no longer necessary since logic has become much simpler 🧹


MicroFish91:
@nturinski Double check this one please

src/commands/deployWorkspaceProject/internal/DefaultResourcesNameStep.ts
```diff
@@ -4,83 +4,72 @@
...
-            ext.outputChannel.appendLog(localize('usingWorkspaceName', 'Using workspace name "{0}" as the default for remaining resource creation.', workspaceName));
-        }
+        ext.outputChannel.appendLog(localize('usingResourceName', 'User provided the new resource name "{0}" as the default for resource creation.', resourceName))
 
-        !context.resourceGroup && (context.newResourceGroupName = workspaceName);
-        !context.managedEnvironment && (context.newManagedEnvironmentName = workspaceName);
-        !context.registry && (context.newRegistryName = await RegistryNameStep.tryGenerateRelatedName(context, workspaceName));
-        !context.containerApp && (context.newContainerAppName = workspaceName);
-        context.imageName = ImageNameStep.getTimestampedImageName(context.containerApp?.name || workspaceName);
+        !context.resourceGroup && (context.newResourceGroupName = resourceName);
```
nturinski, line 27: I still suggest we should add a suffix for each resource type.

`resourceName-rg`
`resourceName-me` or `resourceName-env`
`resourceName-app` or `resourceName-aca`

I think when you see them all in the resource group, it makes it a lot easier to distinguish what each resource is. I notice other tooling will do that as well.

nturinski (approved):
Rest looks good to me. So fresh, so clean

src/commands/deployWorkspaceProject/internal/DefaultResourcesNameStep.ts
MicroFish91: I like that idea, let me do that in a separate PR after we discuss what that convention should look like

PR #637, opened 2024-03-20
Update to `registry` param in the deployment configuration

MicroFish91:
The container registry is actually just called `registry` in most places across the extension / SDK, so let's keep that convention internally.

(We can keep the user settings showing as `containerRegistry` though because I think that looks much better for our users, lmk what you think)


nturinski (approved):
Yeah makes sense to me.

Issue #609, opened 2024-02-29
Leverage existing ACRs automatically if previous apps exist in saved deployment settings


MicroFish91:
Work item related to the monorepo work: https://github.com/microsoft/vscode-azurecontainerapps/issues/607

Before calling into `deployWorkspaceProjectInternal`, we should check to see if `containerApps.deploymentConfigurations` already contains an app config.  If it does, we can try to automatically leverage an existing ACR rather than forcing the user to create a new one.

PR #638, opened 2024-03-20
Leverage ACRs automatically if they exist in the saved deployment settings

MicroFish91:
Closes https://github.com/microsoft/vscode-azurecontainerapps/issues/609

During the loading workspace settings phase, if we can't find a context.registry, we try to look through local workspace settings and automatically set it with an available registry if possible.


src/commands/deployWorkspaceProject/deploymentConfiguration/workspace/TryUseExistingWorkspaceRegistryStep.ts
```diff
@@ -0,0 +1,75 @@
...
+        this.options.shouldSwallowError = true;
+
+        const settings: DeploymentConfigurationSettings[] | undefined = await dwpSettingUtilsV2.getWorkspaceDeploymentConfigurations(nonNullProp(context, 'rootFolder'));
+        if (!settings?.length) {
+            return;
+        }
+
+        progress.report({ message: localize('searchingAvailableRegistries', 'Searching for available registry...') });
+        const registries: Registry[] = await AcrListStep.getRegistries(context);
+        const registryMap: Map<string, Registry> = new Map();
```
MicroFish91, line 30: Used a map here to reduce operations to O(n).  Using a `.find` inside of a for loop would've been ~O(n^2).  Not sure if we really care here though 🤷‍♂️

nturinski: Yeah but that O(n) space you're allocating now 😮 

Haha, I don't have a strong preference either way. I think it makes sense to use a map here.

MicroFish91: Changing the logic quite a bit so no longer relevant as much, though still using a map :D

MicroFish91:
Ready for review!

src/commands/deployWorkspaceProject/deploymentConfiguration/workspace/TryUseExistingWorkspaceRegistryStep.ts
```diff
@@ -0,0 +1,60 @@
...
+export class TryUseExistingWorkspaceRegistryStep extends AzureWizardExecuteStep<WorkspaceDeploymentConfigurationContext> {
+    public priority: number = 220;  /** Todo: Figure out a good priority level */
+
+    public async execute(context: WorkspaceDeploymentConfigurationContext, progress: Progress<{ message?: string | undefined; increment?: number | undefined }>): Promise<void> {
+        const settings: DeploymentConfigurationSettings[] | undefined = await dwpSettingUtilsV2.getWorkspaceDeploymentConfigurations(nonNullProp(context, 'rootFolder'));
+        if (!settings?.length) {
+            return;
+        }
+
+        if (context.deploymentConfigurationSettings) {
```
nturinski, line 26: Is `TryUseExistingWorkspaceRegistryStep` intended to be used anywhere else besides when trying to get the workspace configuration? Wondering why we need this check.

src/commands/deployWorkspaceProject/deploymentConfiguration/workspace/TryUseExistingWorkspaceRegistryStep.ts
```diff
@@ -0,0 +1,60 @@
...
+
+            registryMap.set(registry.name, registry);
+        }
+
+        for (const setting of settings) {
+            if (!setting.containerRegistry) {
+                continue;
+            }
+
+            if (registryMap.has(setting.containerRegistry)) {
```
nturinski, line 49: One thing-- are we only using the name of the registry for the `setting.containerRegistry`? I forget exactly where we landed on this, but we may need the resource group as well. I _think_ it'd be okay to just include the whole id even with the sub id. We've been doing that for deployment settings in App Service for a while and it hasn't seemed to cause any issues.

nturinski (approved):
Just want to make sure that I am understanding the flow here.

When we're loading the configuration settings (so this is before the user is actually going to be prompted),

1. Check all deployment configuration settings.
2. Load all ACRs and make a map by `name`
3. Use first ACR and deployment configuration match

MicroFish91:
> Just want to make sure that I am understanding the flow here.
> 
> When we're loading the configuration settings (so this is before the user is actually going to be prompted),
> 
> 1. Check all deployment configuration settings.
> 2. Load all ACRs and make a map by `name`
> 3. Use first ACR and deployment configuration match

The flow for the entire `deployWorkspaceProject` command up until this point has been.

1. Check workspace settings.  If they exist display them as selectable options for the user.  Whatever gets selected gets loaded into `context.deploymentConfigurationSettings`.  If create is selected, or if no settings exist yet, then this value will be `undefined`.
2. Verify container app resources (if a setting was selected)
3. Verify container registry resource (if a setting was selected)
4. Now this step runs if we still don't have a registry.  This step will run for two reasons.  Either a setting wasn't selected (we're deploying a new app configuration), or the registry provided for the configuration doesn't exist anymore.
  a. Load all ACRs and make a map by name.  The map points to the registry resource.
  b. Use first ACR and deployment configuration match

src/commands/deployWorkspaceProject/deploymentConfiguration/workspace/TryUseExistingWorkspaceRegistryStep.ts
MicroFish91: Currently it's not, though I've been thinking about using this step elsewhere in a future PR.

However, `context.deploymentConfigurationSettings` can also be `undefined` when creating a new app configuration.  In such a scenario, we may still want to run this step because it could potentially reduce the number of prompts since the user might still have other deployments we can search through.

src/commands/deployWorkspaceProject/deploymentConfiguration/workspace/TryUseExistingWorkspaceRegistryStep.ts
MicroFish91: `setting.containerRegistry` only has the registry name.  Registries are globally unique I believe, so we probably don't need the resource group like we do for the container app.  I was trying to avoid showing the ID because I want people to be able to push these changes on GitHub and I don't want subscription IDs being exposed

PR #642, opened 2024-03-23
Scaffold `FilePathsVerifyStep`

MicroFish91:
@motm32 This is the step that will eventually need to be finished for shallow validation.

The part I did was to set the context values if they exist.  This is because I need the context values to be populated to be able to start testing the full flow of the command.

I did not yet add any actual "verification" logic to the step, so you should add that part and any associated success/fail output as necessary.  Feel free to change up the existing logic as necessary


src/commands/deployWorkspaceProject/deploymentConfiguration/workspace/DeploymentConfigurationListStep.ts
```diff
@@ -38,7 +39,7 @@ export class DeploymentConfigurationListStep extends AzureWizardPromptStep<Works
 
         return {
             executeSteps: [
-                // Todo: Shallow (local fs) validation step(s)
+                new FilePathsVerifyStep(),
```
nturinski, line 42: It may be a bit redundant, but should this be broken into three separate steps?

The main reason being it would easier to exclude looking for a dockerfile path in the future. It could also be easier/more streamlined to tell the user _which_ one failed, etc. 

It'd be easy enough to create a base step and then just extend that since all of the logic should nearly be identical.

MicroFish91: Yeah that works too, can we merge this step for now so I can keep testing the logic in later PRs?  Megan can just remove this step in her PR(s) later.

MicroFish91: I might also re-explore exploding out the container app verifications then to see if we can make it a flat list.  I'll probably need to make two steps instead of one.

motm32 (approved):

Issue #39, opened 2022-03-14
Add a getting started walkthrough


anthonychu:

nturinski:
@anthonychu Do we have anybody that could help us with that? 

The getting started walkthrough for Azure Resources was written by @fiveisprime and he was able to work with a designer to get us some really classy images.

PR #653, opened 2024-03-26
Getting Started: Clean Up Resources (Walkthrough)

MicroFish91:
Closes https://github.com/microsoft/vscode-azurecontainerapps/issues/39

image


MicroFish91:
I'm wondering if we should manually add our own custom entry-point for this walkthrough somewhere.  I feel like currently it's not going to be very discoverable.  We can add a link in the readme, but wondering if we can add it to a more traveled area in the editor. 🤔

PR #654, opened 2024-03-26
Alternate walkthrough entry-point

MicroFish91:
What do you all think about an alternate entry-point for better discoverability?

Updated to `Open Walkthrough`

image


MicroFish91:
Discussing the idea of showing on first run of the command**

MicroFish91:
> I personally find it really confusing. I think I'd prefer the command to be called (Open Walkthroughs...) and be at the bottom. But even then, I'm still a little bit 😬 about it

Destroying my hopes and dreams, one comment at a time 😢

Lol jk

nturinski (approved):

PR #656, opened 2024-03-27
Misc. `getWorkspaceDeploymentConfiguration` improvements

MicroFish91:


src/commands/deployWorkspaceProject/deploymentConfiguration/workspace/DeploymentConfigurationListStep.ts
```diff
@@ -37,6 +37,9 @@ export class DeploymentConfigurationListStep extends AzureWizardPromptStep<Works
             return undefined;
         }
 
+        // We mainly want to show activity children if there are deployment settings to verify
+        context.activityChildren ??= [];
```
MicroFish91, line 41: Only showing the activity children when we know there are steps to verify will prevent an empty expandable tree item from being shown in the activity log.

nturinski (approved):

PR #658, opened 2024-03-27
Add shallow validation

motm32:
Add checks for file paths set in the settings.json to ensure those file exist within the current workspace. 

Part of #613


src/commands/deployWorkspaceProject/deploymentConfiguration/workspace/FilePathsVerifyStep.ts
```diff
@@ -3,50 +3,71 @@
...
-        progress.report({ message: localize('verifyingFilePaths', 'Verifying file paths...') });
+        progress.report({ message: localize('verifyingFilePaths', `Verifying file paths...`) });
 
         const rootPath: string = nonNullProp(context, 'rootFolder').uri.fsPath;
 
-        const configDockerfilePath = context.deploymentConfigurationSettings?.dockerfilePath;
-        if (!context.dockerfilePath && configDockerfilePath) {
-            context.dockerfilePath = path.join(rootPath, configDockerfilePath);
-        }
+        this.configPath = nonNullValueAndProp(context.deploymentConfigurationSettings, this.key as keyof typeof context.deploymentConfigurationSettings)
```
MicroFish91, line 30:
```suggestion
        this.configPath = nonNullValueAndProp(context.deploymentConfigurationSettings, this.key)
```

I think we can add in built-in type safety so we don't need to type cast this.  Here's what you can do:

`abstract key: keyof DeploymentConfigurationSettings;`

Then you can just reference the implementing class' property like this:

`key = 'envPath' as const;` and it'll check for type safety automatically!

I think I could incorporate something similar into the Azure resource PR as well so that'll be a nice addition

src/commands/deployWorkspaceProject/deploymentConfiguration/workspace/FilePathsVerifyStep.ts
```diff
@@ -3,50 +3,71 @@
...
-        // Todo: Finish this, if necessary
-        return {};
+        if (!this.configPath || this.configPath === '') {
+            return {};
+        }
+
+        return {
+            item: new GenericTreeItem(undefined, {
+                contextValue: createActivityChildContext(['filePathVerifyStepSuccessItem', activitySuccessContext]),
+                label: localize('verifyPath', 'Verify {0} path', this.fileType),
```
MicroFish91, line 56: Would love to also see the file path here as well

Example:
`Verify src path "src"`
`Verify dockerfile path "src/Dockerfile"`

motm32: I thought about that but then went against it since the path name could be long. It's unlikely it would be but might look a little ugly if it is.

MicroFish91: Yeah that's fair.  I'll leave this one up to you to make the final call then 👍

MicroFish91: Yeah you're right, it could look busy if the user has long path names.  Let's keep it as is then :)

MicroFish91 (approved):
LGTM!!

PR #659, opened 2024-03-27
Automatically try to use a workspace ACR even when deploying to a tree item

MicroFish91:


src/commands/deployWorkspaceProject/deploymentConfiguration/getTreeItemDeploymentConfiguration.ts
```diff
@@ -3,21 +3,35 @@
...
+    await wizard.prompt();
+    await wizard.execute();
+
+    return {
+        rootFolder: wizardContext.rootFolder,
+        managedEnvironment: ManagedEnvironmentItem.isManagedEnvironmentItem(item) ? (item as ManagedEnvironmentItem).managedEnvironment : undefined,
+        containerApp: ContainerAppItem.isContainerAppItem(item) ? (item as ContainerAppItem).containerApp : undefined,
+        registry: wizardContext.registry,
+
+        // If it's a container app item, safe to assume it's a re-deployment, so don't re-prompt to save
```
nturinski, line 33: Is this technically true? Could have a container app from another tool (ie the portal) that we haven't deployed before.

nturinski (approved):
Beside my comment, rest of it looks good.

src/commands/deployWorkspaceProject/deploymentConfiguration/getTreeItemDeploymentConfiguration.ts
MicroFish91: Not a container app, but an actual tree item that gets passed to the command.

I believe the only time this command would ever receive a `ContainerAppItem` is when `Deploy Workspace to Container App` is executed; so I think the assumption still applies.  Let me know if you can think of a condition otherwise.

PR #660, opened 2024-03-28
Add fallback defaulting logic for ACRs within `deployWorkspaceProjectInternal`

MicroFish91:
The internal logic presents the last bastion of hope for our users to find a suitable ACR without prompting.  We should do everything we can to assist them out on the cloudy battlefield ☁⚔️

### DIFF
--- a/src/commands/deployWorkspaceProject/internal/startingConfiguration/TryUseExistingRegistryStep.ts
+++ b/src/commands/deployWorkspaceProject/internal/startingConfiguration/TryUseExistingRegistryStep.ts
@@ -34,7 +34,7 @@ export class TryUseExistingRegistryStep extends AzureWizardExecuteStep<DeployWor
         }
 
         // Prioritize trying to find a registry in the same resource group
-        // Otherwise, just find the first available registry
+        // Otherwise, just use the first available registry
         context.registry = registryInSameResourceGroup || registry;
     }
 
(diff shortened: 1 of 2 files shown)


src/commands/deployWorkspaceProject/internal/startingConfiguration/TryUseExistingRegistryStep.ts
```diff
@@ -0,0 +1,42 @@
...
+import { AcrListStep } from "../../../image/imageSource/containerRegistry/acr/AcrListStep";
+import { type DeployWorkspaceProjectInternalContext } from "../DeployWorkspaceProjectInternalContext";
+
+export class TryUseExistingRegistryStep extends AzureWizardExecuteStep<DeployWorkspaceProjectInternalContext> {
+    public priority: number = 200;  /** Todo: Figure out a good priority level */
+
+    public async execute(context: DeployWorkspaceProjectInternalContext): Promise<void> {
+        const registries: Registry[] = await AcrListStep.getRegistries(context);
+
+        let resourceGroupRegistry: Registry | undefined;
```
nturinski, line 18: Might want to either

a. Change this to be called `registryInSameResourceGroup`
b. Have a comment explaining the preferential registry is one in the same resource group, and then just any registry in that sub

I was confused why you had two separate variables and why you didn't break as soon as you found a registry initially.

nturinski (approved):

PR #662, opened 2024-03-28
Try to add default name suffixes to resources in `deployWorkspaceProjectInternal`

MicroFish91:
If able, will attach suffixes to names (e.g. `-ca`, `-env`, `-rg`, `-log`, etc.)

image

### DIFF
--- a/src/commands/createManagedEnvironment/LogAnalyticsCreateStep.ts
+++ b/src/commands/createManagedEnvironment/LogAnalyticsCreateStep.ts
@@ -19,7 +19,7 @@ export class LogAnalyticsCreateStep extends ExecuteActivityOutputStepBase<IManag
     protected async executeCore(context: IManagedEnvironmentContext, progress: Progress<{ message?: string | undefined; increment?: number | undefined }>): Promise<void> {
         const opClient = await createOperationalInsightsManagementClient(context);
         const resourceGroup = nonNullProp(context, 'resourceGroup');
-        const workspaceName = nonNullProp(context, 'newManagedEnvironmentName');
+        const workspaceName = context.newLogAnalyticsWorkspaceName || nonNullProp(context, 'newManagedEnvironmentName');
 
         const creating: string = localize('creatingLogAnalyticsWorkspace', 'Creating log analytics workspace...');
         progress.report({ message: creating });
@@ -36,10 +36,10 @@ export class LogAnalyticsCreateStep extends ExecuteActivityOutputStepBase<IManag
         return {
             item: new GenericTreeItem(undefined, {
                 contextValue: createActivityChildContext(['logAnalyticsCreateStepSuccessItem', activitySuccessContext]),
-                label: localize('createWorkspace', 'Create log analytics workspace "{0}"', context.newManagedEnvironmentName),
+                label: localize('createWorkspace', 'Create log analytics workspace "{0}"', context.newLogAnalyticsWorkspaceName || context.newManagedEnvironmentName),
                 iconPath: activitySuccessIcon
             }),
-            message: localize('createLogAnalyticsWorkspaceSuccess', 'Created log analytics workspace "{0}".', context.newManagedEnvironmentName)
+            message: localize('createLogAnalyticsWorkspaceSuccess', 'Created log analytics workspace "{0}".', context.newLogAnalyticsWorkspaceName || context.newManagedEnvironmentName)
         };
     }
 
@@ -47,10 +47,10 @@ export class LogAnalyticsCreateStep extends ExecuteActivityOutputStepBase<IManag
         return {
             item: new GenericParentTreeItem(undefined, {
                 contextValue: createActivityChildContext(['logAnalyticsCreateStepFailItem', activityFailContext]),
-                label: localize('createWorkspace', 'Create log analytics workspace "{0}"', context.newManagedEnvironmentName),
+                label: localize('createWorkspace', 'Create log analytics workspace "{0}"', context.newLogAnalyticsWorkspaceName || context.newManagedEnvironmentName),
                 iconPath: activityFailIcon
             }),
-            message: localize('createLogAnalyticsWorkspaceFail', 'Failed to create log analytics workspace "{0}".', context.newManagedEnvironmentName)
+            message: localize('createLogAnalyticsWorkspaceFail', 'Failed to create log analytics workspace "{0}".', context.newLogAnalyticsWorkspaceName || context.newManagedEnvironmentName)
         };
     }
 }
--- a/src/commands/deployWorkspaceProject/internal/DefaultResourcesNameStep.ts
+++ b/src/commands/deployWorkspaceProject/internal/DefaultResourcesNameStep.ts
@@ -34,7 +34,7 @@ export class DefaultResourcesNameStep extends AzureWizardPromptStep<DeployWorksp
             asyncValidationTask: (name: string) => this.validateNameAvailability(context, name)
         })).trim();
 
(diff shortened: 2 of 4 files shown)


src/commands/createManagedEnvironment/LogAnalyticsCreateStep.ts
```diff
@@ -19,7 +19,7 @@ export class LogAnalyticsCreateStep extends ExecuteActivityOutputStepBase<IManag
     protected async executeCore(context: IManagedEnvironmentContext, progress: Progress<{ message?: string | undefined; increment?: number | undefined }>): Promise<void> {
         const opClient = await createOperationalInsightsManagementClient(context);
         const resourceGroup = nonNullProp(context, 'resourceGroup');
-        const workspaceName = nonNullProp(context, 'newManagedEnvironmentName');
+        const workspaceName = context.newLogAnalyticsWorkspaceName || nonNullProp(context, 'newManagedEnvironmentName');
```
nturinski, line 22: Should we fallback to the `newManagedEnvironmentName`? If there isn't a `newLogAnalyticsWorkspaceName`, couldn't that mean that a log analytics workspace wasn't actually created?

src/commands/deployWorkspaceProject/internal/DefaultResourcesNameStep.ts
```diff
@@ -119,4 +136,16 @@ export class DefaultResourcesNameStep extends AzureWizardPromptStep<DeployWorksp
             return undefined;
         });
     }
+
+    private async checkSuffixNamesAvailable(context: DeployWorkspaceProjectInternalContext, resourceName: string): Promise<boolean> {
```
nturinski, line 140: I think it makes more sense to set the names in the context or returning an object with the default names rather than returning a boolean. I'd expect the behavior to be that it would set the names that are available and only falling back to the resource name if it wasn't.

This current behavior is all or nothing.

src/commands/deployWorkspaceProject/internal/DeployWorkspaceProjectSaveSettingsStep.ts
```diff
@@ -27,7 +28,7 @@ export class DeployWorkspaceProjectSaveSettingsStep extends ExecuteActivityOutpu
         const deploymentConfigurations: DeploymentConfigurationSettings[] = await dwpSettingUtilsV2.getWorkspaceDeploymentConfigurations(rootFolder) ?? [];
 
         const deploymentConfiguration: DeploymentConfigurationSettings = {
-            label: context.configurationIdx !== undefined && deploymentConfigurations?.[context.configurationIdx].label || context.containerApp?.name,
+            label: context.configurationIdx !== undefined && deploymentConfigurations?.[context.configurationIdx].label || removeCaSuffixIfExists(nonNullValueAndProp(context.containerApp, 'name')),
```
nturinski, line 31: Might be cleaner to just save the `resourceName` on the context instead of doing this kind of wonky thin with the containerApp name.

src/commands/deployWorkspaceProject/internal/DefaultResourcesNameStep.ts
MicroFish91: Yeah the all or nothing was intentional, you think partial would be okay?

src/commands/createManagedEnvironment/LogAnalyticsCreateStep.ts
MicroFish91: I'm not sure I understand - are you asking in the sense that you don't think we should be falling back to the `newManagedEnvironmentName`?  Because I am falling back to that value, but as a `nonNullProp`

src/commands/deployWorkspaceProject/internal/DefaultResourcesNameStep.ts
nturinski: Personally thought it would be preferred to do partial, but if it was a design choice, I'll leave it to you.

src/commands/deployWorkspaceProject/internal/DeployWorkspaceProjectSaveSettingsStep.ts
MicroFish91: I think it's possible that the label may need to be set but there technically is no new resource being created.  So I don't think that necessarily always solves the problem of sometimes needing to check the container app name (or whichever existing resource you'd like to use to get the name).

MicroFish91: I'm also fine with making the label name just the container app name by default as well if the readability is suffering too much with this.  Tradeoffs I guess 😋

src/commands/deployWorkspaceProject/internal/DefaultResourcesNameStep.ts
MicroFish91: Changing it to allow partials!

src/commands/createManagedEnvironment/LogAnalyticsCreateStep.ts
nturinski: Sorry, yeah I meant more like should we be doing that because I don't think it makes sense for the reasons stated above.

MicroFish91: Yeah maybe not, but that's how it was already designed.  I didn't want to rework the entire managed environment/log analytics design.  Maybe that's a future todo though.

src/commands/deployWorkspaceProject/internal/DeployWorkspaceProjectSaveSettingsStep.ts
nturinski: > label may need to be set but there technically is no new resource being created

I can think of two cases when this happens
1. Converting v1 to v2 settings
2. Deploying to an existing container app for the first time

The `resourceName` context value might be helpful still, but we then fallback to the container app remove suffix behavior if it doesn't exist? But honestly, I'm fine either way.

nturinski (approved):

src/commands/deployWorkspaceProject/internal/DefaultResourcesNameStep.ts
```diff
@@ -29,12 +34,15 @@ export class DefaultResourcesNameStep extends AzureWizardPromptStep<DeployWorksp
             asyncValidationTask: (name: string) => this.validateNameAvailability(context, name)
         })).trim();
 
-        ext.outputChannel.appendLog(localize('usingResourceName', 'User provided the new resource name "{0}" as the default for resource creation.', resourceName))
+        ext.outputChannel.appendLog(localize('usingResourceName', 'User provided the new resource name "{0}" as the base for resource creation.', resourceName))
+
+        const { resourceGroupName, logAnalyticsWorkspaceName, managedEnvironmentName, containerAppName } = await this.getResourceNames(context, resourceName);
```
nturinski, line 39: This actually looks a lot cleaner imo 😊

MicroFish91: Agreed!

src/commands/deployWorkspaceProject/internal/DeployWorkspaceProjectSaveSettingsStep.ts
MicroFish91: And 3. someone accidentally deletes their settings label by editing it directly.

Seems unlikely most of the time, but I bet it will happen to someone eventually.

MicroFish91: How about I'll add the `resourceName` field onto the context if we end up needing this value in another place :)

PR #663, opened 2024-03-29
Improve output logs and wording consistency for `deployWorkspaceProject`

MicroFish91:


src/commands/deployWorkspaceProject/deployWorkspaceProject.ts
```diff
@@ -37,8 +39,13 @@ export async function deployWorkspaceProject(context: IActionContext & Partial<D
         subscription
     });
 
+    ext.outputChannel.appendLog(
+        formatSectionHeader(localize('preparingDeploymentConfiguration', 'Preparing workspace deployment configuration'))
```
nturinski, line 43: I would say anything that sounds like a long running action should end with a `...`

Verbs ending with "ing" tend to make me thing, on-going long running action.

src/commands/deployWorkspaceProject/deploymentConfiguration/workspace/DeploymentConfigurationListStep.ts
```diff
@@ -37,9 +38,17 @@ export class DeploymentConfigurationListStep extends AzureWizardPromptStep<Works
...
         if (!context.deploymentConfigurationSettings) {
+            ext.outputChannel.appendLog(localize('createNewAppConfiguration', 'User chose to create a new app configuration.'));
             return undefined;
         }
 
+        if (context.deploymentConfigurationSettings.label) {
+            ext.outputChannel.appendLog(localize('choseExistingConfiguration', 'User chose to load existing workspace deployment configuration "{0}".', context.deploymentConfigurationSettings.label));
+        } else {
+            ext.outputChannel.appendLog(localize('choseExistingConfiguration', 'User chose to load existing workspace deployment configuration.'));
+
```
nturinski, line 49: nit: Extra line break.

src/commands/deployWorkspaceProject/deployWorkspaceProject.ts
nturinski, line 43: Should we just start using the new localize package? It'll make it easier to convert, but also I find it annoying to always have to come up with a "localize title key" for each string we have.

src/commands/deployWorkspaceProject/formatSectionHeader.ts
```diff
@@ -0,0 +1,11 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+/**
+ * @returns Formatted text to look like a new section header for output channel display
+ */
+export function formatSectionHeader(text: string): string {
```
nturinski, line 9: Is this different from `wrapWithDashFormatting`?

EDIT:

**scrolls down** 

Oh.

src/commands/deployWorkspaceProject/internal/startingConfiguration/TryUseExistingRegistryStep.ts
```diff
@@ -37,6 +39,12 @@ export class TryUseExistingRegistryStep extends AzureWizardExecuteStep<DeployWor
             }
         }
 
+        if (registryInSameResourceGroup) {
+            ext.outputChannel.appendLog(localize('resourceGroupRegistry', 'Found an available registry "{0}" in the provided resource group "{1}".', registryInSameResourceGroup.name, context.resourceGroup?.name));
```
nturinski, line 43: Should these things being output also be displayed in the activity description? I feel like these progress messages are kind of useful (and it'd probably be worth to just make a helper function that outputs in the channel _and_ the activity notification)

My only concern is that it would flash by too quickly for users to see.

nturinski (approved):
Very verbose of you 🤣 

I think the more descriptive logging is useful though.

src/commands/deployWorkspaceProject/deployWorkspaceProject.ts
MicroFish91: I'll make it just "Prepare" then.  I'm trying to make it look more like a title/header than an active loading command.

MicroFish91: Yeah, I agree, I hate having to come up with new id's all the time.  Maybe an item for next release

src/commands/deployWorkspaceProject/internal/startingConfiguration/TryUseExistingRegistryStep.ts
MicroFish91: So we do eventually show which ACR was used or created.  I kind of think that's good enough for the activity log.  I was thinking activity log is more high level, and output log is more detailed.  I don't think the activity log needs to tell the user how we got the resource, just which resource we are using or creating?

PR #664, opened 2024-03-29
Miscellaneous improvements to `deployWorkspaceProject`

MicroFish91:
Added a couple things:

- Added the passing of `rootFolder` across the different function calls so we don't ever have to re-prompt
- Don't convert V1 to V2 if deployment configurations already exist.  Otherwise we might accidentally erase or alter their existing config in a way that they don't like.  This could happen if they are messing around with settings and accidentally add an older v1 setting.


src/commands/deployWorkspaceProject/settings/convertSettings/ConvertSettingsStep.ts
```diff
@@ -15,7 +15,9 @@ export class ConvertSettingsStep extends AzureWizardExecuteStep<ConvertSettingsC
 
     public async execute(context: ConvertSettingsContext): Promise<void> {
         const settingsContentsV1: DeployWorkspaceProjectSettingsV1 = await dwpSettingUtilsV1.getDeployWorkspaceProjectSettings(nonNullProp(context, 'rootFolder'));
-        if (!settingsContentsV1.containerAppResourceGroupName && !settingsContentsV1.containerAppName && !settingsContentsV1.containerRegistryName) {
+        const settingsContentsV2: DeploymentConfigurationSettings[] | undefined = await dwpSettingUtilsV2.getWorkspaceDeploymentConfigurations(nonNullProp(context, 'rootFolder'));
+
+        if (settingsContentsV2?.length || (!settingsContentsV1.containerAppResourceGroupName && !settingsContentsV1.containerAppName && !settingsContentsV1.containerRegistryName)) {
```
nturinski, line 20: > Don't convert V1 to V2 if deployment configurations already exist. Otherwise we might accidentally erase or alter their existing config in a way that they don't like. This could happen if they are messing around with settings and accidentally add an older v1 setting.

Shouldn't we still convert that V1 by just pushing a new configuration into their array?

That being said, I think the chances of them having a V1 and V2 configuration simultaneously is very unlikely and such an edge case that I'm fine with just skipping past it.

nturinski (approved):

src/commands/deployWorkspaceProject/settings/convertSettings/ConvertSettingsStep.ts
MicroFish91: That's the probably better way to do it, but I opted for the easier/lazier path here because of it being such an edge case.

I think it's okay if at that point can just be like:

image

PR #669, opened 2024-04-03
Find previously deployed to managed environment and recommend them

nturinski:
I toyed with the idea of using grouping instead of description. Not sure if we want to go that route though in case we started grouping things by resource group or something.

Also `supressedPersistence` because the (recently used) description looks kind of silly with recommended.

image

### DIFF
--- a/src/commands/deployWorkspaceProject/internal/startingConfiguration/DwpManagedEnvironmentListStep.ts
+++ b/src/commands/deployWorkspaceProject/internal/startingConfiguration/DwpManagedEnvironmentListStep.ts
@@ -26,9 +26,9 @@ export class DwpManagedEnvironmentListStep extends AzureWizardPromptStep<DeployW
 
         await this.setRecommendedPicks(context, picks);
         const pick = await context.ui.showQuickPick(picks, { placeHolder, suppressPersistence: true });
-        context.telemetry.properties.usedRecommendedEnv = pick.description === recommendedPickDescription ? 'true' : 'false';
+        context.telemetry.properties.usedRecommendedEnv = isRecommendedPick(pick) ? 'true' : 'false';
         context.telemetry.properties.recommendedEnvCount =
-            String(picks.reduce((count, p) => count + (p.description === recommendedPickDescription ? 1 : 0), 0));
+            String(picks.reduce((count, pick) => count + (isRecommendedPick(pick) ? 1 : 0), 0));
 
         const managedEnvironment: ManagedEnvironment | undefined = pick.data;
         if (!managedEnvironment) {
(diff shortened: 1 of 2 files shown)


MicroFish91:
> Also supressedPersistence because the (recently used) description looks kind of silly with recommended.
Lol that's funny I also added this same thing to my PR because it was bothering me too 😂

src/commands/deployWorkspaceProject/internal/startingConfiguration/DwpManagedEnvironmentListStep.ts
```diff
@@ -21,8 +24,13 @@ export class DwpManagedEnvironmentListStep extends AzureWizardPromptStep<DeployW
             return;
         }
 
+        await this.setRecommendedPicks(context, picks);
+        const pick = await context.ui.showQuickPick(picks, { placeHolder, suppressPersistence: true });
+        context.telemetry.properties.usedRecommendedEnv = pick.description === recommendedPickDescription ? 'true' : 'false';
+        context.telemetry.properties.recommendedEnvCount =
```
MicroFish91, line 30: Oh hi reduce 👋

src/commands/deployWorkspaceProject/internal/startingConfiguration/DwpManagedEnvironmentListStep.ts
```diff
@@ -66,4 +74,45 @@ export class DwpManagedEnvironmentListStep extends AzureWizardPromptStep<DeployW
...
+                        recommendedPick.description = recommendedPickDescription;
+                    }
+                }
+            }
+            catch (these_hands) {
+                // ignore the error and continue
+            }
+        }
+
+        const recommendedPick = (pick: IAzureQuickPickItem<ManagedEnvironment | undefined>): boolean => {
```
MicroFish91, line 100: This has the same name as the `recommendedPick` variable int he if block.  I know that version of the variable is block scope, but maybe a different name here is a good idea :)

Also, nit but a slightly shorter alternative:

```typescript
const recommendedPick = (pick: IAzureQuickPickItem<ManagedEnvironment | undefined>): boolean => pick.description === recommendedPickDescription;
```

MicroFish91 (approved):
LGTM!  My only question is if we should still show this to the user even when they've entered in through the API entry-point.

src/commands/deployWorkspaceProject/internal/startingConfiguration/DwpManagedEnvironmentListStep.ts
nturinski: Good call. I was doing a counter to keep track of how many were recommended before which is why I had the if/else.

PR #671, opened 2024-04-05
Add better default name prompt pre-fill values for `deployWorkspaceProjectInternal`

MicroFish91:


src/commands/deployWorkspaceProject/internal/AppResourcesNameStep.ts
```diff
@@ -22,6 +23,7 @@ export class AppResourcesNameStep extends AzureWizardPromptStep<DeployWorkspaceP
     public async prompt(context: DeployWorkspaceProjectInternalContext): Promise<void> {
         context.newContainerAppName = (await context.ui.showInputBox({
             prompt: localize('containerAppNamePrompt', 'Enter a name for the new container app'),
+            value: context.dockerfilePath?.split(path.sep).at(-2)?.replace(/[^a-zA-Z0-9-]/g, ''),
```
nturinski, line 26: Maybe just make that regex a constant? I thought you had a util function for removing any non-alphanumeric string too.

MicroFish91: Added the util function back, forgot I had made that before

nturinski (approved):

Issue #426, opened 2023-08-04
Add support for `containers`


MicroFish91:
- [x] Add `containers` support in the tree view (motm32)
- [x] Reconfigure the existing `deployImage` command to use the revision draft rather than automatically deploying a new revision

MicroFish91:
Megan merged the PR for this, closing

PR #673, opened 2024-04-05
Add container view to container apps 

motm32:
Part of #426 

Bringing this branch back from the dead 🧟

Here is how the view looks: 
image

At the moment we are only supporting displaying the information. In the future we hope to add support for adding/uploading environment variables. 

I also extended `RevisionDraftDescendantBase` instead of `ContainerAppsItem` so when we add further support it already basically has drafts/revisions support. Let me know if we want to change it to `ContainerAppsItem` instead.


MicroFish91:
This is looking really nice 😎, can't wait to try it out later

motm32:
This is how it looks in the case of multiple containers: 
image

src/tree/containers/EnvironmentVariableItem.ts
```diff
@@ -0,0 +1,44 @@
...
+    _hideValue: boolean;
+    constructor(public readonly subscription: AzureSubscription,
+        public readonly containerApp: ContainerAppModel,
+        public readonly revision: Revision,
+        readonly containerId: string,
+        readonly container: Container,
+        readonly envVariable: EnvironmentVar) {
+        this._hideValue = true;
+    }
+    id: string = `${this.containerId}/environmentVariable`
```
MicroFish91, line 25: The `id` here won't always be unique and could throw an error if you have multiple environment variables present.  Maybe adding on the `name` property at the end could alleviate this:

`${this.containerId}/environmentVariables/${this.envVariable.name}`

src/tree/containers/ContainersItem.ts
```diff
@@ -0,0 +1,89 @@
...
+export class ContainersItem extends RevisionDraftDescendantBase {
+    id: string;
+    label: string;
+    private containers: Container[] = [];
+
+    constructor(public readonly subscription: AzureSubscription,
+        public readonly containerApp: ContainerAppModel,
+        public readonly revision: Revision,) {
+        super(subscription, containerApp, revision);
+        this.id = `${containerApp.id}/containers`;
```
MicroFish91, line 31: This id may not always be unique, for example in multiple revisions mode.

You can probably fix this by doing something similar to what I did in the ID for `ScaleRuleItem`!

PR #674, opened 2024-04-10
Upgrade API version to 1.0.1

nturinski:
We should have just released the API version at 1.0.0 since 0.0.1 does some funky things with semver. The API itself isn't in preview (even if the extensions are) so there's really no reason to be using 0.0.1 anyway.


motm32 (approved):

MicroFish91 (approved):

MicroFish91:
Before you merge, could you also update the versions in the API changelog as well please :)

PR #691, opened 2024-06-13
Miscellaneous preparations for follow-up nightly tests

MicroFish91:
Figured it would make the follow-up PR a lot cleaner to review if we merge these changes in beforehand.


src/commands/image/imageSource/EnvironmentVariablesListStep.ts
```diff
@@ -84,8 +84,14 @@ export class EnvironmentVariablesListStep extends AzureWizardPromptStep<Environm
         return Object.keys(envData).map(name => { return { name, value: envData[name] } });
     }
 
-    public static async workspaceHasEnvFile(): Promise<boolean> {
-        const envFileUris: Uri[] = await workspace.findFiles(allEnvFilesGlobPattern);
+    public static async workspaceHasEnvFile(rootFolder?: WorkspaceFolder): Promise<boolean> {
```
MicroFish91, line 87: Added a `rootFolder` param to properly support multi-root workspaces.

src/commands/image/imageSource/buildImageInAzure/TarFileStep.ts
```diff
@@ -16,7 +16,7 @@ export class TarFileStep extends AzureWizardExecuteStep<BuildImageInAzureImageSo
     public async execute(context: BuildImageInAzureImageSourceContext): Promise<void> {
         const id: number = Math.floor(Math.random() * Math.pow(10, idPrecision));
         const archive = `sourceArchive${id}.tar.gz`;
-        context.tarFilePath = Utils.joinPath(URI.parse(os.tmpdir()), archive).fsPath;
+        context.tarFilePath = path.join(os.tmpdir(), archive);
```
MicroFish91, line 19: The old carryover path logic sometimes wouldn't work in the pipelines, this fixes that.

src/commands/deployWorkspaceProject/internal/SharedResourcesNameStep.ts
```diff
@@ -39,6 +39,10 @@ export class SharedResourcesNameStep extends AzureWizardPromptStep<DeployWorkspa
 
         !context.resourceGroup && (context.newResourceGroupName = resourceName);
         !context.managedEnvironment && (context.newManagedEnvironmentName = resourceName);
+
+        // The line of code that follows is only applicable during testing.
+        // This is because `newRegistryName` normally gets populated during validation; however, this gets bypassed during test mocks but not during normal execution.
+        !context.registry && !context.newRegistryName && (context.newRegistryName = await RegistryNameStep.tryGenerateRelatedName(context, resourceName));
```
alexweininger, line 45: I'd love to avoid this. Could you explain why the code that sets this is bypassed during testing?

alexweininger (approved):

src/commands/deployWorkspaceProject/internal/SharedResourcesNameStep.ts
MicroFish91: It's because when we set `context.ui = new TestUserInput(vscode);` before testing, the current behavior in the `dev` package is to skip the validation and just set the user input directly with the pre-defined input we provide.

In this step's current implementation, we are setting the `newRegistryName` during the validation phase for various reasons, but since this gets skipped during testing, we end up with a scenario where the field can be `undefined` where normally it would never be.

That said, I think we placed the name generation / setting within the validation logic in case the name generation failed to generate a unique id combination, in which case it would allow the user to try a new name.  Thinking on it some more though, I seriously doubt it's a real concern that a user would ever run into a situation where we were unable to actually generate a name.  Perhaps, given the above case we are now having to workaround during testing, it may just be better to pull that logic out into the prompt step directly and throw an error in the extremely rare case where we are unable to generate a proper name within the time allotted.  That way, we are not having to maintain the idea of an edge case path during testing.

alexweininger: That makes sense. The key here for me is that you are setting inside the validate method. I'm fine keeping that there, or moving it, that's up to you. If you decide to keep it there, I'd link your explanation in the existing comment.

alexweininger (approved):

PR #708, opened 2024-06-27
Add a failing test control case and update main tests to better support intentional fail cases

MicroFish91:
This PR adds an intentional fail test control case for `containerapps-albumapi-javascript`.  The failing test uses a bad Dockerfile to accomplish this.  Future test projects should be able to model off of this to accomplish something similar as well.


alexweininger (commented):
Is there an error we're looking for, or are we just checking if `deployWorkspaceProject` throws?

MicroFish91:
> Is there an error we're looking for, or are we just checking if `deployWorkspaceProject` throws?

With this setup we are only checking if `deployWorkspaceProject` throws.  The idea is that not having any `expectedResults` set is basically equivalent to an error being thrown and the command not being able to finish and return a result.

If more granular error checks are useful, we could definitely add something to the test case params and amend the try/catch a tiny bit to allow for that.  Let me know what you think!

alexweininger:
> If more granular error checks are useful

I'm mainly interested in if Azure gives us back an error like "invalid Dockerfile"? In these cases where is the Dockerfile being built? In Azure or locally? If it's in Azure I assume Azure gives back some sort of error.

MicroFish91:
Yeah, ACR should be throwing an error, let me make some adjustments and we can add a check for that

alexweininger (approved):
Awesome

PR #724, opened 2024-08-12
Add managed environments to command contexts in preparation for managed identity work

MicroFish91:
The managed identity connection flow will rely on granting `acrPull` access to the parent container app environment.  In preparation for this upcoming work, this PR introduces a required environment context to extend from, forcing major commands to set managed environments before starting each flow.

I considered including this work as part of the next PR for which it is most relevant, however, most of this work is just reworking existing context types and setting managed environments on command contexts... so I figured it would be a lot easier to review separately since this is mostly distracting side-work supporting the main implementation.


src/commands/ManagedEnvironmentContext.ts
```diff
@@ -0,0 +1,13 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { type ManagedEnvironment } from "@azure/arm-appcontainers";
+import { type ISubscriptionActionContext } from "@microsoft/vscode-azext-utils";
+import { type AzureSubscription } from "@microsoft/vscode-azureresources-api";
+
+export interface RequiredManagedEnvironmentContext extends ISubscriptionActionContext {
```
MicroFish91, line 10: I used the name "Required" here to help quickly identify that the managed environment resource is not optional.

src/commands/registryCredentials/identity/ManagedIdentityRegistryCredentialsContext.ts
```diff
@@ -0,0 +1,13 @@
...
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { type RegistryCredentials } from "@azure/arm-appcontainers";
+import { type IContainerAppContext } from "../../IContainerAppContext";
+import { type CreateAcrContext } from "../../image/imageSource/containerRegistry/acr/createAcr/CreateAcrContext";
+import { type RequiredManagedEnvironmentContext } from "../../ManagedEnvironmentContext";
+
+export interface ManagedIdentityRegistryCredentialsContext extends CreateAcrContext, RequiredManagedEnvironmentContext, IContainerAppContext {
```
MicroFish91, line 11: This flow has not technically been introduced yet, but I went ahead and set up the context shape so we can enforce the new type requirements.

src/commands/ManagedEnvironmentContext.ts
nturinski: Discussed offline, but I think we settled on using a naming convention that would put the resource ala `ManagedEnvironment` first.

src/commands/createContainerApp/CreateContainerAppContext.ts
```diff
@@ -14,9 +13,6 @@ import { type IngressBaseContext } from '../ingress/IngressContext';
 
 export interface CreateContainerAppBaseContext extends IResourceGroupWizardContext, ImageSourceBaseContext, IngressBaseContext, IContainerAppContext, ExecuteActivityContext {
```
nturinski, line 14: `ContainerAppCreateBaseContext`?

src/commands/createManagedEnvironment/CreateManagedEnvironmentContext.ts
```diff
@@ -9,13 +9,13 @@ import { type IResourceGroupWizardContext } from '@microsoft/vscode-azext-azureu
 import { type ExecuteActivityContext, type ISubscriptionActionContext } from '@microsoft/vscode-azext-utils';
 import { type AzureSubscription } from "@microsoft/vscode-azureresources-api";
 
-export interface IManagedEnvironmentContext extends ISubscriptionActionContext, IResourceGroupWizardContext, ExecuteActivityContext {
+export interface CreateManagedEnvironmentContext extends ISubscriptionActionContext, IResourceGroupWizardContext, ExecuteActivityContext {
```
nturinski, line 12: `ManagedEnvironmentCreateContext`?

MicroFish91:
Updated with feedback, this should be ready for review again

nturinski (approved):
LGTM

PR #753, opened 2024-09-20
Fix an issue with currently deployed recommendations

### LINKED ISSUE #746: Currently deployed description sometimes wrong for repositories
When choosing an image source, we try to indicate which quick picks match what is currently deployed so that users can quickly reselect registries and repositories.  Right now repository selection checks for a match without regarding the prior registry selection.  This can lead to a situation where we indicate that a repository with the same name is currently deployed, even if in a different registry.  We should check the registry even in the repository selection step to ensure this does not happen.

MicroFish91:
Fixes #746

### DIFF
--- a/src/commands/image/imageSource/containerRegistry/acr/AcrListStep.ts
+++ b/src/commands/image/imageSource/containerRegistry/acr/AcrListStep.ts
@@ -3,6 +3,7 @@
  *  Licensed under the MIT License. See License.txt in the project root for license information.
  *--------------------------------------------------------------------------------------------*/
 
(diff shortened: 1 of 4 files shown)


nturinski (approved):
I'm not really following why you got rid of the checks for the image name not being the quick start image though.

MicroFish91:
> I'm not really following why you got rid of the checks for the image name not being the quick start image though.

Because the quick start is actually from the MCR registry and therefore wouldn't show up as ACR or Docker anyway... so it was probably a redundant check

PR #754, opened 2024-09-20
Add a better error message when trying to expand an app that isn't provisioned yet

MicroFish91:
From:

image

To:

image

### DIFF
--- a/src/tree/ContainerAppItem.ts
+++ b/src/tree/ContainerAppItem.ts
@@ -93,7 +93,7 @@ export class ContainerAppItem implements ContainerAppsItem, RevisionsDraftModel 
                 this.containerApp.name);
 
             if (this.containerApp.provisioningState && this.containerApp.provisioningState !== 'Succeeded') {
-                throw new Error(localize('provisioningError', 'The container app "{0}" and its children cannot be accessed until the app is successfully provisioned.', this._containerApp.name));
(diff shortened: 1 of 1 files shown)


src/tree/ContainerAppItem.ts
```diff
@@ -92,6 +92,10 @@ export class ContainerAppItem implements ContainerAppsItem, RevisionsDraftModel
                 this.containerApp.id,
                 this.containerApp.name);
 
+            if (this.containerApp.provisioningState && this.containerApp.provisioningState !== 'Succeeded') {
```
nturinski, line 95: The error makes it sound like the user needs to do something. I assume that they just need to wait, right?

MicroFish91: It depends, if the provisioning is "InProgress" then they can just wait and try again.  But if the provisioning status is "Failed" they might need to do something

nturinski: Then may be worth telling the user what the state is and give them appropriate responses. Otherwise this error message isn't much more helpful.

MicroFish91: The provision state is shown in the tree item description, so I kind of assumed the user could figure it out pretty easily on their own, I was just trying to avoid the nonNull throw because that's truly confusing

nturinski: Yeah, I was just thinking we could change the error to something more like:

`The container app "{0}" cannot be expanded due to its provisioning state of "InProgress". Wait for provisioning to complete."

`The container app "{0}" cannot be expanded due to its provisioning state of "Failed". The provision must be rectified to view this resource."

nturinski (approved):

PR #758, opened 2024-09-27
Don't show `editScalingRange` on another `ScaleItem` if a draft item already exists

MicroFish91:
Fix an issue where `editScalingRange` was showing up on other revision items even when a draft node already existed.  In my opinion, it's confusing / doesn't make sense to allow users to try editing other revisions simultaneously when another draft already exists.

### DIFF
--- a/src/commands/scaling/scaleRange/editScaleRange.ts
+++ b/src/commands/scaling/scaleRange/editScaleRange.ts
@@ -10,7 +10,7 @@ import { type ScaleItem } from "../../../tree/scaling/ScaleItem";
 import { createActivityContext } from "../../../utils/activityUtils";
 import { localize } from "../../../utils/localize";
 import { pickScale } from "../../../utils/pickItem/pickScale";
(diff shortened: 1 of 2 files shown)


nturinski:
An alternative to this method is to keep showing the command, but throw an error when they try to edit another scale item that isn't the current draft.

My thoughts on why this might be less confusing is because sometimes users don't understand/remember why commands mysteriously disappeared on them. By having the error, it will remind the user that "oh, hey, I _do_ have a current working draft"

nturinski:
I'm okay with either approach, just wanted to hear thoughts

MicroFish91:
Yeah hmm, let me think on this one a bit.  I think I initially wanted to do it that way as well but for some reason I decided against it.  I'll get back to you

MicroFish91:
Couldn't remember a good reason not to do it the way you mentioned, I think it's a better solution as well because it's simpler for us to understand and now we can use these utility functions to easily add the same checks to other template items

nturinski (approved):

PR #769, opened 2024-10-11
Add `Edit Container...` command

MicroFish91:
Replace the existing `Update Container Image...` on the container app item with an `Update Container...` command now that the containers view has been added.

Before:
image

After:
image

Also adds support for the new activity log view.  Examples:

(Updating using existing image)
image

(Updating using workspace project and building image from scratch)
image

Update: Changed to use the `Edit Container...` instead of `Update Container...`


src/commands/updateContainer/updateContainer.ts
```diff
@@ -0,0 +1,70 @@
...
+import { ContainerUpdateDraftStep } from "./ContainerUpdateDraftStep";
+import { RegistryAndSecretsUpdateStep } from "./RegistryAndSecretsUpdateStep";
+
+// Updates both the 'image' and 'environmentVariables' portion of the container profile
+export async function updateContainer(context: IActionContext, node?: ContainersItem | ContainerItem): Promise<void> {
+    const item: ContainerItem | ContainersItem = node ?? await pickContainer(context, { autoSelectDraft: true });
+    const { containerApp, subscription } = item;
+
+    if (!isTemplateItemEditable(item)) {
+        throwTemplateItemNotEditable(item);
```
nturinski, line 29: nit: Very small nit-- kind of weird that this is a function and not just an error type.

src/utils/pickItem/pickContainer.ts
```diff
@@ -0,0 +1,94 @@
...
+        return false;
+    }
+
+    public async getSubWizard(context: T): Promise<IWizardOptions<T> | undefined> {
+        const lastNode: unknown = context.pickedNodes.at(-1);
+
+        let containers: Container[] = [];
+        if (ContainerAppItem.isContainerAppItem(lastNode)) {
+            const node = lastNode as ContainerAppItem;
+            containers = node.containerApp.template?.containers ?? [];
```
nturinski, line 77: I think for all of these, you can just do `lastNode.containerApp.template?.containers` since your static methods are asserting all of the types.

{E5402CBF-9169-4817-B443-1F05E31010EF}

src/commands/updateContainer/updateContainer.ts
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    const subscriptionContext: ISubscriptionContext = createSubscriptionContext(subscription);
+    const parentResource: ContainerAppModel | Revision = getParentResourceFromItem(item);
+
+    let containersIdx: number;
+    if (ContainersItem.isContainersItem(item)) {
+        // The 'updateContainer' command should only show up on a 'ContainersItem' when it only has one container, else the command would show up on the 'ContainerItem'
```
alexweininger, line 37: This is a nit, but if it's an easy change you might want to consider it.

I think that it might be better to always show the command on the Container**s**Item. Can we do that and when there are multiple containers just ask the user to pick one? You already have the pick container logic above, you might be able to reuse that.

I believe that users don't like when commands are missing, we want to be consistent so they can do things in the same way each time

nturinski (approved):
After seeing this quick pick in the UI, I feel like it may be kind of confusing.

{A34837F9-F9CD-44FB-A954-7A093A93F49C}

Also, it's a bit odd to me that we have a yes, no, and no, don't ask again. There is probably a scenario where someone would want yes, and don't ask again as well.

I'll try messing with the quick pick a bit to see if we can add a checkbox or button that could be the "don't ask again". Not blocking it over that, just something that I wanted to bring up.

src/commands/updateContainer/ContainerUpdateContext.ts
alexweininger, line 1: nit: the command is named "update container" while the context is the reverse, "container update context". If there's not a good reason for this, then I'd prefer if they matched.

package.json
```diff
@@ -203,6 +198,11 @@
                 "title": "%containerApps.openConsoleInPortal%",
                 "category": "Azure Container Apps"
             },
+            {
+                "command": "containerApps.updateContainer",
+                "title": "%containerApps.updateContainer%",
+                "category": "Azure Container Apps"
+            },
```
alexweininger, line 205: Don't want to block this PR from merging, but I think we need to discuss the possibility of rewording the command to make it more clear to users what's happening.

We need users to understand that there are pretty much two main actions for container apps:
1. Editing their Container App configuration (scale rules, etc.) - currently this is "Edit Container App..."
2. Updating the image running in their containers (within the Container App) - currently this is "Update Container..."

Nathan and I think that "Update Container..." might be too close to "Edit Container App...". What about using Deploy instead of Update?

Whatever we decide to do, we can make that change after this is merged.

alexweininger (approved):

MicroFish91:
> After seeing this quick pick in the UI, I feel like it may be kind of confusing.
> 
> {A34837F9-F9CD-44FB-A954-7A093A93F49C}
> 
> Also, it's a bit odd to me that we have a yes, no, and no, don't ask again. There is probably a scenario where someone would want yes, and don't ask again as well.
> 
> I'll try messing with the quick pick a bit to see if we can add a checkbox or button that could be the "don't ask again". Not blocking it over that, just something that I wanted to bring up.

Although a user could want yes and don't ask again, I would think we should naturally prevent this scenario as I think it's a bad pattern to enable and goes against the point of batching revision edits.

I do kind of agree that it looks a little funny though, but I'm not sure what the better alternative is yet,  may be worth discussing in our next meetup

src/commands/updateContainer/updateContainer.ts
MicroFish91: Yeah that's fair, would you rather I create a function that returns a string and throw that as the error message?  Or don't worry about standardizing the error message?

src/utils/pickItem/pickContainer.ts
MicroFish91: Oh true, I don't know why I thought I needed to assert the types like that...

src/commands/updateContainer/updateContainer.ts
MicroFish91: Yeah that's a good point, that's also how we have it implemented for picking revisions so this would help keep that consistent.  I will make a change in a follow-up PR to add this

src/commands/updateContainer/ContainerUpdateContext.ts
MicroFish91: I think I'm bouncing this one to @nturinski 🤭
I think he asked in a previous meeting people's preferences and was trying to improve naming conventions... but I can't remember who all was there

src/commands/updateContainer/updateContainer.ts
nturinski: You could do something like:

```
export class TemplateItemNotEditableError extends Error {
    constructor(item: RevisionsItemModel) {
        super(localize('itemNotEditable', 'Action cannot be performed on revision "{0}" because a draft is currently active.', item.revision.name));
    }
}
```

And then `throw new TempalteItemNotEditableError`

package.json
MicroFish91: Yup, I agree, and this is something I feel is super important to address.  Unfortunately, as of the time of this PR, we are still transitioning to the ideal end state.  I will try to remember to bring this one up in our next group meeting because I think it's a bit difficult to explain over a comment.

MicroFish91: I agree, I'll try to squeeze this discussion into one of our next group meetings (probably just the engineers for now). I think these sorts of improvements overlap a ton with my big goal for the next release, and I have some extra context to provide that would also be helpful for this.

PR #771, opened 2024-10-14
Change draft deploy pop-up from an informational window to a quick pick

### LINKED ISSUE #683: There is an extra step "Select subscription" when deploying changes by clicking "Yes" button from the prompt notification
**OS:** Win10
**Build Version:** 20240430.1
**Regression:** Not a regression 

**Repro Steps:**
1. Make sure more than one subscription is selected.
2. Execute "Edit Scaling Range..." action for a container app.
3. Click "Yes" button in the prompt notification "Deploy changes now?".
image
4. Check whether there is no extra step "Select subscription".

**Expect:**
There is no extra step "Select subscription".

**Actual:**
There is an extra step "Select subscription".
image

MicroFish91:
Discussed this with Nathan a while back, and we agreed it would probably be better to use a quick pick instead of a pop-up window after each command.  This means the prompt to deploy becomes something we ask at the start rather than something we ask after the draft command finishes.

I'm proposing the process now would be - each draft command should push in the following quick pick step as the last prompt:
`new RevisionDraftDeployPromptStep()`

Almost no 'new' logic has actually been added here, mostly prompt logic from the old `RevisionDraftUpdateBaseStep` got moved into the new `RevisionDraftDeployPromptStep`.

Before (informational window):
image

After (prompt quick pick):
image

To see in action, you can test with this command:
image

Edit: I forgot, I also snuck in a small fix for this...
Fixes #683

### DIFF
--- a/package.nls.json
+++ b/package.nls.json
@@ -4,7 +4,7 @@
     "containerApps.revealResource": "Reveal Resource",
     "containerApps.description": "An Azure Container Apps extension for Visual Studio Code.",
     "containerApps.enableOutputTimestamps": "Prepends each line displayed in the output channel with a timestamp.",
-    "containerApps.showDraftCommandDeployPopup": "Show an informational deploy pop-up message whenever a draft command is run.",
+    "containerApps.showDraftCommandDeployPopup": "Prompt to deploy revision draft whenever a draft command is run.",
     "containerApps.browse": "Browse",
     "containerApps.createContainerApp": "Create Container App...",
     "containerApps.createContainerAppFromWorkspace": "Create Container App from Workspace...",
--- a/src/commands/revisionDraft/RevisionDraftDeployPromptStep.ts
+++ b/src/commands/revisionDraft/RevisionDraftDeployPromptStep.ts
@@ -4,14 +4,14 @@
  *--------------------------------------------------------------------------------------------*/
 
 import { AzureWizardPromptStep, type IAzureQuickPickItem } from "@microsoft/vscode-azext-utils";
-import { showDraftCommandDeployPopupKey } from "../../constants";
+import { showDraftCommandDeployPopupSetting } from "../../constants";
 import { localize } from "../../utils/localize";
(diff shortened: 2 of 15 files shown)


package.nls.json
```diff
@@ -4,7 +4,7 @@
     "containerApps.revealResource": "Reveal Resource",
     "containerApps.description": "An Azure Container Apps extension for Visual Studio Code.",
     "containerApps.enableOutputTimestamps": "Prepends each line displayed in the output channel with a timestamp.",
-    "containerApps.showDraftCommandDeployPopup": "Show an informational deploy pop-up message whenever a draft command is run.",
```
alexweininger, line 7: is this enabled by default?

src/commands/revisionDraft/RevisionDraftDeployPromptStep.ts
```diff
@@ -0,0 +1,48 @@
...
+import { localize } from "../../utils/localize";
+import { settingUtils } from "../../utils/settingUtils";
+import { type RevisionDraftContext } from "./RevisionDraftContext";
+
+export class RevisionDraftDeployPromptStep<T extends RevisionDraftContext> extends AzureWizardPromptStep<T> {
+    public async prompt(context: T): Promise<void> {
+        if (!await settingUtils.getGlobalSetting(showDraftCommandDeployPopupSetting)) {
+            context.shouldDeployRevisionDraft = false;
+            return;
+        }
```
alexweininger, line 17: Should you try to move this into `shouldPrompt`? To do so, you could set this context before running the wizard.

package.nls.json
MicroFish91: Yeah true by default

src/commands/revisionDraft/RevisionDraftDeployPromptStep.ts
MicroFish91: Yeah good call

alexweininger (approved):

PR #778, opened 2024-10-23
Add an `Edit Container Image...` command

MicroFish91:
Basically the same as the `Update Container...` command, but skips the prompt to update environment variables.  

This is a more specific command to only update the image child only. 

image

Update: Changed to `Edit Container Image...`


src/utils/pickItem/pickImage.ts
```diff
@@ -0,0 +1,30 @@
...
+
+function getPickImageStep(): AzureWizardPromptStep<QuickPickWizardContext> {
+    return new ContextValueQuickPickStep(ext.rgApiV2.resources.azureResourceTreeDataProvider, {
+        contextValueFilter: { include: ImageItem.contextValueRegExp },
+        skipIfOne: true,
+    });
+}
+
+export async function pickImage(context: IActionContext, options?: RevisionDraftPickItemOptions): Promise<ImageItem> {
+    const containerAppItem: ContainerAppItem = await pickContainerApp(context);
```
nturinski, line 22: What do you think about leveraging the `getPickContainerAppSteps` instead and including it in the prompt steps here? The fact that this would end up being two separate wizards may be a bit jarring plus it means a user can't go back if they accidentally selected the wrong container.

nturinski (approved):
Similar to Alex's comment in your `Update Container...` PR, I think that this could benefit from some verbiage changes.

Specifically, I feel like `Update Container Image...` could be more easily conflated with actually updating the container app or people may think that they can only make minor edits to an existing image.

`Deploy Container Image...` or `Deploy new Container Image...` gives me the impression that I am overwriting whatever the current image is.

Not blocking for this, just noting that we should have a conversation about it.

MicroFish91:
> Similar to Alex's comment in your `Update Container...` PR, I think that this could benefit from some verbiage changes.
> 
> Specifically, I feel like `Update Container Image...` could be more easily conflated with actually updating the container app or people may think that they can only make minor edits to an existing image.
> 
> `Deploy Container Image...` or `Deploy new Container Image...` gives me the impression that I am overwriting whatever the current image is.

I'm thinking there may be some confusion by the image I posted in the previous PR, I will address this in our next meeting

src/utils/pickItem/pickImage.ts
MicroFish91: This is a good point, I'll do this instead

MicroFish91:
> Similar to Alex's comment in your `Update Container...` PR, I think that this could benefit from some verbiage changes.
> 
> Specifically, I feel like `Update Container Image...` could be more easily conflated with actually updating the container app or people may think that they can only make minor edits to an existing image.
> 
> `Deploy Container Image...` or `Deploy new Container Image...` gives me the impression that I am overwriting whatever the current image is.
> 
> Not blocking for this, just noting that we should have a conversation about it.

Changed to `Edit Container Image...` based on recent discussions

PR #785, opened 2024-11-01
Fix the showing of stale `View Properties` data

### LINKED ISSUE #330: The replicas value  doesn't auto refresh in the "xxx_Scaling.json" file after editing scaling range
**Environment:** VSCode.dev
**Container Apps Build Version:** v0.4.2-alpha.0
**Resource Group Build Version:** v0.6.3-alpha.0

**Repro Steps:**
1. Create a container app and expand it.
2. Right click "Scaling" node -> Edit Scaling Range… -> Set the range -> Enter.
3. Right click "Scaling" node -> View Properties.
4. Check whether the replicas value auto refreshes in the `xxx_Scaling.json `file.

**Expect:**
The replicas value auto refreshes in the `xxx_Scaling.json `file.

**Actual:**
The replicas value  doesn't auto refresh in the `xxx_Scaling.json `file.
image

**More Info:**
1. This issue also reproduces after executing "Add Scale Rule..." action.
2. This issue also reproduces after executing "Switch Ingress Visibility.../Edit Target Port...".
3. Reload VSCode can fix this issue.

MicroFish91:
Fixes #330 

After updates and refreshes, the View Properties data is often stale.  It seems that although the view itself is showing the correct and updated data, the tree item that is passed into the View Properties command is often a stale one.  

I looked but could not find what would be causing this to happen, however, there are some easy ways for me to help combat this in the ACA extension until we figure out a more robust solution.  To combat this locally, I have made a global resource cache that all relevant tree items would have reference to. The idea is that if a stale item is passed to the view properties command, the item now has a link to a global reference from which it can pull the latest data, independent from the state of the current item.  The cache is refreshed naturally with the latest data whenever we issue refresh commands because the setting is done whenever the tree items are first constructed / refreshed.  This should theoretically always keep it in sync with the latest view.


MicroFish91:
Closing in favor of an alternative custom solution

PR #789, opened 2024-11-14
Add separate suite of tests for managed identity and admin credential deployments

MicroFish91:
Add new managed identity tests for connecting to ACRs.  Convert old tests into the `admincred` suite of tests which used to be default.

I duplicated the monorepo test project so that `monorepo-admincreds` and `monorepo-identity` can both run their own deployments in parallel without issue since deployments are expected to temporarily modify the test project with saved settings.

Local (Note the 3 tests):
image

Remote (Note only 2 tests):
image

FYI, I found that the managed identity tests were not working remotely due to an auth privilege issue not allowing us to do role assignments in the manual test sub.  Thus the managed identity tests should only be run locally for now in a sub where the user has full auth privileges to assign the proper roles until we come up with a long term solution.


MicroFish91:
With the latest update from main, these tests will break again, will fix in next PR.

nturinski (approved):

PR #796, opened 2024-12-03
Add `editEnvironmentVariableName` command

MicroFish91:


package.json
```diff
@@ -484,6 +490,11 @@
                     "when": "view =~ /(azureResourceGroups|azureFocusView)/ && viewItem =~ /environmentVariablesItem/i",
                     "group": "1@1"
                 },
+                {
+                    "command": "containerApps.editEnvironmentVariableName",
+                    "when": "view =~ /(azureResourceGroups|azureFocusView)/ && viewItem =~ /environmentVariableItem/i",
+                    "group": "1@1"
```
nturinski, line 496: I think technically this should be 1@2

src/commands/environmentVariables/editEnvironmentVariable/EnvironmentVariableEditContext.ts
```diff
@@ -0,0 +1,22 @@
...
+*--------------------------------------------------------------------------------------------*/
+
+import { type EnvironmentVar } from "@azure/arm-appcontainers";
+import { type SetTelemetryProps } from "../../../telemetry/SetTelemetryProps";
+import { type ContainerUpdateTelemetryProps as TelemetryProps } from "../../../telemetry/commandTelemetryProps";
+import { type ISecretContext } from "../../secret/ISecretContext";
+import { type EnvironmentVariablesBaseContext } from "../EnvironmentVariablesContext";
+import { type EnvironmentVariableType } from "../addEnvironmentVariable/EnvironmentVariableTypeListStep";
+
+export interface EnvironmentVariableEditBaseContext extends EnvironmentVariablesBaseContext, Pick<ISecretContext, 'secretName'> {
```
nturinski, line 13: Seems to be very similar to the `EnvironmentVariableAddBaseContext`. Any reason to make a new one?

src/commands/environmentVariables/editEnvironmentVariable/EnvironmentVariableEditDraftStep.ts
```diff
@@ -0,0 +1,40 @@
...
+import { type Container, type EnvironmentVar } from "@azure/arm-appcontainers";
+import { nonNullValue } from "@microsoft/vscode-azext-utils";
+import { type Progress } from "vscode";
+import { type ContainerAppItem } from "../../../tree/ContainerAppItem";
+import { type RevisionsItemModel } from "../../../tree/revisionManagement/RevisionItem";
+import { localize } from "../../../utils/localize";
+import { RevisionDraftUpdateBaseStep } from "../../revisionDraft/RevisionDraftUpdateBaseStep";
+import { type EnvironmentVariableEditContext } from "./EnvironmentVariableEditContext";
+
+export class EnvironmentVariableEditDraftStep<T extends EnvironmentVariableEditContext> extends RevisionDraftUpdateBaseStep<T> {
```
nturinski, line 15: It's not a huge deal, but I feel like this step could almost entirely be consolidated with the "AddStep".

But if you like having them separate just for organization/logical purposes, makes sense to me too.

nturinski (approved):

src/commands/environmentVariables/editEnvironmentVariable/EnvironmentVariableEditDraftStep.ts
MicroFish91: Yeah, although there are a lot of overlaps, there's enough differences that I'd prefer to leave them uncoupled to reduce complexity if we ever need to come back and alter the behavior.

src/commands/environmentVariables/editEnvironmentVariable/EnvironmentVariableEditContext.ts
MicroFish91: Yeah you're right, they are super similar.  To simplify a bit, I extended from the add context and just made the `environmentVariable` required because it's needed up front for the edits.  I think I still prefer to keep a separate context just because I'm trying to keep the two command paths decoupled for the most part.

package.json
MicroFish91: I think 1@1 is correct, but it's confusing because the one above it is `environmentVariablesItem` with an s and they look super similar 😆

PR #798, opened 2024-12-03
Add `deleteEnvironmentVariable` command

MicroFish91:


src/commands/environmentVariables/deleteEnvironmentVariable/EnvironmentVariableDeleteContext.ts
```diff
@@ -0,0 +1,15 @@
...
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+import { type EnvironmentVar } from "@azure/arm-appcontainers";
+import { type SetTelemetryProps } from "../../../telemetry/SetTelemetryProps";
+import { type ContainerUpdateTelemetryProps as TelemetryProps } from "../../../telemetry/commandTelemetryProps";
+import { type EnvironmentVariablesBaseContext } from "../EnvironmentVariablesContext";
+
+export interface EnvironmentVariableDeleteBaseContext extends EnvironmentVariablesBaseContext {
+    environmentVariable: EnvironmentVar;
```
nturinski, line 12: Should this just be part of the edit context?

src/commands/environmentVariables/deleteEnvironmentVariable/deleteEnvironmentVariable.ts
```diff
@@ -0,0 +1,56 @@
...
+        ...await createActivityContext(),
+        subscription,
+        managedEnvironment: await getManagedEnvironmentFromContainerApp({ ...context, ...subscriptionContext }, containerApp),
+        containerApp,
+        containersIdx: item.containersIdx,
+        environmentVariable: item.envVariable,
+    };
+    wizardContext.telemetry.properties.revisionMode = containerApp.revisionsMode;
+
+    const confirmMessage: string = localize('confirmDeleteEnv', 'Are you sure you want to delete environment variable "{0}"?', item.envVariable.name);
```
nturinski, line 42: Should we put a message saying that this isn't destructible (until you upload the draft?)

MicroFish91: Yeah that's a good idea.  I added `(draft)` to the message to help clarify this, do you think that would be clear enough?

src/commands/environmentVariables/deleteEnvironmentVariable/EnvironmentVariableDeleteContext.ts
MicroFish91: Just to clarify, which edit context are you referring to?

nturinski: There is an interface called `EnvironmentVariableEditBaseContext ` or something like that. But it is shaped very similarly to this interface.

src/commands/environmentVariables/deleteEnvironmentVariable/deleteEnvironmentVariable.ts
nturinski: I don't think saying (draft) in the message would really clarify it for users unless they were already familiar with the (draft) vernacular in our extension. I'd probably say something to the effect of: "Changes are not finalized until uploaded" or "Are you sure you want to delete environment variable "{0}" locally? This will not be reflected in your container app until you deploy changes."

Something along those lines.

src/commands/environmentVariables/deleteEnvironmentVariable/EnvironmentVariableDeleteContext.ts
MicroFish91: Oh I see what you mean, yeah they are super similar, I think the notable difference is the base interface being extended from.  I didn't want to include all the editable properties on the context for the delete command

PR #805, opened 2024-12-18
Add retries for container app updates that fail after enabling admin credential-based access

### LINKED ISSUE #791: [Unstable] An error occurs when updating an ACR image to a container app
**OS:** All
**Build Version:** 20241114.1

**Repro Steps:**

1. Create and push an image to ACR.
2. Update the container app image using the newly pushed ACR image.
3. Select "Docker Login Credentials" registry method during the update process.
4. Check whether succeeds to update and deploy an image to the container app.

**Expect:**
Succeed to deploy the image to the container app.
Image

**Actual:**
Fails to deploy the image to the container app.
Image
Image

**More Info:** 
1. This issue only reproduces when it is a new image in registry and update to the container app at the first time.
2. This issue does not reproduce when updating and deploying the image for the second time.
3. This issue does not reproduce when selecting "managed identity" registry method.
4. Tried the reproduce steps 8 times and this issue can be reproduced 6 times.

MicroFish91:
Fix #791 

From my experience, this error is specifically due to needing more time for the admin credentials to propagate.  This does not reproduce for managed identities.

The reason this often doesn't reproduce for workspace project deployments is because there's a natural delay taken since ACR needs to build the image first, and so often the permissions have finished propagating by the time the image is ready.

This natural delay doesn't happen in cases where the image already exists, so we should add retries to help stabilize this issue.


src/commands/image/imageSource/ContainerAppUpdateStep.ts
```diff
@@ -51,13 +53,35 @@ export class ContainerAppUpdateStep<T extends ImageSourceContext & IngressContex
...
+                ext.outputChannel.appendLog(message);
+
+                await ext.state.runWithTemporaryDescription(containerApp.id, localize('updating', 'Updating...'), async () => {
+                    context.containerApp = await updateContainerApp(context, context.subscription, containerAppEnvelope);
+                    ext.state.notifyChildrenChanged(containerApp.managedEnvironmentId);
+                });
+            },
+            {
+                onFailedAttempt: (err: retry.FailedAttemptError) => {
+                    if (!/authentication\srequired/i.test(err.message)) {
```
MicroFish91, line 77: Not sure if I should only run retries after seeing "authentication required" in the error message, or if I should also check that admin credential-based auth was selected.

src/commands/image/imageSource/ContainerAppUpdateStep.ts
```diff
@@ -51,13 +53,36 @@ export class ContainerAppUpdateStep<T extends ImageSourceContext & IngressContex
             name: getContainerNameForImage(nonNullProp(context, 'image')),
         });
 
-        const updating = localize('updatingContainerApp', 'Updating container app...', containerApp.name);
-        progress.report({ message: updating });
+        // Related: https://github.com/microsoft/vscode-azurecontainerapps/pull/805
+        const retries = 4;
+        await retry(
+            async (currentAttempt: number): Promise<void> => {
+                if (currentAttempt === 2) {
```
nturinski, line 60: Worth checking the error code instead of just automatically retrying and outputting this message? If it really is a propagation issue, I kind of lean towards not telling the user and just retry in the background a few times.

MicroFish91: Will reach out to discuss offline

src/commands/image/imageSource/ContainerAppUpdateStep.ts
```diff
@@ -51,13 +55,25 @@ export class ContainerAppUpdateStep<T extends ImageSourceContext & IngressContex
...
         });
 
-        const updating = localize('updatingContainerApp', 'Updating container app...', containerApp.name);
-        progress.report({ message: updating });
-
-        await ext.state.runWithTemporaryDescription(containerApp.id, localize('updating', 'Updating...'), async () => {
-            context.containerApp = await updateContainerApp(context, context.subscription, containerAppEnvelope);
-            ext.state.notifyChildrenChanged(containerApp.managedEnvironmentId);
-        });
+        // Related: https://github.com/microsoft/vscode-azurecontainerapps/pull/805
```
nturinski, line 58: Did you mean to link the issue: https://github.com/microsoft/vscode-azurecontainerapps/issues/791?

I feel like that makes more sense but it's all linked so either is fine.

nturinski (approved):
LGTM 👍

src/commands/image/imageSource/ContainerAppUpdateStep.ts
MicroFish91: Oh yeah I meant to link that, I'll fix real quick before merging

MicroFish91: Oh yeah I figured this PR description would get readers more quickly up to speed on why the retries were implemented

PR #806, opened 2024-12-19
Improve output and activity logs for `deployContainerApp`

MicroFish91:
Purpose: Add a simplified, more generic, one-click deploy command on the container app node.  This deploy command does not rely on the revision draft system.

New command `Deploy to Container App`:
image

Replaces `Deploy Workspace to Container App`:
image

Now makes this the first prompt:
image

Choosing `Container Registry` runs a one-click deploy version of `Edit Container Image...` (draft command).
Choosing `Workspace Project` runs `deployWorkspaceProject` (already a one-click deploy command so we can just call directly).

Example of updated activity log children (there are multiple possible child combinations, just showing one possible version of selecting `Container Registry`):

image


src/tree/ContainerAppItem.ts
```diff
@@ -212,6 +212,8 @@ export async function getContainerEnvelopeWithSecrets(context: IActionContext, s
     const concreteContainerAppEnvelope = <Required<ContainerApp>>containerAppEnvelope;
     const webClient: ContainerAppsAPIClient = await createContainerAppsAPIClient([context, createSubscriptionContext(subscription)]);
 
+    // Todo: If a container app fails to provision, this list command may error with code 400
```
MicroFish91, line 215: Randomly ran into this issue while testing things, making note of it and will turn this into an issue for later.

src/commands/image/deployImageApi/deployImage.ts
```diff
@@ -33,19 +31,12 @@ export async function deployImage(context: IActionContext & Partial<ContainerReg
...
-    // If more than the image tag changed, prompt for ingress again
-    if (getImageNameWithoutTag(wizardContext.containerApp?.template?.containers?.[0].image ?? '') !== getImageNameWithoutTag(wizardContext.image ?? '')) {
-        promptSteps.push(new IngressPromptStep());
-    }
-    promptSteps.push(new ContainerAppOverwriteConfirmStep());
-
     const wizard: AzureWizard<DeployImageApiContext> = new AzureWizard(wizardContext, {
         title: localize('deploy', 'Deploy image to container app "{0}"', containerApp.name),
-        promptSteps,
+        promptSteps: [
```
MicroFish91, line 36: Can simplify the steps added here because `ContainerRegistryImageConfigureStep` would now determine whether an `IngressPromptStep` needs to be added

nturinski (approved):

PR #809, opened 2024-12-19
Improvements to ACR selection prompting

### LINKED ISSUE #440: Improve the ACR selection dropdown
Currently, the list displays all ACRs in a subscription in no specific order. It can be difficult to find the ACR to be used.

image

Usually, the ACA environment to be used is already known at this point, and it's likely the user wants to use a registry in the same subscription. Is it possible to:

* Show the resource groups in the list
* Identify in the list which ones are in the same RG as the ACA env, and ideally sort them to the top

MicroFish91:
Closes #440 

Based on feedback from Anthony, let's no longer default the registries if a registry has not already been saved in the workspace settings.

Also, now sort registries by resource group.  Resource groups are sorted alphabetically and registries in the same resource group are sorted to the top as `recommended`.

image

image

If a currently deployed registry exists, don't mark the same resource group registries as recommended, but mark the currently deployed registry instead:

image

Also mention the registry name when prompting for the registry connection method:

image


MicroFish91:
Update: I updated to use group dividers when registries exist within the same resource group, it currently looks like this but we can change based on feedback

image

nturinski (approved):

Issue #719, opened 2024-08-02
Managed Identities: Container Registry Credentials


MicroFish91:
#### Required
- [x] 1. Add a managed identity option for new registry credentials.  This flow will impact the following commands which may need to be verified/updated:
    - [x] Create container app
    - [x] Update image
    - [x] Deploy workspace project
    - [x] Deployments through the docker extension
    - [x] Deployments through the functions extension (backwards compatibility to preserve existing behavior by default, can expand to support managed identity once we have more clearly defined plan)

#### Nice to haves
- [x] 2. More robust activity log support for the primary commands (show activity children).  This will help to make our changes more transparent to the user (e.g. highlighting when we set system assigned identity, rbac, etc.):
    - [x] Create container app
    - [x] Update image
    - [x] Deploy workspace project (This is probably mostly done already)
    - [x] Deployments through the docker extension
    
- [ ] 3. ~~Registry configuration and commands in the tree view~~
    - [ ] ~~Add a tree view representation of registry credentials with warning symbols shown next to non-managed identity credentials~~
    - ~~Potential related commands that could be added:~~
        - [ ] ~~Convert existing credential to managed identity~~
        - [ ] ~~Update existing docker login credentials (e.g. one way to quickly rotate a cred)~~
        - [ ] ~~Add new credential~~

MicroFish91:
Closing this issue as finished (Part 1 & 2).  Part 3 will be re-opened as a separate issue for the next milestone.

PR #814, opened 2024-12-24
Don't show ingress prompts in draft mode and add additional activity children to core draft commands

MicroFish91:
Partially addresses part 2 of https://github.com/microsoft/vscode-azurecontainerapps/issues/719

Examples:

Edit container (edit w/ container registry image):
image

Edit container image (edit w/ workspace project image):
image


src/commands/editContainer/editContainer.ts
```diff
@@ -48,9 +52,42 @@ export async function editContainer(context: IActionContext, node?: ContainersIt
         managedEnvironment: await getManagedEnvironmentFromContainerApp({ ...context, ...subscriptionContext }, containerApp),
         containerApp,
         containersIdx,
+        isDraftCommand: true,
     };
     wizardContext.telemetry.properties.revisionMode = containerApp.revisionsMode;
 
+    if (isAzdExtensionInstalled()) {
```
nturinski, line 59: Why is this being added here? Doesn't seem to be related to the PR?

src/commands/editContainer/editContainer.ts
```diff
@@ -48,9 +52,42 @@ export async function editContainer(context: IActionContext, node?: ContainersIt
...
+    }
+
+    const resourceGroups: ResourceGroup[] = await ResourceGroupListStep.getResourceGroups(wizardContext);
+    wizardContext.resourceGroup = nonNullValue(
+        resourceGroups.find(rg => rg.name === item.containerApp.resourceGroup),
+        localize('containerAppResourceGroup', 'Expected to find the container app\'s resource group.'),
+    );
+
+    // Log resource group
+    wizardContext.activityChildren?.push(
```
nturinski, line 70: I feel like it might be time to make a utility function to just generate all of this boilerplate for you. 

If not a util function, maybe include adding these activity children when you call `pickContainerApp` or `pickManagedEnvironment`? Since that seems to be where you are setting all of the wizard's contexts.

I feel like just adding them into the prompt steps themselves would be the best way to do it, but since these are top-level Azure resources, that might not be possible?

nturinski (approved):
Logically it makes sense to me. My comments are suggestions, but I'm not going to try to require it.

src/commands/editContainer/editContainer.ts
MicroFish91: Yeah, it was small enough that I snuck it in, like you said below this part is mostly boilerplate

src/commands/editContainer/editContainer.ts
MicroFish91: Oh interesting, I hadn't thought about doing it that way, but that might be a clean way to handle this boilerplate.  I'll investigate that as a potential solution.

MicroFish91: I consolidated most of the boilerplate under a new starting resource log step, let me know if you like that implementation or think it should be encapsulated differently!

nturinski (approved):
Definitely less concerned about StartingResourcesLogStep since it's always set to false for shouldPrompt 👍

Issue #774, opened 2024-10-16
Block `deployWorkspaceProject` in multiple revisions mode


MicroFish91:
Needs discussion, but should we force users to use the draft mode to deploy in multiple revisions mode?

PR #815, opened 2024-12-24
Add extra guard clauses to deployment commands

### LINKED ISSUE #773: Block or add an extra warning for users who try to run `deployWorkspaceProject` with multiple containers already active

MicroFish91:
Close #773, #774

I think these guard clauses will improve reliability for the one click deploy commands.

Let me know if you disagree with having any of these guard clauses in place.

Update:
image
image


src/commands/deployContainerApp/deployContainerApp.ts
```diff
@@ -25,6 +26,13 @@ export async function deployContainerApp(context: IActionContext, node?: Contain
     const subscriptionContext: ISubscriptionContext = createSubscriptionContext(item.subscription);
     const subscriptionActionContext: ISubscriptionActionContext = { ...context, ...subscriptionContext };
 
+    if (item.containerApp.revisionsMode === KnownActiveRevisionsMode.Multiple) {
+        throw new Error(localize('multipleRevisionsNotSupported', 'The container app "{0}" cannot be updated using "Deploy to Container App..." while in multiple revisions mode.', item.containerApp.name));
```
nturinski, line 30: I think it'd be helpful to say how they can deploy it if not "Deploy to Container App." Same for the error below. I assume that you have to just right-click the specific revision or container depending on the scenario?

MicroFish91: Yeah good idea.  I eventually would like to enable use of this command on the revisions directly, but I think that's a change for a future release

src/commands/deployContainerApp/deployContainerApp.ts
```diff
@@ -16,15 +17,25 @@ import { getVerifyProvidersStep } from "../../utils/getVerifyProvidersStep";
...
 
+const deployContainerAppCommandName: string = localize('deployContainerApp', 'Deploy to Container App...');
+
 export async function deployContainerApp(context: IActionContext, node?: ContainerAppItem): Promise<void> {
     const item: ContainerAppItem = node ?? await pickContainerApp(context);
     const subscriptionContext: ISubscriptionContext = createSubscriptionContext(item.subscription);
     const subscriptionActionContext: ISubscriptionActionContext = { ...context, ...subscriptionContext };
 
+    if (item.containerApp.revisionsMode === KnownActiveRevisionsMode.Multiple) {
+        throw new Error(localize('multipleRevisionsNotSupported', 'The container app cannot be updated using "{0}" while in multiple revisions mode. Navigate to the revision\'s container and execute "{1}" instead.', deployContainerAppCommandName, editContainerCommandName));
```
nturinski, line 33: nit: Add these error strings to the `const` file.

nturinski (approved):

PR #816, opened 2024-12-27
Add backward compatibility to API in support of the new managed identity features

MicroFish91:
Partially addresses #719 

Updates the api consumed by functions so that it keeps working in face of all the new features we've added related to managed identities


src/commands/api/deployWorkspaceProjectApi.ts
```diff
@@ -38,12 +48,14 @@ export async function deployWorkspaceProjectApi(deployWorkspaceProjectOptions: a
             rootFolder,
             srcPath: srcPath ? Uri.file(srcPath).fsPath : undefined,
             dockerfilePath: dockerfilePath ? Uri.file(dockerfilePath).fsPath : undefined,
+            newRegistryCredentialType: RegistryCredentialType.DockerLogin,
```
MicroFish91, line 51: We need discussions about whether we want to default to managed identity here in the future.  For now, I'm keeping it was `Docker Login` to preserve existing behavior.

src/commands/api/getAzureContainerAppsApiProvider.ts
```diff
@@ -9,7 +9,7 @@ import type * as api from "./vscode-azurecontainerapps.api";
 
 export function getAzureContainerAppsApiProvider(): apiUtils.AzureExtensionApiProvider {
     return createApiProvider([<api.AzureContainerAppsExtensionApi>{
-        // Todo: Change this to 0.0.2 later.  0.0.2 is backwards compatible anyway so this change should be fine either way.
+        // Todo: Change this to 0.0.3 later.  0.0.3 is backwards compatible anyway so this change should be fine either way.
```
MicroFish91, line 12: When we eventually add managed identity support for the api, I will bump the major version

src/commands/api/deployWorkspaceProjectApi.ts
```diff
@@ -38,12 +48,14 @@ export async function deployWorkspaceProjectApi(deployWorkspaceProjectOptions: a
...
             shouldSaveDeploySettings: !!shouldSaveDeploySettings,
         });
 
         const deployWorkspaceProjectContext: DeployWorkspaceProjectContext = await deployWorkspaceProjectInternal(deployWorkspaceProjectInternalContext, {
             suppressActivity: true,
             suppressConfirmation,
+            suppressRegistryPrompt: suppressRegistryPrompt ?? true,
```
MicroFish91, line 58: This [PR](https://github.com/microsoft/vscode-azurecontainerapps/pull/809) introduced the prompt, so add ability to suppress and default it to true to preserve existing behavior

src/commands/api/getAzureContainerAppsApiProvider.ts
nturinski: Okay, but just so you know, this changing is purely aesthetic. semver doesn't actually register this as a different version number since there's no major version number.

src/commands/registryCredentials/dockerLogin/DockerLoginRegistryCredentialsContext.ts
```diff
@@ -17,4 +17,6 @@ export interface DockerLoginRegistryCredentialsContext extends CreateAcrContext,
 
     newRegistrySecret?: Secret;
     newRegistryCredential?: RegistryCredentials;
+
+    suppressEnableAdminUserPrompt?: boolean;
```
nturinski, line 21: Can't this just check `suppressConfirmation`?

nturinski (approved):

src/commands/registryCredentials/dockerLogin/DockerLoginRegistryCredentialsContext.ts
MicroFish91: I don't think that prop exists on the context currently.  There is the `suppressNotification` option in user settings for resources, but I don't think we want to use that because this is ordinarily a prompt that I would always want to show.  `suppressConfirmation` in this case is an explicit option that is being passed into the `deployWorkspaceProjectInternal` command because an outside extension specifically wants to bypass.

PR #835, opened 2025-01-10
Update telemetry to reflect new release prep changes

MicroFish91:


src/commands/deployContainerApp/deployContainerApp.ts
```diff
@@ -49,7 +50,6 @@ export async function deployContainerApp(context: IActionContext, node?: Contain
         managedEnvironment: await getManagedEnvironmentFromContainerApp(subscriptionActionContext, item.containerApp),
         imageSource,
     };
-    wizardContext.telemetry.properties.revisionMode = item.containerApp.revisionsMode;
```
MicroFish91, line 52: Because of the current guard clauses, we know this must be single revision

nturinski: I don't mind changing it, but I feel like it's better to just keep it explicitly there. I could see there being confusion in its absence if someone were to track it.

nturinski (approved):

PR #846, opened 2025-01-17
Make env updates additive when updating directly

### LINKED ISSUE #842: The existing environment variables disappear after bulk editing environment variables
**OS:** All
**Build:** v0.7.2-alpha.0

**Repro Steps:**

1. Run `git clone https://github.com/MicroFish91/aca-hello-world-v2` and then open this project in VS Code.
2. Make sure the container app contains several environment variables.
3. Expand above container app -> Expand the "Container" node -> Right-click "Environment Variables" -> Click "Edit Environment Variables (.env)..." -> Select a .env file -> Save and deploy the changes.
4. Check whether the existing environment variables still exist.

**Expect:** 
The existing environment variables still exist.

**Actual:** 
The existing environment variables disappear.

**More Info:** 
This issue also reproduces when deploying a workspace project that contains a .env file.

MicroFish91:
Closes #842


src/commands/environmentVariables/editEnvironmentVariables/EnvironmentVariablesEditDraftStep.ts
```diff
@@ -21,10 +22,17 @@ export class EnvironmentVariablesEditDraftStep<T extends EnvironmentVariablesEdi
     public async execute(context: T, progress: Progress<{ message?: string | undefined; increment?: number | undefined }>): Promise<void> {
         progress.report({ message: localize('editingEnv', 'Editing environment variables (draft)...') });
         this.revisionDraftTemplate.containers ??= [];
-
         const container: Container = this.revisionDraftTemplate.containers[context.containersIdx] ?? {};
-        container.env = context.environmentVariables;
 
+        const envMap: Map<string, EnvironmentVar> = new Map();
+        for (const env of container.env ?? []) {
```
nturinski, line 28: Dunno if it impacts perf or not, but you could also just make a new Map from the JSON object, and then overwrite with the for of statement below.

I think it's more readable since it's a little confusing at first seeing both of these loops, but that may just be me.

PR #896, opened 2025-06-24
Add functionality to copilot button

motm32:
This opens the chat and adds a prompt when the copilot button is pressed within the confirmation view. Here is an example of what the prompt looks like and the copilot response: 
image

### DIFF
--- a/src/webviews/ConfirmationViewController.ts
+++ b/src/webviews/ConfirmationViewController.ts
@@ -11,6 +11,7 @@ import { WebviewController } from "./extension-server/WebviewController";
 export type ConfirmationViewControllerType = {
     title: string;
     description: string;
+    commandName: string;
     items: Array<ConfirmationViewProperty>
 }
 
--- a/src/webviews/ConfirmationViewController.ts
+++ b/src/webviews/ConfirmationViewController.ts
@@ -11,7 +11,7 @@ import { WebviewController } from "./extension-server/WebviewController";
 export type ConfirmationViewControllerType = {
     title: string;
     description: string;
-    commandName: string;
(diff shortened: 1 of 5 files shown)


src/webviews/extension-server/WebviewController.ts
```diff
@@ -62,6 +63,14 @@ export class WebviewController<Configuration> extends WebviewBaseController<Conf
                     case 'confirm':
                         SharedState.itemsToClear = message.itemsToClear ?? 0;
                         this._panel.dispose();
+                        break;
+                    case 'copilot':
+                        await vscode.commands.executeCommand("workbench.action.chat.newChat");
+                        await vscode.commands.executeCommand("workbench.action.chat.open", {
```
MicroFish91, line 69: I was running into an issue where if the chat wasn't already open, `newChat` wouldn't actually start a new chat.  Do you run into this issue at all?  If so, I was calling open before `newChat` to fix this behavior, see:

https://github.com/microsoft/vscode-azureresourcegroups/blob/main/src/chat/askAgentAboutActivityLog.ts

src/webviews/ConfirmationViewController.ts
```diff
@@ -11,6 +11,7 @@ import { WebviewController } from "./extension-server/WebviewController";
 export type ConfirmationViewControllerType = {
     title: string;
     description: string;
+    commandName: string;
```
alexweininger, line 14: We should document that the commandName is currently only used to help construct the copilot prompt

alexweininger (approved):

alexweininger (approved):

PR #901, opened 2025-07-08
Add telemetry for confirmation view

motm32:
Still a work in progress need to do some more testing for each scenario. 

Here is how the information message looks:
image

The question being asked changes based off the interactions the user makes with the view.


src/webviews/OpenConfirmationViewStep.ts
```diff
@@ -43,6 +43,7 @@ export class OpenConfirmationViewStep<T extends IActionContext> extends AzureWiz
             confirmationView.onDisposed(() => {
                 try {
                     if (SharedState.itemsToClear > 0) {
```
nturinski, line 45: Discussed offline:

I think we should add telemetry based on if they clicked any of the copilot buttons as well. A non-intrusive information window could pop up the first time that they use this feature that asks if they felt like it was helpful with a 👍 and 👎 as the response as well.

src/webviews/OpenConfirmationViewStep.ts
```diff
@@ -63,3 +107,16 @@ export class OpenConfirmationViewStep<T extends IActionContext> extends AzureWiz
...
 }
+
+export async function confirmationViewButtonActions(context: IActionContext, result: vscode.MessageItem | undefined): Promise<void> {
+    if (result) {
+        if (result.title === 'Great') {
+            context.telemetry.properties.confirmationViewHelpful = 'true';
+        } else if (result.title === 'Unhelpful') {
+            context.telemetry.properties.confirmationViewNotHelpful = 'true';
+        } else if (result.title === 'Provide Feedback') {
+            const confirmationViewSurveyLink = 'https://www.surveymonkey.com/r/DD5LC5Y'
```
alexweininger, line 123: should this be an aka.ms link?

alexweininger (approved):

src/telemetry/ConfirmationViewTelemetryProps.ts
```diff
@@ -0,0 +1,14 @@
+/*---------------------------------------------------------------------------------------------
+*  Copyright (c) Microsoft Corporation. All rights reserved.
+*  Licensed under the MIT License. See License.md in the project root for license information.
+*--------------------------------------------------------------------------------------------*/
+
+export interface ConfirmationViewTelemetryProps {
```
nturinski, line 6: Why are these all only true? Shouldn't they be 'true' | 'false' ?

motm32: I was following what Matt Fisher did for the azd telemetry props: https://github.com/microsoft/vscode-azurecontainerapps/blob/main/src/telemetry/AzdTelemetryProps.ts. 

My thoughts on this is we only really care if this event is true. I thought this would basically give us a count of how many users are doing these events but let me know if that thought process is incorrect.

nturinski: I guess we should add `isCopilotEvent` is as well?

MicroFish91: In this case these aren't optional, so I would either make them optional or allow 'false' **

nturinski (approved):

nturinski (approved):

Issue #906, opened 2025-07-16
Deployment silently fails but shows up as a success if image builds but doesn't run


MicroFish91:
From Alex:

Clone this -> https://github.com/Azure-Samples/containerapps-albumapi-csharp
 
Change the ENTRYPOINT in the dockerfile to something that doesn't exist
 
Creates a container runtime error that makes ACA not switch to the new revision

PR #909, opened 2025-07-17
Add new step to automatically verify if the deployed container app started successfully

MicroFish91:
Closes https://github.com/microsoft/vscode-azurecontainerapps/issues/906

image

image

### DIFF
--- a/src/commands/deployContainerApp/deployContainerApp.ts
+++ b/src/commands/deployContainerApp/deployContainerApp.ts
@@ -14,6 +14,7 @@ import { getVerifyProvidersStep } from "../../utils/getVerifyProvidersStep";
 import { localize } from "../../utils/localize";
 import { pickContainerApp } from "../../utils/pickItem/pickContainerApp";
 import { OpenConfirmationViewStep } from "../../webviews/OpenConfirmationViewStep";
+import { CommandAttributes } from "../CommandAttributes";
 import { ContainerAppOverwriteConfirmStep } from "../ContainerAppOverwriteConfirmStep";
 import { deployWorkspaceProject } from "../deployWorkspaceProject/deployWorkspaceProject";
 import { editContainerCommandName } from "../editContainer/editContainer";
@@ -50,17 +51,7 @@ export async function deployContainerApp(context: IActionContext, node?: Contain
         containerApp: item.containerApp,
         managedEnvironment: await getManagedEnvironmentFromContainerApp(subscriptionActionContext, item.containerApp),
         imageSource,
-        activityAttributes: {
-            description: `Deploys an existing image from a container registry to a target Azure Container App.
-            The container registry and image must already be available for pulling.
-            Supports public images from any registry, and both public and private images from Azure Container Registry (ACR).
-            For private image deployment from other third party registries, we support deployment through the 'vscode-containers' extension
-            via the command titled "Container Registries: Deploy Image to Azure Container Apps...".`,
-            troubleshooting: [
-                `If a container app resource envelope is provided in attributes, do not confuse null secrets as missing container app secrets. This is because secrets are not typically
-                copied over with the core resource metadata. Any issues with secrets will require inspecting the remote resource directly.`,
-            ]
-        },
+        activityAttributes: CommandAttributes.DeployContainerAppContainerRegistry,
     };
 
     if (isAzdExtensionInstalled()) {
--- a/src/commands/image/imageSource/ContainerAppStartVerificationStep.ts
+++ b/src/commands/image/imageSource/ContainerAppStartVerificationStep.ts
@@ -122,7 +122,7 @@ export class ContainerAppStartVerificationStep<T extends ContainerAppStartVerifi
                 break;
             }
 
-            await delayWithExponentialBackoff(attempt, 2000 /** baseDelayMs */, maxWaitTimeMs);
+            await delayWithExponentialBackoff(attempt, 1000 /** baseDelayMs */, maxWaitTimeMs);
             attempt++;
 
             revision = await this._client.containerAppsRevisions.getRevision(parsedRevision.resourceGroup, containerAppName, parsedRevision.resourceName);
--- a/src/commands/CommandAttributes.ts
+++ b/src/commands/CommandAttributes.ts
@@ -1,0 +1,35 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { type ActivityAttributes } from "@microsoft/vscode-azext-utils";
+
+export class CommandAttributes {
+
+    static readonly DeployContainerAppContainerRegistry: ActivityAttributes = {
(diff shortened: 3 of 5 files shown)


src/commands/image/imageSource/ContainerAppStartVerificationStep.ts
```diff
@@ -0,0 +1,200 @@
...
+        let revisions: Revision[];
+
+        const start: number = Date.now();
+
+        while (true) {
+            if ((Date.now() - start) > maxWaitTimeMs) {
+                break;
+            }
+
+            await delay(1000);
```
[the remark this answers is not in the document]
nturinski: Also I think delay should accept a parameter for controlling the ms. Maybe this is too complicated, but I like doing exponential backoff so we're not pinging as often.

src/commands/image/imageSource/ContainerAppStartVerificationStep.ts
```diff
@@ -0,0 +1,200 @@
...
+
+    /**
+     * Try to query for any logs associated with the revision and add them to the Copilot activity attributes
+     */
+    private async tryAddLogAttributes(context: T, revisionName: string) {
+        // Basic validation check since we're including a name directly in the query
+        if (revisionName.length > 54 || !/^[\w-]+$/.test(revisionName)) {
+            const invalidName: string = localize('unexpectedRevisionName', 'Internal warning: Encountered an unexpected revision name format "{0}". Skipping log query for the revision status check.', revisionName);
+            ext.outputChannel.appendLog(invalidName);
+            throw new Error(invalidName);
```
nturinski, line 149: Should this actually throw an error? Keeping in mind that if this step fails, it'll make the whole deployment be considered a failed operation.

src/commands/image/imageSource/ContainerAppStartVerificationStep.ts
```diff
@@ -0,0 +1,200 @@
...
+        if (queryResult.status !== LogsQueryResultStatus.Success) {
+            return;
+        }
+
+        const lines: string[] = [];
+        const table: LogsTable = queryResult.tables[0];
+
+        if (!table.rows.length) {
+            // Note: Often times we will only be able to find logs when the image source was for `RemoteAcrBuild`
+            throw new Error(localize('noQueryLogs', 'No query logs were found for revision "{0}".', revisionName));
```
nturinski, line 179: Similar to comment above.

src/commands/image/imageSource/ContainerAppUpdateStep.ts
```diff
@@ -87,4 +88,8 @@ export class ContainerAppUpdateStep<T extends ImageSourceContext & IngressContex
     public shouldExecute(context: T): boolean {
         return !!context.containerApp;
     }
+
+    public addExecuteSteps(): AzureWizardExecuteStep<T>[] {
```
nturinski, line 92: Is there any reason to not just add this step upfront when we create the wizard?

src/utils/azureClients.ts
```diff
@@ -43,3 +46,35 @@ export async function createAuthorizationManagementClient(context: AzExtClientCo
         return createAzureClient(context, (await import('@azure/arm-authorization')).AuthorizationManagementClient);
     }
 }
+
+/**
+ * @throws Throws an error if logsQueryClient is created by a user in a sovereign cloud environment.
```
nturinski, line 51: Should we throw? I feel like we should just return undefined and let the operation be skipped since this shouldn't be blocking the users from whatever command we're trying to run (probably deploy?)

src/commands/deployContainerApp/deployContainerApp.ts
```diff
@@ -50,6 +50,17 @@ export async function deployContainerApp(context: IActionContext, node?: Contain
         containerApp: item.containerApp,
         managedEnvironment: await getManagedEnvironmentFromContainerApp(subscriptionActionContext, item.containerApp),
         imageSource,
+        activityAttributes: {
```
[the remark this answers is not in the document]
MicroFish91: Yeah that's a good idea, it is definitely a bit ugly here too

src/commands/image/imageSource/ContainerAppStartVerificationStep.ts
MicroFish91: The error is caught, logged into telemetry, and then swallowed, so it should be fine!

src/utils/azureClients.ts
MicroFish91: This is a tough one, I don't know.  I feel like we should throw, but I could go either way.

src/commands/image/imageSource/ContainerAppStartVerificationStep.ts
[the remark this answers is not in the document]
MicroFish91: Hmm, let me think about this, maybe there's something I can do to emulate exponentional backoff

src/utils/azureClients.ts
nturinski: I mean, if you're on a sovereign cloud, it would really suck for your deployment to always report a failure just because the log client isn't compatible.

It's not technically blocking the user, but I do think it'll be confusing. I think we can report it as a blue or gray node, saying skipped due to no availability in a sovereign cloud environment or something.

MicroFish91:
I added most of the feedback, I need to test all the scenarios again just to double check everything is still good, but probably ready to review again for the most part

alexweininger:
Yeah this is awesome

src/commands/image/imageSource/ContainerAppStartVerificationStep.ts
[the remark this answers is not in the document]
MicroFish91: > I think we should put the delay at the end. It makes the first attempt always delayed by a second if you do it before trying to list revisions.

I’m leaning towards keeping the delay at the front for a couple of reasons. First, we usually need to wait anyway, so delaying the first attempt isn’t a real drawback. Second, putting the delay at the front ensures that, towards the end, the logic always gets one final attempt—even if the total delay is slightly exceeded.

package-lock.json
```diff
@@ -17,12 +17,13 @@
...
                 "@azure/container-registry": "1.0.0-beta.5",
                 "@azure/core-rest-pipeline": "1.10.3",
+                "@azure/monitor-query": "^1.2.0",
                 "@azure/storage-blob": "^12.4.1",
                 "@fluentui/react-components": "^9.56.2",
                 "@fluentui/react-icons": "^2.0.265",
                 "@microsoft/vscode-azext-azureutils": "^3.3.3",
                 "@microsoft/vscode-azext-github": "^1.0.2",
-                "@microsoft/vscode-azext-utils": "^3.2.1",
+                "@microsoft/vscode-azext-utils": "^3.3.0",
```
nturinski, line 26: Whoa.. any idea why there are 7273 lines of code deleted? Kind of wild.

nturinski (approved):
Looks good! Really excited about this feature tbh

package-lock.json
MicroFish91: It's from Azure Monitor Query, it blew my mind too, I don't really understand how it led to so many lines removed lol

alexweininger: Did it change the lockfile version?

PR #961, opened 2025-08-25
Add `Basic` and `Advanced` flow for `Deploy Workspace Project`

### LINKED ISSUE #853: Advanced scenarios for existing container apps

MicroFish91:
Update `deployWorkspaceProject` to better handle onboarding existing container apps.  The original implementation worked to reduce prompts by making a host of assumptions, but only allowed creation of brand new resources (more or less what we do for `basic` creates).  This version of the command introduces two different deployment modes which is more closely aligned with how our other extensions typically handle creation/deployment.

The deployment modes are -
- **Basic**: Assumes all resources are newly created; defaults as much as possible.
- **Advanced**: Offers more customization and way more prompts for key decisions.  Allows onboarding of existing container apps.

image

I also refactored list steps to be more generally reusable by allowing pick sorting / filtering to be controlled via injected strategies.

Closes #853

### DIFF
--- a/src/commands/deployWorkspaceProject/deploymentConfiguration/workspace/DeploymentModeListStep.ts
+++ b/src/commands/deployWorkspaceProject/deploymentConfiguration/workspace/DeploymentModeListStep.ts
@@ -17,7 +17,7 @@ export class DeploymentModeListStep extends AzureWizardPromptStep<WorkspaceDeplo
         const picks: IAzureQuickPickItem<DeploymentMode>[] = [
             {
                 label: localize('basic', 'Basic'),
-                detail: localize('basicDetails', 'Deploy a basic app with minimal customizations.'),
+                detail: localize('basicDetails', 'Deploy a basic app with minimal setup. Automatically creates all required resources.'),
                 data: DeploymentMode.Basic,
             },
             {


src/commands/deployWorkspaceProject/internal/deployWorkspaceProjectInternal.ts
```diff
@@ -100,145 +108,83 @@ export async function deployWorkspaceProjectInternal(
...
-
-        await LocationListStep.setLocation(wizardContext, wizardContext.resourceGroup.location);
-        ext.outputChannel.appendLog(localize('usingResourceGroup', 'Using resource group "{0}".', resourceGroupName));
     } else {
-        wizardContext.telemetry.properties.existingResourceGroup = 'false';
-        executeSteps.push(new ResourceGroupCreateStep());
-    }
-
-    // Managed environment
-    if (wizardContext.managedEnvironment) {
```
MicroFish91, line 135: All of the starting resource displaying is now handled by `StartingResourcesLogStep` which allows me to remove a ton of boilerplate code from this file.  Now it's really only only a matter of pushing in the right basic vs. advanced steps.

PR #965, opened 2025-08-27
Add `Deploy to Container App with Copilot` command 

motm32:
This add a new command using `CopilotUserInput` so copilot will provide all suggestions for the prompt steps.

Here is a video of how the new process looks (some of the wording in the video has been changed): 

https://github.com/user-attachments/assets/ad1f0041-1559-4112-ba1f-49373ce66e0a

Essentially what happens is once the command starts running a loading web view panel screen  will pop up. Then once the copilot responses have generated the summary view will pop up where users can either confirm or make edits. 

Todo:
- [ ] Update the utils package once released
- [ ] If copilot gives an invalid response the it is popped back to the user discard of the loading view


src/commands/deployContainerApp/deployContainerApp.ts
```diff
@@ -60,18 +61,28 @@ export async function deployContainerApp(context: IActionContext, node?: Contain
     wizardContext.telemetry.properties.revisionMode = item.containerApp.revisionsMode;
 
     const confirmationViewTitle: string = localize('summary', 'Summary');
-    const confirmationViewDescription: string = localize('viewDescription', 'Please select an input you would like to change. Note: Any input proceeding the changed input may need to change as well');
-    const confirmationViewTabTitle: string = localize('deployContainerAppTabTitle', 'Summary - Deploy Image to Container App');
-    const title: string = localize('deployContainerAppTitle', 'Deploy image to container app');
+    let confirmationViewDescription: string = localize('viewDescription', 'Please select an input you would like to change. Note: Any input proceeding the changed input may need to change as well');
```
nturinski: nit: Right now, it's not a may, it's guaranteed that it needs to change, right?

Also maybe we should note that a changed input may lead to different prompts?

nturinski (approved):
Looks great to me. 

Just thinking out loud, but I think in the future, all the wrapper code you put into deployContainerApp may be better to make a util function that basically just wraps the command with all of the Copilot trappings. It should make it easier to just apply to all of our commands.

alexweininger (approved):

MicroFish91:
Okay we released, so should be okay to start merging new changes!

PR #980, opened 2025-09-24
Cherrypick changes for MFA

nturinski:
There were a lot of changes to Deploy from workspace so I'm making a release branch from v0.9.0 and cherry-picking over the changes that we need for MFA.

### DIFF
--- a/test/runTest.ts
+++ b/test/runTest.ts
@@ -9,7 +9,7 @@ import * as path from 'path';
 
 async function main(): Promise<void> {
     try {
-        const vscodeExecutablePath = await downloadAndUnzipVSCode('insiders');
+        const vscodeExecutablePath = await downloadAndUnzipVSCode('stable');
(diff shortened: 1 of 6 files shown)


test/runTest.ts
```diff
@@ -9,14 +9,13 @@ import * as path from 'path';
 
 async function main(): Promise<void> {
     try {
-        const vscodeExecutablePath = await downloadAndUnzipVSCode('stable');
+        const vscodeExecutablePath = await downloadAndUnzipVSCode('insiders');
```
bwateratmsft, line 12: This is no longer needed.

bwateratmsft (changes requested):

test/runTest.ts
nturinski: Oh nice, I guess I'll go around and change all the extensions back

PR #999, opened 2025-12-10
Add changes for re running "Deploy to Container App" command 

motm32:
This PR adds a couple things:

- changes to the deploy to container app command to use an internal function. The command id when deploy to container app is run the first time around gets set directly to the internal command so when copilot re runs this command it goes straight into the internal command (see this [rg PR](https://github.com/microsoft/vscode-azureresourcegroups/pull/1316) for how that is done)
- adds ACA list step and managed environment list steps to the confirmation view

Some things to note
- If no subscription is passed in from the activity log items (which there almost always should be) this will revert to the regular deploy flow. I have found that copilot is not very good at picking a subscription. 
- If there is an error I added logic so the loading view closes.

Things to do: 
- [ ] update and release [utils ](https://github.com/microsoft/vscode-azuretools/pull/2136)changes


src/commands/deployContainerApp/deployContainerApp.ts
```diff
@@ -105,3 +137,7 @@ async function promptImageSource(context: ISubscriptionActionContext): Promise<I
 
     return nonNullProp(promptContext, 'imageSource');
 }
+
+function isCopilotUserInput(context: IActionContext): boolean {
+    return context.ui.constructor.name === 'CopilotUserInput';
```
nturinski: Can you do `context.ui instanceof CopilotUserInput`? I think it's a more robust detection method.

src/commands/deployContainerApp/deployContainerApp.ts
```diff
@@ -45,29 +48,58 @@ export async function deployContainerApp(context: IActionContext, node?: Contain
...
     const confirmationViewTitle: string = localize('summary', 'Summary');
     let confirmationViewDescription: string = localize('viewDescription', 'Please select an input you would like to change. Note: Any input proceeding the changed input will need to change as well');
     let confirmationViewTabTitle: string = localize('deployContainerAppTabTitle', 'Summary - Deploy Image to Container App');
     let title: string = localize('deployContainerAppTitle', 'Deploy image to container app');
 
-    const promptSteps: AzureWizardPromptStep<ContainerAppDeployContext>[] = []
-    if (wizardContext.ui instanceof CopilotUserInput) {
-        promptSteps.push(new OpenLoadingViewStep());
+    if (isCopilotUserInput(wizardContext)) {
         confirmationViewDescription = localize('viewDescription', 'Please review AI generated inputs and select any you would like to modify. Note: Any input proceeding the modified input will need to change as well');
```
nturinski:
```suggestion
        confirmationViewDescription = localize('viewDescription', 'Please review AI generated inputs and select any you would like to modify. Note: Any input preceding the modified input will need to change as well');
```

silentsnider-max:
@copilot fix all

src/commands/deployContainerApp/deployContainerApp.ts
```diff
@@ -105,3 +137,7 @@ async function promptImageSource(context: ISubscriptionActionContext): Promise<I
 
     return nonNullProp(promptContext, 'imageSource');
 }
+
+function isCopilotUserInput(context: IActionContext): boolean {
+    return context.ui instanceof CopilotUserInput;
```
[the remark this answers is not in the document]
motm32: I implemented this in this [PR ](https://github.com/microsoft/vscode-azuretools/pull/2136)so it could be shared with the resource groups extension.

src/commands/deployContainerApp/deployContainerApp.ts
```diff
@@ -39,62 +42,110 @@ export async function deployContainerApp(context: IActionContext, node?: Contain
...
+        let imageSource: ImageSource | undefined;
+        let registryDomain: SupportedRegistries | undefined;
+        if (isCopilotUserInput(context)) {
+            // If the input is coming from Copilot we want to default to Container Registry as we don't support 'deployWorkspaceProject' flow with Copilot at the moment
+            imageSource = ImageSource.ContainerRegistry;
+            registryDomain = acrDomain;
+            await openLoadingViewPanel(context);
+        } else {
+            // Prompt for image source before initializing the wizard in case we need to redirect the call to 'deployWorkspaceProject' instead
+            imageSource = await promptImageSource(context);
```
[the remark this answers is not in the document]
MicroFish91: Oh, I might be misunderstanding something—promptImageSource only runs when it’s not a Copilot input, right? If so, I'm hoping we can keep the prompt in the parent logic. That would also make things cleaner by keeping the internal logic more focused on the core wizard behavior we want for rerun.

As for centralizing Copilot logic, I was hoping we could do that within the internal logic since that's its natural entry-point. Ideally, I was hoping the centralization would live within the internal logic, although there may be constraints I’m not fully aware of so let me know your thoughts on what you think the best approach would be here

package-lock.json
```diff
@@ -11099,6 +11134,7 @@
             "version": "8.2.1",
             "resolved": "https://registry.npmjs.org/vscode-jsonrpc/-/vscode-jsonrpc-8.2.1.tgz",
             "integrity": "sha512-kdjOSJ2lLIn7r1rtrMbbNCHjyMPfRnowdKjBQ+mGq6NAW5QY2bEZC/khaC5OR8svbbjvLEaIXkOq45e2X9BIbQ==",
+            "devOptional": true,
```
nturinski, line 11137: I think you may need to update your npm version so these changes to the package-lock.json stop happening.

nturinski (approved):

package-lock.json
motm32: I updated my npm version (v11.12.1) and tried again but it still made these same changes. Going to merge this for now.

PR #1027, opened 2026-02-10
Add dynamic loading view when using `CopilotUserInput`

motm32:
Here is a video on how this currently works: 

https://github.com/user-attachments/assets/2488d1f5-7d80-4613-95d4-6a0bc3645117

To do: 

- [ ] Add in subscription, container app environment and container app to the view (needs to be tested with the re run work)


bwateratmsft (approved):

src/commands/createContainerApp/ContainerAppListStep.ts
```diff
@@ -6,6 +6,7 @@
 import { type ContainerApp, type ContainerAppsAPIClient } from "@azure/arm-appcontainers";
 import { LocationListStep, parseAzureResourceId, ResourceGroupListStep, uiUtils } from "@microsoft/vscode-azext-azureutils";
 import { AzureWizardPromptStep, nonNullProp, nonNullValueAndProp, type AzureWizardExecuteStep, type ConfirmationViewProperty, type IAzureQuickPickItem, type IWizardOptions } from "@microsoft/vscode-azext-utils";
+import { updateLoadingViewProgress } from "src/webviews/SharedViewState";
```
alexweininger: Although these imports are valid, I'd prefer if we change them to relative imports to match the existing code.

`import { updateLoadingViewProgress } from "../../webviews/SharedViewState";`

src/commands/createContainerApp/ContainerAppListStep.ts
```diff
@@ -53,6 +54,8 @@ export class ContainerAppListStep<T extends ContainerAppCreateContext> extends A
         if (containerApp) {
             context.containerApp = ContainerAppItem.CreateContainerAppModel(containerApp);
         }
+
+        updateLoadingViewProgress(localize('containerApp', 'Selected container app'));
```
alexweininger: the keys should more closely match the content
```suggestion
        updateLoadingViewProgress(localize('selectContainerApp', 'Selected container app'));
```

alexweininger (commented):
Two small comments to apply to a bunch of the code

src/webviews/LoadingView.tsx
```diff
@@ -4,10 +4,61 @@
...
+        };
+
+        window.addEventListener('message', handleMessage);
+        return () => window.removeEventListener('message', handleMessage);
+    }, []);
+
+    return (
+        <div className='loadingView'>
+            <div className='loadingContent'>
+                <Spinner labelPosition="below" label="Generating Copilot responses..." />
```
alexweininger, line 38: Does this label need to be localized and/or customizable?

src/webviews/LoadingView.tsx
```diff
@@ -4,10 +4,61 @@
...
+import { LoadingViewCommands } from './webviewConstants';
 
-export const LoadingView = () =>
-    <div className='loadingView'>
-        <Spinner labelPosition="below" label="Generating Copilot responses..." />
-    </div>;
+type LoadingViewProgressItem = {
+    name: string;
+    completed: boolean;
+}
```
alexweininger: `LoadingViewProgressItem` is declared here and again in `LoadingViewController.ts` with the same shape.

src/webviews/LoadingView.tsx
```diff
@@ -4,10 +4,61 @@
...
+export const LoadingView = () => {
+    const [progressItems, setProgressItems] = useState<LoadingViewProgressItem[]>([]);
+
+    useEffect(() => {
+        const handleMessage = (event: MessageEvent<ProgressMessage>) => {
+            const message = event.data;
+            if (message.command === LoadingViewCommands.AddProgressItem) {
+                setProgressItems(prev => [
+                    ...prev,
+                    { name: message.name, completed: true }
```
alexweininger: `completed: true` is hardcoded on every appended item and is never read on the render side (the check icon is always shown). Should we remove it?

src/webviews/LoadingView.tsx
```diff
@@ -4,10 +4,61 @@
...
+    }, []);
+
+    return (
+        <div className='loadingView'>
+            <div className='loadingContent'>
+                <Spinner labelPosition="below" label="Generating Copilot responses..." />
+                <div className='progressList'>
+                    {progressItems.length > 0 ? (
+                        progressItems.map((item, index) => (
+                            <div key={index} className='progressItem'>
```
alexweininger: `key={index}` is fine while the list is strictly append-only, but if items ever become updatable/removable this will cause reconciliation bugs. Using a stable id (or at least `${name}-${index}`) is more future proof

src/webviews/LoadingView.tsx
```diff
@@ -4,10 +4,61 @@
...
+                                <span className='checkmark codicon codicon-check'></span>
+                                <span className='itemName'>{item.name}</span>
+                            </div>
+                        ))
+                    ) : (
+                        Array.from({ length: 3 }).map((_, index) => (
```
alexweininger, line 48: Should this length still be hardcoded?

src/webviews/LoadingView.tsx
motm32: What do you mean by customizable? This view is only showing for copilot responses so the text shouldn't be changing if that is what you mean. 

As for localization a while ago I looked into it and there wasn't a super good way I found to localize without passing in the content outside the web views. I can probably do that here if we want or if you know of a way to localize within web views directly lmk.

src/webviews/LoadingView.tsx
motm32: This is for the placholder shining icon: 
image

It is hardcoded to 3 mostly cause I thought that looked nice lol.

src/webviews/LoadingView.tsx
alexweininger: Ok, we can leave it as-is for now

alexweininger (approved):

PR #1068, opened 2026-04-27
Remove web view stuff and add in new package

motm32:


esbuild.mjs
```diff
@@ -4,4 +4,18 @@
...
+const require = createRequire(import.meta.url);
+
+const pkgDist = dirname(require.resolve('@microsoft/vscode-azext-webview/package.json')) + '/dist';
+const outDir = 'dist/webview-assets';
+mkdirSync(outDir, { recursive: true });
+copyFileSync(join(pkgDist, 'views.js'), join(outDir, 'views.js'));
+copyFileSync(join(pkgDist, 'views.css'), join(outDir, 'views.css'));
+
+const configs = autoSelectEsbuildConfig();
+configs.extensionConfig.plugins.unshift(fixTasClientPlugin);
```
bwateratmsft: Where is `fixTasClientPlugin` defined?

motm32: Oh whoops I removed it but forgot to remove this line. My bad let me delete that.

PR #1069, opened 2026-04-28
Add ACA "flat" tree item provider and "reveal app" command.

philliphoff:
### Problem

The Azure Resources view was showing **two** sibling top-level Container Apps nodes per subscription:

1. **Container Apps** (mapped from `AzExtResourceType.ContainerAppsEnvironment`) — the rich hierarchy this extension contributes: environments → apps → revisions/configuration/logs/etc.
2. **ContainerApps** (mapped from `AzExtResourceType.ContainerApps`) — an unhandled stub that listed every `Microsoft.App/containerApps` resource as a bare leaf with no children, default Azure resource context menu, and the un-localized enum string as its label.

This happens because the Azure Resources host renders one top-level group per `AzExtResourceType` it discovers via Resource Graph, and `Microsoft.App/containerApps` resources surface as a separate type alongside `Microsoft.App/managedEnvironments`. Without a branch data provider registered for `AzExtResourceType.ContainerApps`, the host falls back to a default flat rendering — hence the asymmetric and confusing pair of root nodes.

This is a similar pattern to other Azure extensions where multiple related resource types each get their own top-level group (e.g. SQL servers + SQL databases, Cosmos variants), but in our case the secondary group contributes little value.

### Approach

Rather than try to suppress the secondary group (the host doesn't expose an API to do so), this PR registers a minimal branch data provider for `AzExtResourceType.ContainerApps` that:

- Renders each container app with the proper ACA icon, label, portal URL, and "View Properties" support.
- Has no children — the rich hierarchy continues to live under the environments group, which remains the canonical place to manage apps.
- Adds a context menu entry, **"Reveal in Container Apps Environments"**, that navigates to the equivalent app's parent managed environment under the environments hierarchy and expands it, so users coming in via the secondary group can quickly hop to the full tree.

### Why reveal goes to the *environment*, not the app

The Azure Resources host's `revealAzureResource` walks the tree using a `startsWith(resourceId)` ancestor check. A container app's ARM id is a *sibling* of its managed environment's id, not a descendant, so revealing the app id from the standalone group can never descend into the environments hierarchy — it only ever finds the same flat node we started from. Revealing the parent environment with `expand: true` is the closest behavior the host's public API supports, and it lands the user one click away from the target app.

### Notes / known limitations

- **Group label is still "ContainerApps"** rather than something nicer like "Container Apps". The host's group display names come from a hardcoded map in `vscode-azureresourcegroups` that has an entry for `ContainerAppsEnvironment` but not for `ContainerApps`, so the host falls back to the raw enum string. A friendlier label needs an upstream change to the host extension's `azExtDisplayInfo` map. (See [vscode-azureresourcegroups#1432](https://github.com/microsoft/vscode-azureresourcegroups/issues/1432)
- The secondary group still duplicates each app as a flat leaf. We can't avoid that without host-side changes; this PR makes the duplication useful instead of broken.
- Reveal requires the Azure Resources view to be visible; the command focuses it via `azureResourceGroups.focus` before calling `revealAzureResource` to make the navigation reliable.

### Testing

Manually verified in an Extension Development Host against a subscription containing managed environments and apps:

- Both top-level groups appear; the new "ContainerApps" group shows leaves with the correct ACA icon, label, "View Properties", and "Open in Portal".
- Right-clicking a leaf in the secondary group exposes only **Reveal in Container Apps Environments**, which scrolls to and expands the app's managed environment in the environments group.
- The existing rich hierarchy under the environments group is unchanged.

### DIFF
--- a/src/commands/revealInEnvironment.ts
+++ b/src/commands/revealInEnvironment.ts
@@ -1,0 +1,71 @@
+/*---------------------------------------------------------------------------------------------
+ *  Copyright (c) Microsoft Corporation. All rights reserved.
+ *  Licensed under the MIT License. See License.txt in the project root for license information.
+ *--------------------------------------------------------------------------------------------*/
+
+import { getResourceGroupFromId } from '@microsoft/vscode-azext-azureutils';
+import { parseError, type IActionContext } from '@microsoft/vscode-azext-utils';
+import * as vscode from 'vscode';
+import { ext } from '../extensionVariables';
+import { ContainerAppItem } from '../tree/ContainerAppItem';
+import { ContainerAppResourceItem } from '../tree/ContainerAppResourceItem';
+import { localize } from '../utils/localize';
+
+interface MaybeContainerAppRaw {
+    properties?: { managedEnvironmentId?: string };
+}
+
+/**
+ * Navigates from a {@link ContainerAppResourceItem} (under the standalone
+ * `AzExtResourceType.ContainerApps` group) to the equivalent app's parent
+ * environment under the `AzExtResourceType.ContainerAppsEnvironment` group.
+ *
+ * The host's `revealAzureResource` walks the tree using a `startsWith` ancestor
+ * check on Azure resource ids -- but a container app's ARM id is a *sibling* of
+ * its managed environment's id, not a descendant. So host-side reveal can only
+ * locate the env in the env-rooted hierarchy. We reveal the env with
+ * `expand: true` so the user immediately sees the target app under it.
+ */
+export async function revealInEnvironment(context: IActionContext, item: ContainerAppResourceItem): Promise<void> {
+    if (!item || typeof (item as ContainerAppResourceItem).containerAppId !== 'string') {
+        return;
+    }
+
+    const containerAppId = item.containerAppId;
+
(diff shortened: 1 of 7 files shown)


src/commands/revealInEnvironment.ts
```diff
@@ -0,0 +1,71 @@
...
+ * environment under the `AzExtResourceType.ContainerAppsEnvironment` group.
+ *
+ * The host's `revealAzureResource` walks the tree using a `startsWith` ancestor
+ * check on Azure resource ids -- but a container app's ARM id is a *sibling* of
+ * its managed environment's id, not a descendant. So host-side reveal can only
+ * locate the env in the env-rooted hierarchy. We reveal the env with
+ * `expand: true` so the user immediately sees the target app under it.
+ */
+export async function revealInEnvironment(context: IActionContext, item: ContainerAppResourceItem): Promise<void> {
+    if (!item || typeof (item as ContainerAppResourceItem).containerAppId !== 'string') {
```
bwateratmsft: Probably can use `isContainerAppResourceItem` here?

philliphoff: Yes, this works, too. Updated. (The original iterations did not which is how we ended up with this approach.)

bwateratmsft (approved):

PR #1081, opened 2026-05-19
Fix long running tests in support of migration to esbuild

MicroFish91:
Tests are now updated to be compatible with the ES build work. I also fixed a few smaller issues that cropped up since the tests hadn’t been running for a while.

One annoying change: I used to just delete the .vscode/ folder in each project between deployments and that worked just fine... but that’s apparently no longer enough. I assume a newer VS Code update introduced some kind of caching behavior that now has to be explicitly reset through the API; otherwise deployments can still pick up older cached settings even if the folder was deleted.

Update: The remote subscription we use only seems to support 1 container environment so I'm requesting an increase in quotas before I do a final verification pass


.vscode/settings.json
```diff
@@ -13,5 +13,7 @@
         "out": true,
         "**/node_modules": true,
         ".vscode-test": true
-    }
+    },
+    "js/ts.tsdk.path": "node_modules/typescript/lib",
```
MicroFish91: Fixes the mocha warnings that have been showing up

.vscode/launch.json
```diff
@@ -78,7 +78,7 @@
             "preLaunchTask": "Watch: ESBuild",
             "env": {
                 "DEBUGTELEMETRY": "",
-                "AzCode_EnableLongRunningTestsLocal": ""
+                "AzCode_EnableLongRunningTestsLocal": "true"
```
MicroFish91: Todo: Clear before merge

src/commands/createManagedEnvironment/ManagedEnvironmentListStep.ts
```diff
@@ -63,7 +63,7 @@ export class ManagedEnvironmentListStep<T extends ManagedEnvironmentCreateContex
     public confirmationViewProperty(context: T): ConfirmationViewProperty {
         return {
             name: localize('containerAppEnvironment', 'Container Apps Environment'),
-            value: nonNullValueAndProp(context.managedEnvironment, 'name'),
+            value: context.managedEnvironment?.name ?? context.newManagedEnvironmentName ?? '',
```
MicroFish91, line 66: Note for future confirmation views - we should prefer optionals over `nonNull` for these sections, I think it can still run when users select to create a new resource, in which case it could still be `undefined` when this gets run

MicroFish91:
Going to go ahead and take this out of draft so people can review.  I did verify that it passes remotely, but ever since they expanded our quota, we've still been getting intermittent restrictions which I messaged them about:

image

Just a heads up as you might notice that in some of the pipeline runs

bwateratmsft (approved):
Looks good but see my chat message about the Mocha warnings

bwateratmsft (approved):

